PR #159, opened 2014-12-29
NWChem: parse aonames

ATenderholt:
Use the section about shells to generate ```aonames```. 

This is still unfinished, and needs the following prior to merge:

* Higher shells and larger basis set examples. Ideally have f- and g-orbitals, in addition to d-orbitals.
* Unittest to make sure everything works as expected.
* Verify the order of cartesian/spherical labels, especially for d-orbitals and higher.


langner:
I see. Perhaps let's get #162 done first, because it kind of overlaps with this issue.

ATenderholt:
I'm comfortable with this PR being merged at this state, even though there is still work to do (#181).

PR #161, opened 2015-01-15
Upstream

Clyde-fare:
Made some modifications to the Gaussian parser:

* Reverted to the previous PES scan parsing code as the current version referenced an undefined variable 'hyphens'.

* Added an exception to the AO basis set extraction for ONIOM calculations that include GFPRINT. The current version assumes that basis sets are provided for atoms in atomic order whereas basis sets for the model region of an ONIOM calculation violate this assumption.

* Fixed a problem parsing IR intensities from frequency calculations where Gaussian has reported ****** inplace of the ferquency which causes the current version to break and pre-emptively applied the fix to raman intensities too.

* Added an oniomenergies field which extracts the energies associated with the components of an oniom calculation.

### DIFF
--- a/src/cclib/parser/data.py
+++ b/src/cclib/parser/data.py
@@ -61,6 +61,7 @@ class ccData(object):
         nocoeffs -- natural orbital coefficients (array[2])
         nooccnos -- natural orbital occupation numbers (array[1])
         optdone -- flags whether an optimization has converged (Boolean)
+        oniomenergies -- energies of components involved in an oniom calculation (list)
         scancoords -- geometries of each scan step (array[3], angstroms)
         scanenergies -- energies of potential energy surface (list)
         scannames -- names of varaibles scanned (list of strings)
@@ -123,6 +124,7 @@ class ccData(object):
         "nmo":            int,
         "nocoeffs":       numpy.ndarray,
         "nooccnos":       numpy.ndarray,
+        "oniomenergies":  list,
         "optdone":        bool,
         "scancoords":     numpy.ndarray,
         "scanenergies":   list,
--- a/src/cclib/parser/gaussianparser.py
+++ b/src/cclib/parser/gaussianparser.py
@@ -670,10 +670,8 @@ class Gaussian(logfileparser.Logfile):
 
             scanenergies = []
             scanparm = []
-
             colmnames = next(inputfile)
-            self.skip_line(inputfile, 'dashes')
-
+            hyphens = next(inputfile)
             line = next(inputfile)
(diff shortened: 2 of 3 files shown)


Clyde-fare:
I'm interested in looking at what the components of ONIOM calculations are doing - is there are smarter way to include this information without adding an attribute?

Clyde-fare:
At the moment there are fields for, "fragment orbital names", "fragment orbital overlap matrix", "names of fragments" and "indices of atoms in a fragment".

How about instead of using oniomenergies as a field we use "fragment energies"?

Reading in your other comments I see you're thinking of attaching separate ccdata objects for each fragment that looks like quite a big change. Leading up to such a change would it make sense to  create fields for fragment properties and implement population of those fields, then the refactor for 2.x could move the fields together into seperate ccdata fragment objects?

langner:
> At the moment there are fields for, "fragment orbital names", "fragment orbital overlap matrix", "names of fragments" and "indices of atoms in a fragment".
> 
> How about instead of using oniomenergies as a field we use "fragment energies"?

I like this idea, since it is a natural extension of the current attributes.

> Reading in your other comments I see you're thinking of attaching separate ccdata objects for each fragment that looks like quite a big change. Leading up to such a change would it make sense to create fields for fragment properties and implement population of those fields, then the refactor for 2.x could move the fields together into seperate ccdata fragment objects?

I'm not sure! I think there is a simple solution to this (although I don't have it yet), and I would tend to work towards that right away. It might be that some combination of fragment attributes and multiple ccData object would be the right way to go.

Clyde-fare:
I have example log files that fail but I'm not entirely certain how the cclib testing framework works. 

I think I need to add something like:

    def testGaussian_Gaussian09_OPT_oniom(logfile):
        """AO basis extraction broke with ONIOM"""

    def testGaussian_Gaussian09_oniom_IR_intensity(logfile):
        """Problem parsing IR intensity from mode 192"""
        assert hasattr(logfile.data, 'vibirs')
        assert len(logfile.data.vibirs) == 216

to regression.py

and then include the files OPT_oniom.log and oniom_IR_intensity.log somewhere but I don't see where.

Some of the files listed in regressionfiles.txt don't seem to exist in the repo?

langner:
Yes, you are thinking correctly. It is just that we keep the regressions in a separate repository since they can be large. See http://cclib.github.io/development.html#regression-tests in the docs for some details (will add more as time permits).

langner:
OK, this looks good. Thanks!

I'm going to merge later today after I get a chance to run the tests manually on this.

PR #162, opened 2015-01-17
Update nwchem files to 6.5 and ADF fixes to aooverlaps/foverlaps

ATenderholt:
I think I've done this correctly, but please verify the paths and filenames.


langner:
The tests are broken for NWChem. You are missing the AO overlap output for SP. Did you use the same input files?

Since we only have one version of NWChem right now, we could also opt for keeping both 6.0 and 6.5 in the unittests. Apparently there are some differences.

Besides that, I created a PR for the visual tests in ATenderholt/cclib#6.

ATenderholt:
Both ```dvb_sp_hf.nw``` and ```dvb_sp_ks.nw``` are identical between 6.0 and 6.5, and the checksums for each are identical to the input files I used on our cluster (still present).

langner:
According to the forum (http://www.nwchem-sw.org/index.php/Special:AWCforum/st/id902/overlap_matrix_of_atomic_orbital....html) you need to put in either `print "overlap"` or `print "ao overlap"` in the newer versions.

ATenderholt:
Good find.

I had moved the 6.0 files out of unittest since I thought we had decided to only have one version at a time, but I see that most parsers have multiple versions. Should be easy to take care of.

Do we want other versions like 6.1 and 6.3? Or does 6.0 and 6.5 suffice?

langner:
I think two versions in unittest is what we decided on some time ago, around when we moved ADF2004 to regressions. For NWChem, 6.5 is the newest, and my laziness tells me we should stick with 6.0 since we have (almost) all files already. I can rerun that one file in 6.0, since I have that installed on some Ubuntu 12.0 boxes.

ATenderholt:
Looks like ```print "overlap"``` and ```print "ao overlap"``` yield two different (equivalent?) formats:

```
 ====================================
 Begin overlap 1-e integrals
 ====================================
1eov          7   C  s     1   C  s           0.00030276
1eov          8   C  p     1   C  s           0.00013634
1eov          9   C  p     1   C  s           0.00081417
1eov         17   C  s     1   C  s           0.00158642
1eov         18   C  p     1   C  s           0.00234748
1eov         19   C  p     1   C  s           0.00287577
1eov         26   C  s     1   C  s           0.00000058
...
```

```
 global array: Temp Over[1:60,1:60],  handle: -996 

            1           2           3           4           5           6  
       ----------- ----------- ----------- ----------- ----------- -----------
   1       1.00000     0.24836    -0.00000    -0.00000     0.00000     0.00000
   2       0.24836     1.00000     0.00000    -0.00000     0.00000     0.00030
   3      -0.00000     0.00000     1.00000     0.00000     0.00000    -0.00014
   4      -0.00000    -0.00000     0.00000     1.00000     0.00000    -0.00081
   5       0.00000     0.00000     0.00000     0.00000     1.00000     0.00000
   6       0.00000     0.00030    -0.00014    -0.00081     0.00000     1.00000
   7       0.00030     0.02187    -0.00590    -0.03524     0.00000     0.24836
   8       0.00014     0.00590     0.00661    -0.01021     0.00000     0.00000
   9       0.00081     0.03524    -0.01021    -0.05268     0.00000     0.00000
...
```

The precision is higher in the first, so that's my preference. Plus, it yields smaller file sizes which could be significant with much larger systems.

langner:
> The precision is higher in the first, so that would normally be my preference. Plus, it yields smaller file sizes which could be significant with much larger systems. However, upon further investigation, it looks like it only prints some of the overlaps (e.g. there is no overlap printed for 1 & 2).

Great! Then let's have the first in unittest and the second as a regression since its straightforward to parse.

ATenderholt:
I initially thought the first would be in the unittest, but I'm concerned that it's missing some crucial data. For example, the 1-2 overlap is 0.24836 in the matrix but completely missing from the list. This value is basically identical to that of the Gaussian09 output.

langner:
> I initially thought the first would be in the unittest, but I'm concerned that it's missing some crucial data. For example, the 1-2 overlap is 0.24836 in the matrix but completely missing from the list. This value is basically identical to that of the Gaussian09 output.

I see. That is bad and we should report it to the devs. In that case we might consider doing it the other way around, and the testing becomes more important.

langner:
I just noticed that the overlap matrix as parsed from 6.0 has diagonal elements that are not 1.0 -- is it not normalized, or were we parsing the wrong thing?

ATenderholt:
I thought it was odd that the ```aooverlaps``` were found in a section that began with ```Mulliken population analysis``` since they should be independent of any population analysis. The block currently being parsed looks similar to the density matrix in the Gaussian09/dvb_sp.out file.

langner:
Yes... the thing we are parsing now seems to be the "population matrix": overlap matrix multiplied by the density matrix, element-wise.

langner:
> I initially thought the first would be in the unittest, but I'm concerned that it's missing some crucial data. For example, the 1-2 overlap is 0.24836 in the matrix but completely missing from the list. This value is basically identical to that of the Gaussian09 output.
> 
> I see. That is bad and we should report it to the devs. In that case we might consider doing it the other way around, and the testing becomes more important.

So I see @ATenderholt asked the question in the NWChem forum:
http://www.nwchem-sw.org/index.php/Special:AWCforum/st/id1564/#post_5583

I had some problems submitting there, so sent the same to the nwchem-users list. Same answer: basically it is a bug, so we should move it to regressions and perhaps do some special test on it.

ATenderholt:
@langner: good call on making the SP unit test add another check. Now we have ADF files failing. I don't understand how it ends up with a value of 0 for ```aooverlaps[0, 0]```. Nothing in the output file jumps out for that problem.

It also looks like we may be parsing the wrong sections here too. In dvb_sp_b.adfout, see: 

```
 S F O   P O P U L A T I O N S ,   M O   A N A L Y S I S
 =======================================================
  
 This section contains the SFO overlap matrices. This data is relevant to determine the bonding/anti-
 bonding nature of the SFO coefficients in the Molecular Orbitals (earlier section).
 A Mulliken population analysis is performed on (input-)selected MOs. All populations refer to SFOs.
 BAS populations may have been printed directly after the SCF part.


                                       === A.g ===


 ======  SFO Overlap Matrix (valence part only)

 column           1                     2                     3                     4
 row
    1    1.00003661461439E+00
    2   -1.19423653183338E-03  1.02672082257557E+00
    3    1.61252610407082E-03 -3.20130616061383E-02  1.03585855701835E+00
    4   -8.76000614071210E-04  1.73910124955406E-02 -2.39025517966704E-02  1.00484417063361E+00
...
```

vs. 

```
======  smat

 column           1                     2                     3                     4
 row
    1    1.00000000000000E+00
    2    2.43370854175314E-01  1.00000000000000E+00
    3    0.00000000000000E+00  0.00000000000000E+00  1.00000000000000E+00
    4    0.00000000000000E+00  0.00000000000000E+00  0.00000000000000E+00  1.00000000000000E+00
    5    0.00000000000000E+00  0.00000000000000E+00  0.00000000000000E+00  0.00000000000000E+00
...
```

ATenderholt:
And now I think I remember....

ADF makes symmetry adapted orbitals before before performing calculations, so the SFO overlap is for these. As an example, the first SFO in the Ag set is C1 (1s) + C4 (1s).

There still is a problem since it's not parsing correctly. I think it's just an off-by-one error:

```
>>> data.fooverlaps[0, 0]
0.0
>>> data.fooverlaps[1, 0]
1.00003661461439
>>> data.fooverlaps[2, 0]
-0.0011942365318333801
>>> data.fooverlaps[3, 0]
0.0016125261040708201
>>> data.fooverlaps[4, 0]
-0.00087600061407121005
```

The diagonal is also all zeros.

langner:
I see... so fooverlaps seems OK, except for the one-off error?

And, I guess the second example ('smat') is not from ADF?

ATenderholt:
Ok, so the last commit should fix the parsing of the overlap matrix. Unfortunately, it still fails because ```fooverlaps[0, 0]``` is 1.00003661461439. Should we create an ADF-specific test or relax the test to an ```assertAlmostEquals```?

ATenderholt:
Oh, and the ```smat``` matrix is for the BAS functions: http://www.scm.com/Doc/Doc2014/ADF/ADFUsersGuide/page225.html#keyscheme%20PRINT

langner:
> Oh, and the smat matrix is for the BAS functions: http://www.scm.com/Doc/Doc2014/ADF/ADFUsersGuide/page225.html#keyscheme%20PRINT

Do you know what BAS functions are?

langner:
> Oh, and the smat matrix is for the BAS functions: http://www.scm.com/Doc/Doc2014/ADF/ADFUsersGuide/page225.html#keyscheme%20PRINT
> 
> Do you know what BAS functions are?

To answer myself from the ADF manual :)...
<pre>The abbreviation BAS is used for references to the elementary Cartesian basis functions.</pre>

langner:
> Ok, so the last commit should fix the parsing of the overlap matrix. Unfortunately, it still fails because fooverlaps[0, 0] is 1.00003661461439. Should we create an ADF-specific test or relax the test to an assertAlmostEquals?

Perhaps. But... can we use the the `smat` output from ADF to get aooverlaps and test foverlaps additionally?

langner:
> Perhaps. But... can we use the the smat output from ADF to get aooverlaps and test foverlaps additionally?

Seems to work fine; see ATenderholt/cclib#8 for the code... this caused one regression for ADF to fail, but I think that is related to a conflict between the BAS and SFO function counts. It's the ADF/ADF2004.01/mo_sp.adfout regression, crashes while parsing.

langner:
> Seems to work fine; see ATenderholt/cclib#8 for the code... this caused one regression for ADF to fail, but I think that is related to a conflict between the BAS and SFO function counts. It's the ADF/ADF2004.01/mo_sp.adfout regression, crashes while parsing.

Following up on this, I think a little bit more work is needed to get this done properly for ADF. Perhaps we should get some other things out of the way before we deal with that, so I created #170 to continue that thread. The regression -- we can ignore for now or pass temporarily.

langner:
Shall we postpone to after 1.3.1?

langner:
I added a PR to your branch that resolves the remaining test failures (SFOs are not normalized, so we should just test some specific values).

langner:
Oh, still some regression tests failing with the new overlap tests...

ATenderholt:
Perhaps I should close this PR, and re-open ones with just the NWChem changes and the ADF changes/fixes.

langner:
Yeah, it's getting pretty long. But the I think we're good with the overall direction, so I would just merge this in and we can take care of the regressions later. What do you think?

ATenderholt:
I'd rather keep master passing so that other PRs can be pulled in without having to figure out whether the it fails due to a new change or not.

langner:
I'm on board with that. I'll try to get those regressions out of the way tomorrow.

ATenderholt:
I'm trying to tackle a few now, so make sure you pull from my branch first to see what I figure out.

ATenderholt:
For the ```mo_sp.adfout```, the error is related to using ```nbasis``` to set the sizes of the SMAT overlap matrix (in the atomic basis). Since ```nbasis``` is set using the (C)SFOs, it's using the fragment basis and can't always be used for SMAT.

It seems like we need at least one new variable to represent the size of the SFOs if we are to continue down the path of parsing both ```aooverlaps``` and ```fooverlaps```. Perhaps ```nfragbasis```?

ATenderholt:
I fixed one of the ORCA and Jaguar custom tests in ```regression.py```. @langner , double check to make sure you agree with how I did it.

langner:
> I fixed one of the ORCA and Jaguar custom tests in regression.py. @langner , double check to make sure you agree with how I did it.

I think it's OK, but could be simplified. See ATenderholt/cclib#10 for that and the final ADF fixes (well, one ignore until #170 is resolved).

langner:
It's green at last :tada:

PR #168, opened 2015-01-25
CML/CJSON export support

berquist:
Just to get things moving; a lot of things still need to be cleaned up.

One change I haven't made yet that would make life easier would be for the ccData object to carry the logfile's file name.

### DIFF
--- a/manifest.py
+++ b/manifest.py
@@ -24,7 +24,7 @@ files.append(os.path.join(source, "__init__.py"))
 files.append(os.path.join("src", "scripts", "ccget"))
 files.append(os.path.join("src", "scripts", "cda"))
 
-folders = ['bridge', 'method', 'parser', 'progress']
+folders = ['bridge', 'method', 'parser', 'progress', 'writer']
 for folder in folders:
     files.extend(glob.glob(os.path.join(source,folder,'*.py')))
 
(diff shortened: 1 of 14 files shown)


berquist:
For example, the build fails because we don't fail gracefully if Open Babel can't be imported.

Should the writers be imported by default?

I was also thinking of adding a command-line script, similar to `ccget`.

langner:
This is very exciting!

> Should the writers be imported by default?

I think they should, because this is a major thing, very useful. I would also think about making a script called `ccwrite` or something.

berquist:
Ok, I'll create a script called `ccwrite` for now.

Another goal is to minimize dependence on Open Babel, so if it can't be imported, the CML and CJSON writers are still usable (the XYZ writer doesn't depend on it).

langner:
Eric, I would prefer to keep this in a separate branch at least until 1.3.1 is out. Will create one... could you re-PR?

berquist:
Yes. When I create the PR, do I just pick another branch I want to push into, rather than `cclib:master`?

langner:
> Yes. When I create the PR, do I just pick another branch I want to push into, rather than cclib:master?

Yes, you can change it by clicking around.

PR #169, opened 2015-01-26
CML/CJSON writers

berquist:
Current goals are to "compartmentalize" Open Babel usage and create a `ccwrite` command-line script similar to `ccget`.

### DIFF
--- a/src/cclib/writer/cjsonwriter.py
+++ b/src/cclib/writer/cjsonwriter.py
@@ -25,6 +25,12 @@ class CJSON(filewriter.Writer):
 
     def __init__(self, ccdata, jobfilename=None,
                  *args, **kwargs):
+        """Initialize the chemical JSON writer object.
+
+        Inputs:
+          ccdata - An instance of ccData, parsed from a logfile.
+          jobfilename - The filename of the parsed logfile.
+        """
 
         # Call the __init__ method of the superclass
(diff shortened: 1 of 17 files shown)


langner:
This is great.

src/cclib/writer/cjsonwriter.py
```diff
@@ -0,0 +1,116 @@
...
+                cjson_dict['bonds']['connections']['index'].append(bond[1] + 1)
+            cjson_dict['bonds']['order'] = [bond[2] for bond in self.bond_connectivities]
+
+        cjson_dict['properties'] = dict()
+        if has_openbabel:
+            cjson_dict['properties']['molecular mass'] = self.pbmol.molwt
+
+        if has_openbabel:
+            cjson_dict['atomCount'] = self.obmol.NumAtoms()
+            cjson_dict['heavyAtomCount'] = self.obmol.NumHvyAtoms()
```
ATenderholt: Any reason these are set only if openbabel is present? Seems like the following should also work:

```
cjson_dict['atomCount'] = len(self.ccdata.atomnos)
cjson_dict['heavyAtomCount'] = len([ x for x in self.ccdata.atomnos if x > 1 ])
```

berquist: No, there is no particular reason; in fact, anything that can be calculated easily with cclib and Open Babel should use cclib, like in this case. I just missed it.

Issue #160, opened 2015-01-06
test suite in v1.3 fails with python 2.7


nonamedotc:
I am trying to run the test suite with python 2.7 and python 3 as part of packaging for Fedora. It seems that the test suite fails consistently with py2.7 but no problems in python3.

```bash
# python testall.py 
Traceback (most recent call last):
  File "testall.py", line 19, in <module>
    from cclib.parser import (ADF, GAMESS, GAMESSUK, Gaussian,
  File "/usr/lib/python2.7/site-packages/cclib/__init__.py", line 13, in <module>
    from . import parser
  File "/usr/lib/python2.7/site-packages/cclib/parser/__init__.py", line 16, in <module>
    from .adfparser import ADF
  File "/usr/lib/python2.7/site-packages/cclib/parser/adfparser.py", line 17, in <module>
    from . import logfileparser
  File "/usr/lib/python2.7/site-packages/cclib/parser/logfileparser.py", line 35, in <module>
    from .data import ccData
  File "/usr/lib/python2.7/site-packages/cclib/parser/data.py", line 50
SyntaxError: Non-ASCII character '\xc3' in file /usr/lib/python2.7/site-packages/cclib/parser/data.py on line 51, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

```bash
# python3 testall.py

[ ... ]

**** Jaguar8.3/dvb_scan.out: Customized relaxed potential energy surface scan unittest ****
testindices (testScan.JaguarScanTest)
Do the indices match the results from geovalues. ... ok
testnumindices (testScan.JaguarScanTest)
Do the number of indices match number of scan points. ... ok

----------------------------------------------------------------------
Ran 2 tests in 0.000s

OK

********* SUMMARY PER PACKAGE ****************
               Total	Passed	Failed	Errors	Skipped
ADF2007.01       56	 55	  0	  0	  1
ADF2013.01       56	 55	  0	  0	  1
Firefly8.0       73	 72	  0	  0	  1
GAMESS-UK7.0     62	 62	  0	  0	  0
GAMESS-UK8.0    112	112	  0	  0	  0
GAMESS-US2012    80	 79	  0	  0	  1
Gaussian03       99	 97	  0	  0	  2
Gaussian09       97	 95	  0	  0	  2
Jaguar7.0        59	 49	  0	  0	 10
Jaguar8.3       128	113	  0	  0	 15
Molpro2006       68	 63	  0	  0	  5
[...]

PR #171, opened 2015-02-03
Set encoding to UTF-8 in all parsers and methods for consistency

langner:
Per @berquist in #160, it won't hurt to declare the encoding everywhere.


berquist:
Should the encoding be added to every single Python file, not just the methods and parsers?

langner:
> Should the encoding be added to every single Python file, not just the methods and parsers?

It could, but parsers and methods seem to be the most likely to contain special characters so I did that for now.

Issue #163, opened 2015-01-21
Parsing XYZ Files


exit-1:
Would an xyz file (or other common chemical file formats) parser be appropriate for cclib? I imagine the supported data types for xyz file parsing would be atomcoords and atomnos.

ATenderholt:
I don't think we've ever really thought about supporting other data types like XYZ since there are other options (e.g. OpenBabel).

I'm not completely opposed to the idea, although I think it's a bit out of the scope of cclib. Also, for me, it's below other priorities such as better PES support, CML and other export, etc.. @langner , @berquist , and @baoilleach : what do you think?

langner:
I think we can easily support this via the openbabel bridge. Currently it does cclib->openbabel, but I see no reason to go the other way, too.

berquist:
I think that this goes hand-in-hand with support for the writers. Other than CML, I think most other common formats don't contain quantum data.

This leads to the problem where we couldn't rely entirely on OpenBabel for reading in *quantum chemical* data from a CML-type file, but I don't think other programs can either, except for [JUMBO](http://quixote.wikispot.org/JUMBO-Converters). I haven't checked what the status of AvogadroIO is other than for CJSON, which I spoke to Marcus about; it's not standardized yet for quantum data.

My opinion is that the bridge can be extended, but we shouldn't worry (for now) about extra information that OpenBabel can't handle and throws away during the parsing.

langner:
> My opinion is that the bridge can be extended, but we shouldn't worry (for now) about extra information that OpenBabel can't handle and throws away during the parsing.

I agree. It will solve the issue at hand, though.

> This leads to the problem where we couldn't rely entirely on OpenBabel for reading in quantum chemical data from a CML-type file, but I don't think other programs can either, except for JUMBO.

I've played around with Peter's converters a lot, with the intent of fusing them with cclib somehow. But the philosophy is really very different. So, while I think if there ever is quantum chemical data processing database online it should be running both Jumbo an cclib, I don't see a way to connect them naturally.

> I haven't checked what the status of AvogadroIO is other than for CJSON, which I spoke to Marcus about; it's not standardized yet for quantum data.

There has been talk again recently about using cclin within Avogadro, to expand the number of programs supported. The idea was to do this as a Googel Summer of Code project. @berquist maybe you are interested? Maybe you know someone interested? Here is the thread from avogadro-devel:
https://www.mail-archive.com/avogadro-devel@lists.sourceforge.net/msg00928.html

[...]

Issue #152, opened 2014-11-07
Move regression tests to cclib-data?


langner:
It seems to me it might be more appropriate to move all the regression tests over to cclib-data. That would make it easier to update (one commit to one repo instead of the current split), and it just seems more natural to have the tests alongside the files.

Of course, we can keep a way to call the regressions from this repo.

Issue #69, opened 2014-03-20
Unit tests for methods


langner:
The methods are not currently tested systematically and it is becoming apparent the code would benefit from that (start from #60 and #67).

langner:
We did some of this already (for v1.2?), but more remains.

langner:
Will review before v1.5.3 gets out.

langner:
The situation has improved a lot with Eric's moving of the doctests into regular tests. Will keep this open for v1.5.4 to do a thorough review.

Issue #54, opened 2014-03-07
Unit tests for scripts


langner:
Discussion about this started with #53

langner:
I think we've made progress here, so will review before getting v1.5.3. If there's something left to do, I'll make more specific issues.

langner:
Georgy added some tests in https://github.com/cclib/cclib/blob/master/test/io/testscripts.py

Additional action items:
* That doesn't really belong in `test/io` - find a better place for the scripts tests (their own dir?)
* Add more cases

langner:
Two additional things to check from #472:
* Add some more checks for script command line arguments
* Suppress stdout, because now it messes up the travis logs

PR #172, opened 2015-02-04
[bridge] Add a fallback mechanism to read XYZ files with OpenBabel (#153...

langner:
This is to address #163, and creates a fallback mechanism we can use in the future for `ccopen` to read file format with other libraries in order to import data.

The downside of this change is that `ccopen` return a ccData object in this case, instead of `Logfile`. It would be ideal to sidestep this somehow, but I haven't had time yet to figure out the simplest way. Any ideas?


langner:
The conditional import is needed to get `ccopen` (and the standard unit tests) working when openbabel is not installed. Also, on Ubuntu 12/14 the Python3 version of openbabel is not available yet.

langner:
I resolved the `ccData`/`Logfile` issue in c873bbb. Basically, by introducing another step between `ccget` and `ccopen` called `ccread`, which is responsible for parsing the data from the logfile once it is opened by `ccopen`. 

`ccread` is a simple function and will not change how cclib works, but it will provide more fine grained control. So, if `ccopen` does not recognize a logfile and return `None`, I made `ccread` call a fallback function that will check if the extension is ".xyz" and will attempt to use OpenBabel to load some data. That is specifically for #163, but can be extended for more formats.

berquist:
This starts to answer some of the questions I had about testing the new writers and all the other non-ccopen functionality.

test/bridge/openbabel.py
```diff
@@ -0,0 +1,11 @@
+from cclib.parser import ccData
+from cclib.bridge import cclib2openbabel
+
+
+def test_uracil():
+    data = ccData(cclib2openbabel.readfile("../data/misc/uracil.xyz", "XYZ"))
```
berquist: I'm getting an error here; I think the path should be ```"../../data/misc/uracil.xyz"```.

Can you also rename the Python test file? It was causing some weird problems with importing openbabel due to the name.

langner: Yeah, I commit hastily and will fix that in a commit soon. Thanks!

langner:
> This starts to answer some of the questions I had about testing the new writers and all the other non-ccopen functionality.

Well, one of my long term goals with cclib is structure the tests better and raise the coverage, since consistently the bugs crop up in places that do not have tests. There are some separate issues create for that (#45, #152, #69, #54), but this is a good chance to go in that direction. The tests we have are a lifesaver, so this will definitely be worthwhile.

So yeah, I think we need to test the bridges, scripts and writers. Best to separate that a bit from the unit tests. Separate directories seem like the most natural way to do that.

src/cclib/bridge/cclib2openbabel.py
```diff
@@ -40,6 +40,35 @@ def makeopenbabel(atomcoords, atomnos, charge=0, mult=1):
...
+        attributes['atomcoords'].append([atom.GetX(), atom.GetY(), atom.GetZ()])
+        attributes['atommasses'].append(atom.GetAtomicMass())
+        attributes['atomnos'].append(atom.GetAtomicNum())
+    return attributes
+
+def readfile(fname, type):
+    """Read a file with OpenBabel and extract cclib attributes."""
+
+    obc = ob.OBConversion()
+    obc.SetInFormat(type)
```
ATenderholt: Should we check that the format library is correctly loaded and if not, log a warning?

```
retval = obc.SetInFormat(type)
if not retval:
   logging.warning("Unable to load the %s reader from OpenBabel.")
```

Also, should we ensure that ```type``` is a string?

langner: I'm afraid I don't know enough about how OpenBabel works to answer that ATM.

langner: > Should we check that the format library is correctly loaded and if not, log a warning?
> 
> retval = obc.SetInFormat(type)
> if not retval:
>    logging.warning("Unable to load the %s reader from OpenBabel.")

OK, this certainly won't hurt (see 3db3a9f). Although I don't see how to get a logger in there. We could import ccData and return an object instead of a dict as happens now.

> Also, should we ensure that type is a string?

Well, we could put an assert there, but all the functions using `readfile` will be passing a string for sure.

ATenderholt:
I think this looks good, so I support merging it. What do you think @berquist?

langner:
So... how about getting this in today? Any problems? I've pushed all other things over to 1.4, so after this and perhaps some version changes I will release 1.3.1.

langner:
Thanks!

PR #176, opened 2015-02-16
Merge writer into master

langner:
There were some trivial conflicts.


langner:
Any reason not to merge this now?

berquist:
I can't think of any reason, other than it has no tests, but I'm still thinking of how to do those.

langner:
> I can't think of any reason, other than it has no tests, but I'm still thinking of how to do those.

We can add tests later. What's important is that current tests are not broken.

PR #177, opened 2015-02-18
Merge DALTON parser

langner:
OK, I guess it's also time to pull in the DALTON parser!

I fixed some minor conflicts in ccopen and unittests.


src/cclib/parser/daltonparser.py
```diff
@@ -0,0 +1,277 @@
...
+            self.moenergies[0] = moenergies
+            self.mosyms = [[]]
+            self.mosyms[0] = mosyms
+
+            if not hasattr(self, "nmo"):
+                self.nmo = self.nbasis
+                if len(self.moenergies[0]) != self.nmo:
+                    self.set_attribute('nmo', len(self.moenergies[0]))
+
+            self.mocoeffs = [numpy.zeros((self.nmo, self.nbasis), "d")]
```
berquist: If this is going to be merged in as-is, we probably want to implement this right away, otherwise it should get removed.

langner: Good point. I don't understand fully how DALTON print the MO coefficients, so I've commented it out for now.

berquist:
Not crucial because there aren't test entries for it yet, but the TDDFT job will need to be altered (change the basis set from 3-21G to STO-3G).

langner:
Yes, quite a bit to work on here.

Issue #167, opened 2015-01-22
Increase coverage for DALTON


langner:
Thanks to @cstein we have a DALTON parser:
https://github.com/cclib/cclib/tree/dalton

We will keep this branch separate until after 1.3.1, and aim to release 1.4 with the new parser after that.

cstein:
@langner it looks like it fails the regression testing and that is because the cclib-data hasn't been merged - if you merge that pull-request, it should work. :+1:

langner:
> @langner it looks like it fails the regression testing and that is because the cclib-data hasn't been merged - if you merge that pull-request, it should work. :+1:

Indeed. I hesitated with that merge a bit, because I wanted to make sure it wouldn't break the regression pipeline for master. There is a warning, but that's fine until the merge.

langner:
We now have decent coverage for DALTON, so I consider this fixed.

PR #179, opened 2015-02-23
Some updates to DALTON parser

langner:
See #167


test/testdata
```diff
@@ -133,7 +133,8 @@ Scan      Jaguar       JaguarScanTest       basicJaguar8.3        dvb_scan.out
 
 SP        ADF         ADFSPTest             basicADF2007.01       dvb_sp_b.adfout
 SP        ADF         ADFSPTest             basicADF2013.01       dvb_sp_b.adfout
-SP        DALTON      DALTONSPTest          basicDALTON-2013      b3lyp_energy_dvb_sp.out
+SP        DALTON      DALTONSPTest          basicDALTON-2013      sp_b3lyp_dvb.out
+SP        DALTON      DALTONSPTest          basicDALTON-2013      sp_rhf_dvb.out
```
ATenderholt: Why ```sp_b3lyp_dvb.out``` and ```sp_rhf_dvb.out``` instead of ```dvb_sp_ks.out``` and ```dvb_sp_hf.out```, respectively, like was done with NWChem?

langner: Well, it's like this. DALTON takes two files as input: a .dal files that contains the calculation instructions and a .mol files that described the molecule (both are in the repo, and they are repeated in the output). The default output filename is the concatenation of the basenames of the two (in the order I gave aboce). So, it seemed natural to do it this way.

Since we are at the initial stages of the parser, it might make sense to regularize this to the cclib convention. I am actually for this, but did not care enough to decide myself. If you also think it's a good idea, I will rename.

ATenderholt: Now I understand. I think the current filenames are fine, although there could be minor changes to somewhat better match the NWChem files (i.e. b3lyp -> ks and rhf -> hf).

berquist: You actually don't need two files; you can have just a .dal file, with the contents of the .mol file at the top, a blank line, and then the job directives.

Which raises the question: should geometries be combined with the inputs, or kept separate where possible? Not all programs allow the separation.

langner: > You actually don't need two files; you can have just a .dal file, with the contents of the .mol file at the top, a blank line, and then the job directives.

I did not know that. I will try it out soon. That would clear things up and we could use the cclib convention.

> Which raises the question: should geometries be combined with the inputs, or kept separate where possible? Not all programs allow the separation.

I think for the purposes of testing it is better that they are together, especially since all other programs we support have it that way by default.

langner: OK, did this.

langner:
Please also pull in cclib/cclib-data#9

langner:
Minor conflicts in testSP had to be resolved with master.

PR #182, opened 2015-03-03
DALTON: Updates for recent 2015 version

cstein:
A small patch to fix two minor issues for the recent DALTON 2015 release.

The dalton branch is gone, so this is headed straight into master I guess.

@langner very nice work on the parser.


langner:
Ah yes, this is great! Thanks a lot. Looks good, and I will look closer at this a bit later.

PR #185, opened 2015-03-09
Update/fix some the ORCA relaxed scan examples and associated test

ATenderholt:
Minor fix for the ORCA v2.9 relaxed scan example and add the corresponding v3.0 example. Also updates to the testdata and scan unit test.


langner:
Great!

PR #186, opened 2015-03-09
QChem: add the quantum number for each shell in aonames.

berquist:


langner:
So are you planning to add the tests to this PR?

berquist:
I wasn't; I think that the tests for `aonames` should be added simultaneously for as many parsers as possible.

langner:
OK

PR #187, opened 2015-03-09
Modernization of unit tests

langner:
Besides using some features available in Python 2.7 that have matured since 2009 and getting rid of cclib's venerable bettertests, these changes simplify greatly how skipped tests can be handled (with decorators).


test/testScan.py
```diff
@@ -32,6 +31,7 @@ def testnumindices(self):
         else:
             self.assertEquals(len(self.data.optdone), 12 + self.extra)
 
+    @skipForParser('ORCA', "?")
```
ATenderholt: @langner Any reason this was added for ORCA? Were you working on these tests prior to my recent update to the ORCA data files?

ATenderholt:
Nice work! Was about to merge, but then I saw that there was a customized skip for the ORCA scan test (see above).

test/testScan.py
langner: I don't know why this test was skipped -- you added the exception in 2014 according to blame.

I think many of the exceptions currently applied can now be lifted. But I didn't explore that direction for this PR.

PR #189, opened 2015-03-13
Move regressions to cclib-data

langner:
This is to get #152 rolling and needs to be considered together with cclib/cclib-data#10. My main goal at this point is to get the regression code into cclib-data and still have Travis happy and things working as usual. Later we can worry about what the best way to import things is.


langner:
OK, Travis won't work until cclib/cclib-data#10 is merged.

ATenderholt:
How do you kick off another Travis build for this PR? I just merged cclib/cclib-data#10.

langner:
Another commit is needed...

ATenderholt:
Is there anything else to be added?

langner:
Yes, but perhaps separately. It'd be good to keep the tests running smooth, and with this out of the way it will be clear where new regressions should go.

berquist:
Ok, I can get everything to run smoothly. Looks good.

PR #190, opened 2015-03-17
Some more modernization to tests

langner:
Here are some further changes to the tests, which I hope make them easier manage. Note that cclib/cclib-data#11 needs to be merged in before Travis will be happy.


langner:
Comment on the renaming - I think it makes sanse to call the tests that deal with logfiles "data tests", as opposed to unit tests for parsers, bridges, etc. (added a stub for parsers in one commit).

So I put all the "data tests" in the data/ subdirectory and refactored the `testsall` code in `test_data.py`.

langner:
Any comment on this?

berquist:
I'm going to take a look at this tomorrow; haven't been able to give it a thorough look-over until now.

langner:
Great! Looking forward to that. I might add a few commits later today for you then.

test/data/testSP.py
```diff
@@ -14,8 +14,8 @@
 
 import numpy
 
-from testall import skipForParser
-from testall import skipForLogfile
+from skip import skipForParser
+from skip import skipForLogfile
```
berquist: These are giving me an `ImportError`, unless I change them to relative imports. Same with `testSPun` and `testScan`.

langner: > These are giving me an ImportError, unless I change them to relative imports. Same with testSPun and testScan.

Are you sure the import from `skip` is the problem? Could you show the actual error you get here after the latest commits?

langner: > These are giving me an ImportError, unless I change them to relative imports. Same with testSPun and testScan.
> 
> Are you sure the import from skip is the problem? Could you show the actual error you get here after the latest commits?

OK, now I see this problem pertains to Python3 only, where the importing is a bit different. This should be fixed in one of the last commits.

langner:
I forgot to test everything in Python3, and there are some importing differences and other changes. Commit  552f1c9 takes care of these things. The code is still somewhat of a mess, but the tests should now pass and they are more comprehensive than before.

Also, the method tests are not working/organized yet.

langner:
Like I suggested in langner/cclib#4, just not testing bridges with Travis is the easy way out. And reasonable. I think it's OK, since it's not a core part of cclib. I created #200 to work on this in the unspecified future.

PR #192, opened 2015-03-24
ccopen: fix cclib2openbabel warning

langner:
This is to address a warning issue raised by Noel:
https://sourceforge.net/p/cclib/mailman/message/33632090/

### DIFF
--- a/src/cclib/parser/ccopen.py
+++ b/src/cclib/parser/ccopen.py
@@ -36,8 +36,9 @@ from .qchemparser import QChem
 
 try:
     from ..bridge import cclib2openbabel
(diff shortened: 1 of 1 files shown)


baoilleach:
I think the print statement should be a few lines after where it is now.

langner:
> I think the print statement should be a few lines after where it is now.

Yes, the current commit should be better. I don't know how that happened :)

langner:
Note to self: add some unit tests for the function in `ccopen.py` to #190. For this, I think the file will need to be renamed to sidestep name conflict in cclib.parser.

PR #198, opened 2015-04-27
DALTON: expand parser capabilities and test coverage

berquist:
Work on extending the parser and tests will continue in this pull request; I've decided to open it now so progress can be tracked.


langner:
As I mentioned in berquist/cclib#4, DALTON does not seem to support UHF, se we can remove those data tests.

Also, concerning how to get output for gbasis: http://daltonprogram.org/forum/viewtopic.php?f=9&t=560&view=unread#unread

berquist:
What might be misleading is that DALTON only supports ROHF references, not UHF (very strange, I know). So "unrestricted" really means performing ROHF calculations.

langner:
> What might be misleading is that DALTON only supports ROHF references, not UHF (very strange, I know). So "unrestricted" really means performing ROHF calculations.

So would you suggest just using ROHF as our unrestricted test case? Isn't it equivalent to RHF for the DVB example we have where all electrons are paired?

berquist:
I thought all of the `dvb_sp_un` calculations were the doublet cation at the neutral singlet geometry?

langner:
You are correct!

berquist:
Just to let you know of my next goal, it's to get the TD attributes parsed.

In addition to a few missing DALTON attributes, it looks like Travis is failing due to `testcontractions`.

langner:
OK. @berquist: are the SP_un (ROHF) logfiles final in you mind?

berquist:
I don't think so; I need to add the MOLBAS lines to them and re-run them, then I'll make sure the SPun tests pass.

berquist:
Ok, after a couple of busy months, it's time for me to finally polish this off. Here are some notes for myself on what needs to happen to close this.

Travis is failing due to:

1. No `data.mult` or `data.mocoeffs` in `dvb_sp_un_ks.out/dvb_sp_un_hf.out` for DALTON 2013 or 2015. I remember there being precision issues for `data.mocoeffs` (in addition to some problems due to symmetry), but `data.mult` is definitely required for when we parse any ROHF output.
2. Missing the following TDDFT attributes (there aren't CIS inputs, just TD-B3LYP).
    * `etenergies`
    * `etoscs`
    * `etsecs`
    * `etsyms`
3. `data.contractions` is incorrect for the following `C_bigbasis.out` tests:
    * `basicFirefly8.0`
    * `basicGAMESS-US2012`
    * `basicJaguar8.3`
    * `basicMolpro2006`
    * `basicMolpro2012`
    * `regression/GAMESS/GAMESS-US2006/C_bigbasis_2006.02.22.r3.out`
    * `regression/GAMESS/PCGAMESS/C_bigbasis.out`
    * `regression/Molpro/Molpro2006/C_bigbasis_cart.out`
4. There are array shape issues for the ROHF DALTON 2013 and 2015 outputs. There is an old comment above that's telling me I need to make sure the inputs are correct for these.

The plan is to finish 1, 2, and 4 in this PR, and 3 should be handled elsewhere.

test/data/testBasis.py
```diff
@@ -46,30 +52,35 @@ def testsizeofbasis(self):
...
-            for fns in atom:
-
-                 # Add 3 for P, 5 or 6 for D, and so forth.
-                ftype = fns[0]
+            for (ftype, contraction) in atom:
                 total += multiple[ftype]
 
         self.assertEquals(self.data.nbasis, total)
 
+    def testcontractions(self):
```
berquist: I have to admit I'm confused right now, not because these tests cause non-DALTON failures, but because these tests only seem to be in this branch/PR and not in master. How did they get in here? I can't find them in the commit list.

langner: I added them in a PR to your PR :) I think here: ce70756a516f76fa82caeef4b41c8dbc96484325

PR #199, opened 2015-04-27
Update main README to use badges

berquist:
A lot of projects use [colored badges](http://shields.io/) on their READMEs to quickly show some important information about the project, and some of them automatically update, like the PyPI/GitHub release versions and the Travis CI build information. Each badge has been turned into a link for the appropriate information.


langner:
So I've thought about adding something pretty here before, at least once, and decided it was not worth the effort. When cclib was created, READMEs were simply for reading and GitHub did not exist. Who knows where cclib will end up in another 5 years. Aside from that, I don't have anything against spricing it up.... what do you think @ATenderholt?

ATenderholt:
I actually like it, so I think this PR should be merged.

PR #202, opened 2015-07-17
Prevent duplicate parsing of symmetry labels, and parsing of indices …

martinp23:
…as labels.

See gist https://gist.github.com/martinp23/0239fb8148cd767e00fb

### DIFF
--- a/src/cclib/parser/gaussianparser.py
+++ b/src/cclib/parser/gaussianparser.py
@@ -834,14 +834,14 @@ class Gaussian(logfileparser.Logfile):
             while line.strip() != "":
 
                 # The line with indices
-                if line[1:15].strip() == "" and line[15:23].strip().isdigit():
-                    freqbase = int(line[15:23])
+                if line[1:15].strip() == "" and line[15:60].split()[0].isdigit():
+                    freqbase = int(line[15:60].split()[0])
                     if freqbase == 1 and hasattr(self, 'vibfreqs'):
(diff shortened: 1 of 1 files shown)


TDI-tenderholt:
Thanks for the PR. It looks like you're correctly handling the re-parse of the second frequency block.

Do you have a test file we can add to our test or regression suite? This is something that should have been caught since we check for appropriate lengths of ```vibsyms``` with our two ```dvb_*``` Gaussian test files.

mwykes:
It looks like c752478 in #202 fixes the incorrect parsing of mode indices as symmetry labels, but did not fully resolve the issue of duplication of symmetry labels. Before c752478, the code would pass indices as symmetry labels, and parse all the labels twice, so vibsyms contained three times as many elements as it should. len(vibsyms)=630 in the 'before' case in https://gist.github.com/martinp23/0239fb8148cd767e00fb . After c752478, the code no longer parsed indices as labels, but the labels were still being parsed twice, so len(vibsyms)=420 in the 'before' case in https://gist.github.com/martinp23/0239fb8148cd767e00fb. While if you run the same test with the updated code in 8eb321f (in #204) you should obtain len(vibsyms)=210. I've also changed the G09 dvb_ir test log file so that it now contains two frequency blocks. The tests with the new log file yielded an incorrect number of vibsyms with c752478 and before, but pass with 8eb321f.

PR #203, opened 2015-08-06
gaussianparser: Support high precision vib displ

mwykes:
Parse high precision vibrational displacements (freq=hpmodes keyword) if they are present.

When vibrational displacements are used for purposes other than visualization (e.g. for subsequent calculations, such as vibronic coupling in fluorescence), it is preferable to be able to parse higher-precision vibrational displacements printed to 5 figures by Gaussian when the freq=hpmodes keyword is specified. 

If freq=hpmodes is specified, two displacement blocks are printed - the first in higher precision, the second in lower precision. cclib currently only parses the lower precision displacements. The code changes proposed in this pull request parse the high precision displacements if they are present. 

I have also added new regression tests to cclib-data to test the new functionality, which I will also request to be pulled. 

The changes I made passed the unit tests, and regression tests when I ran them.

PR #204, opened 2015-08-07
Parse G09 high precision vibdisps if they are present and prevent duplicate parsing of vibsyms.

mwykes:
Parse high precision vibrational displacements (freq=hpmodes keyword) if they are present, and prevent duplicate parsing of vibrational symmetry labels which was occurring. 

When vibrational displacements are used for purposes other than visualization (e.g. for subsequent calculations, such as vibronic coupling in fluorescence), it is preferable to be able to parse higher-precision vibrational displacements printed to 5 figures by Gaussian when the freq=hpmodes keyword is specified. If freq=hpmodes is specified, two displacement blocks are printed - the first in higher precision, the second in lower precision. cclib currently only parses the lower precision displacements. The code changes proposed in this pull request parse the high precision displacements if they are present. This pull request includes c752478, which addressed duplicate parsing of symmetry labels and parsing of indices for freq=hpmodes log files, but did not completely resolve the duplication issue. In this pull request the duplication issue is fully resolved, and the G09 dvb_ir basic regression log file has been changed to include a high-precision frequency block. The old code did not pass the regression test with this new log file (wrong number of vibsyms), but the new code in this pull request does pass with both the new and old log files. This pull request replaces #202 and #203. 

I have also added new regression tests to cclib-data to test the new functionality, which I will also request to be pulled.


mwykes:
I discovered a bug introduced in this pull reqeust, so closed this request and opened a new one, #205, in which the bug is fixed.

PR #208, opened 2015-09-08
Fix typo pointed out by Alexis Otero-Calvis

langner:


langner:
The failed checks are not related to this fix.

PR #209, opened 2015-09-09
NWChem: calculations without titles result in needing different skip_lines().

berquist:


berquist:
This goes together with https://github.com/cclib/cclib-data/pull/15.

Since Travis is going to keep failing due to the DALTON symmetry labels, I'll take a look at that now.

langner:
That's fine. This touches only NWChem and the tests for that pass. We will deal with the DALTON issues separately.

PR #210, opened 2015-09-19
ORCA: extract/convert energies in hartree rather than extract energies in eV

berquist:
There is a precision problem when extracting the SCF energies from ORCA outputs. For example, using [this file](https://github.com/berquist/qmspeedtest/blob/master/berquist/frank_shared/orca/2.9.1/hf/hf.out#L5459),
```
Total Energy : -644.67570604 Eh -17542.51781 eV
```
the energy returned by cclib is -644.6756671064783 Eh, which is too high by 0.0389335217 mEh. This must be due to ORCA internally performing a truncated unit conversion. The error is small, but I think we should use the output in hartree and convert it, since the programs are doing their calculations in hartree to begin with.

### DIFF
--- a/src/cclib/parser/orcaparser.py
+++ b/src/cclib/parser/orcaparser.py
@@ -128,7 +128,7 @@ class ORCA(logfileparser.Logfile):
 
             while not "Total Energy       :" in line:
                 line = next(inputfile)
-            energy = float(line.split()[5])
+            energy = utils.convertor(float(line.split()[3]), "hartree", "eV")
             self.scfenergies.append(energy)
 
             self._append_scfvalues_scftargets(inputfile, line)
@@ -151,7 +151,7 @@ class ORCA(logfileparser.Logfile):
             if not hasattr(self, "scftargets"):
                 self.scftargets = []
 
-            energy = self.scfvalues[-1][-1][0]
+            energy = utils.convertor(self.scfvalues[-1][-1][0], "hartree", "eV")
             self.scfenergies.append(energy)
 
             self._append_scfvalues_scftargets(inputfile, line)
@@ -379,8 +379,8 @@ class ORCA(logfileparser.Logfile):
             line = next(inputfile)
             while len(line) > 20:  # restricted calcs are terminated by ------
                 info = line.split()
-                self.moenergies[0].append(float(info[3]))
-                if float(info[1]) > 0.00:  # might be 1 or 2, depending on restricted-ness
+                self.moenergies[0].append(utils.convertor(float(info[2]), "hartree", "eV"))
+                if float(info[1]) > 0.00: # might be 1 or 2, depending on restricted-ness
                     self.homos[0] = int(info[0])
                 line = next(inputfile)
 
@@ -396,7 +396,7 @@ class ORCA(logfileparser.Logfile):
                 line = next(inputfile)
                 while len(line) > 20:  # actually terminated by ------
                     info = line.split()
-                    self.moenergies[1].append(float(info[3]))
+                    self.moenergies[1].append(utils.convertor(float(info[2]), "hartree", "eV"))
                     if float(info[1]) == 1.00:
                         self.homos[1] = int(info[0])
                     line = next(inputfile)


src/cclib/parser/orcaparser.py
```diff
@@ -151,7 +151,7 @@ def extract(self, inputfile, line):
             if not hasattr(self, "scftargets"):
                 self.scftargets = []
 
-            energy = self.scfvalues[-1][-1][0]
```
langner: Was the lack of conversion here previously a bug?

berquist: At first I thought I made a mistake here, but now I'm not so sure. I have a different output where the first geometry step is unconverged:

```
...
123  -2921.2788160254  -0.000000336370 0.00012032  0.00000135  0.0000431 0.0000
124  -2921.2788164185  -0.000000393072 0.00012020  0.00000135  0.0000419 0.0000

*****************************************************
*                      ERROR                        *
*        SCF NOT CONVERGED AFTER 125 CYCLES         *
*****************************************************
...
---------------
SCF CONVERGENCE
---------------

Last Energy change         ...   -3.9307e-07  Tolerance :   1.0000e-08
Last MAX-Density change    ...    1.2020e-04  Tolerance :   1.0000e-07
Last RMS-Density change    ...    1.3465e-06  Tolerance :   5.0000e-09
Last DIIS Error            ...    4.1883e-05  Tolerance :   5.0000e-07
...

FINAL SINGLE POINT ENERGY     -2921.277851575335
...
122  -2921.3318360174  -0.000000001792 0.00001354  0.00000010  0.0000050 0.0000
123  -2921.3318359899   0.000000027559 0.00001309  0.00000010  0.0000049 0.0000
                 **** Energy Check signals convergence ****

               *****************************************************
               *                     SUCCESS                       *
               *           SCF CONVERGED AFTER 124 CYCLES          *
               *****************************************************
...
Total Energy       :        -2921.33098298 Eh          -79493.45742 eV
  Last Energy change         ...   -1.4893e-09  Tolerance :   1.0000e-08
  Last MAX-Density change    ...    1.2671e-05  Tolerance :   1.0000e-07
...
FINAL SINGLE POINT ENERGY     -2921.330982975234
...
```
With the current version of the parser (pre-pull request),
```
scfvalues[0][-1] == array([ -3.93070000e-07,   1.20200000e-04,   1.34650000e-06])
scfvalues[1][-1] == array([ -1.48930000e-09,   1.26710000e-05,   1.00000000e-07])
scftargets[0] == array([  1.00000000e-08,   1.00000000e-07,   5.00000000e-09])
scftargets[1] == array([  1.00000000e-08,   1.00000000e-07,   5.00000000e-09])
scfenergies[0] == -3.9307200000000001e-07
scfenergies[1] == -79493.457420000006
```
With the version of the parser from this pull request,
```
scfvalues[0][-1] == array([ -3.93070000e-07,   1.20200000e-04,   1.34650000e-06])
scfvalues[1][-1] == array([ -1.48930000e-09,   1.26710000e-05,   1.00000000e-07])
scftargets[0] == array([  1.00000000e-08,   1.00000000e-07,   5.00000000e-09])
scftargets[1] == array([  1.00000000e-08,   1.00000000e-07,   5.00000000e-09])
scfenergies[0] == -1.06960335443736e-05
scfenergies[1] == -79493.462236363775
```
So, it looks like there's no conversion for `scfvalues/scftargets` from hartree. I'm not sure what the other parsers do, since they all use different convergence criteria, and sometimes not delta_E at all. So, I think we should be self-consistent and not convert this line to eV unless *all* of the energies in `scfvalues/scftargets` are converted to eV. Is that a bug or an oversight?

langner: I don't think there's a good answer here. I would say that as long as it's documented and clear what the units are, it does not matter too much. Since scfenergies/scfvalues are somehow "raw" (depend on the program), perhaps it is better not to convert the numbers.

langner:
I think this is relatively easy to get finished now, but I'm pushing back to 1.4.1

berquist:
Ok, I've fixed the merge conflicts. Is there anything else I should do here?

langner:
... and I added a regression test in cclib/cclib-data#41

PR #214, opened 2015-11-19
Gaussian formatted checkpoint files, transition dipoles, basic turbomole parser & bug fixes

mwykes:
There are quite a few changes in this pull request, some big, some small. The bigger ones include:
- Parsing Gaussian formatted checkpoint files which contain higher-precision data
- New attributes for various et transition dipoles, now parsed from Gaussian fchk and log files. 
- A basic Turbomole parser which parses output from a vibrational frequency calculation.
- New functionality in logfileparser set_attribute; it can check if new and old values are within tolerances
The smaller changes include:
- Several bug fixes and where relevant, improvement of tests to pick up similar bugs in future

All unit tests are passing with Python 2.7 and 3.3. Note that this pull request will fail with the current regression tests, but work with the new regression tests which are in my latest cclib-data pull request.


langner:
Looks good generally. I'll get to this next week probably.

langner:
Digging deeper into this PR. It would be much easier if you could break this up into several smaller pull requests... is that possible?

mwykes:
Sure, I can split this into smaller pull requests. I have just made the first one, see 'some small fixes' #215

langner:
Thanks! That will make things much easier.

PR #216, opened 2015-12-05
Turbomole parser & basic tests, only for vibs

mwykes:
Currently only vibrational properties are parsed and tested with the basic dvb_ir test.


langner:
Could you include the input file as we do for other programs?

Also - we do have a branch for the turbomole parser which was never included in any release. AFAIK the main problem was that Turbomole writes a number of files with things we wanted to read and we had to merge them to make any sense of it all. Maybe @ATenderholt could shed some more light on this as he was the main author there.

Perhaps with the changes in cclib since then (reading multiple files is supported now) and new major version of Turbomole support would be feasible. What I will do for now is update the turbomoleparser branch with the current master so we can see if merging with your changes makes any sense.

langner:
I have updated the turbomoleparser branch in my repository:
https://github.com/langner/cclib/tree/turbomoleparser

At first glance it seems it would be possible to combine this code with yours, which would give us a more complete parser. Could you perhaps rerun the tests we have in that branch in Turbomole6?

mwykes:
Agreed that it makes sense to try to get the old turbomole branch working and merged into the main cclib. However, after a quick look and a few changes, I could not get the old parser to work on your updated turbomole and unfortunately I can't dedicate any more time to it right now. I'd like to get my other changes not related to the turbomole parser merged into cclib though, so will continue with further  pull requests bringing the other changes in, and hopefully get back to the turbomole parser when I can spend more time on it.

langner:
OK - I'll try to get this to work soon. Would be a shame to loose that code.

However - would you be willing to rerun the test jobs we have on that branch in a newer Turbomole version? That would be really useful.

mwykes:
Yes, sure, I can rerun those tests with a newer version of Turbomole. I'll make a pull request once I have them.

PR #223, opened 2016-01-06
DALTON: expand parser capabilities and test coverage

berquist:
I didn't realize that deleting a remote branch also deletes any corresponding pull requests. For reference, here is the original: https://github.com/cclib/cclib/pull/198/

According to my list, we're currently missing:

        # aonames
        # aooverlaps
        # atomcharges
        # atomspins
        # coreelectrons
        # enthalpy
        # entropy
        # etoscs
        # etrotats
        # freeenergy
        # grads
        # hessian
        # mocoeffs
        # nocoeffs
        # nooccnos
        # scancoords
        # scanenergies
        # scannames
        # scanparm
        # temperature
        # vibanharms

Which of these should be considered essential? The only ones I can't figure out right now are `etoscs` and `etrotats`.


langner:
> Which of these should be considered essential? The only ones I can't figure out right now are etoscs and etrotats.

I'd say skip tests for whatever attributes are missing and get this merged in.

PR #224, opened 2016-01-07
Implement and test optstatus for Gaussian

langner:


src/cclib/parser/gaussianparser.py
```diff
@@ -213,14 +214,24 @@ def extract(self, inputfile, line):
 
             if not hasattr(self, 'optdone'):
                 self.optdone = []
-
             self.optdone.append(len(self.geovalues) - 1)
 
+            if not hasattr(self, 'optstatus'):
+                self.optstatus = []
+            assert len(self.optstatus) > 0
```
ATenderholt: Shouldn't this throw an assertion if line 220 is ever called (i.e. ```optstatus``` isn't created until now)? Same with line 232.

ATenderholt:
```testScan.py``` also has checks for ```optdone```, so should be updated for ```optstatus```.

src/cclib/parser/gaussianparser.py
langner: > Shouldn't this throw an assertion if line 220 is ever called (i.e. optstatus isn't created until now)? Same with line 232.

Quite right! Updated.

langner:
> testScan.py also has checks for optdone, so should be updated for optstatus.

Done.

PR #226, opened 2016-01-11
gamessukparser: Bug fix, convert atomcoord to bohr

mwykes:
Example of bug: basicGAMESS-UK8.0/dvb_ir.out atomcoord were parsed in bohr before this fix, now they are correctly converted in to angstroms. 
As requested (see discussion of cclib-data pull request 17) I have added a regression test to cclib-data, for which I will shortly make a pull request. I have also removed the unit test testing atomic coordinates in testvib.py (which was in cclib pull request #214, commit 2bec113bf3d0227a7a47dd2f891b88099b50eaf5).


langner:
This is perfect. Thanks!

PR #229, opened 2016-01-15
Test atomcoords units in both SP and GeoOpt test

langner:
This is a folllow-up to cclib/cclib#226. The check would not have passed for GAMESS-UK before that fix.

### DIFF
--- a/test/data/common.py
+++ b/test/data/common.py
@@ -1,0 +1,31 @@
+# This file is part of cclib (http://cclib.github.io), a library for parsing
+# and interpreting the results of computational chemistry packages.
+#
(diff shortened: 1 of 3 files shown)


berquist:
Why check for the minimum carbon separation rather than some text in output files that tells us whether the units are in bohr or angstroms?

langner:
We can do that as well, of course. Nonetheless, the minimum carbon separation should always be correct. I think that, due to this, my attitude is changing and I think it's OK to have redundant tests. The more tests the better.

berquist:
Yes, I agree on both points.

Issue #231, opened 2016-01-22
CJSON writer fails


berquist:
I don't know how this didn't come up sooner, but:

```python
data.write(outputtype='cjson')
```

results in:

```python
/media/Backup/virtualenvs/3/cclib/lib/python3.5/site-packages/cclib/writer/cjsonwriter.py in generate_repr(self)
     75
     76         cjson_dict['atomCount'] = len(self.ccdata.atomnos)
---> 77         cjson_dict['heavyAtomCount'] = len((x for x in self.ccdata.atomnos if x > 1))
     78
     79         if has_openbabel:

TypeError: object of type 'generator' has no len()
```

PR #232, opened 2016-01-23
CJSON writer: can't take len() of a generator.

berquist:
Also:
* Treat Python 2 as a special case, not 3.
* spelling

See https://github.com/cclib/cclib/issues/231.


langner:
Looks good!

PR #234, opened 2016-01-26
ccData: add convenience write methods for various formats

langner:


src/cclib/parser/data.py
```diff
@@ -280,6 +280,16 @@ def write(self, filename=None, *args, **kwargs):
         outputstr = ccwrite(self, outputdest=filename, *args, **kwargs)
         return outputstr
 
+    def writejson(self, filename=None):
+        """Write parsed attributes to a JSON file."""
+        return self.write(filename=filename, outputtype='cjson')
+    def writejson(self, filename=None):
```
berquist: This should probably be "writecml".

langner: Yes, of course. Fixed.

PR #236, opened 2016-01-26
Q-Chem: fix charge/mult and fragment/supsersystem parsing

berquist:
This fixes a bug where two-part calculations (like a frequency calculation after a geometry optimization) failed, because a shortcut for using the last geometry from the previous job was used:

```
$molecule
0 1
<coordinates>
$end

@@@@

$molecule
read
$end
```

and the charge/multiplicity were read from the echoed user input. This bugfix calculates the charge and multiplicity from the number of electrons and the atomic numbers, since the aren't echoed anywhere else.

However, for this to work, as a side effect, I had to finally fix the fragment parsing too so that even if individual fragment sections (SCF, MP/CC, TDDFT, etc.) are echoed back, they're ignored.

See https://github.com/cclib/cclib-data/pull/20 for the regression tests, which needed to be altered so that fragment calculations only considered the supersystem.

Unfortunately, GitHub thinks I've changed the entire file because of their diff settings; `git diff --color-words --ignore-all-space` will show a much better diff. This is because I added a "guard" check at the top of the `extract()` block and indented everything below it a level.

### DIFF
--- a/src/cclib/parser/qchemparser.py
+++ b/src/cclib/parser/qchemparser.py
@@ -50,7 +50,7 @@ class QChem(logfileparser.Logfile):
 
         # Keep track of whether or not this is a fragment calculation,
         # so that only the supersystem is parsed.
-        self.is_fragment_calculation = False
+        self.is_fragment_section = False
 
         # Compile the dashes-and-or-spaces-only regex.
         self.re_dashes_and_spaces = re.compile('^[\s-]+$')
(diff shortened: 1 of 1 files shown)


berquist:
Here is the diff (`git diff --ignore-all-space origin/master`), in case one can't grab my branch.

```
diff --git a/src/cclib/parser/qchemparser.py b/src/cclib/parser/qchemparser.py
index 8ef4b3b..cd2953d 100644
--- a/src/cclib/parser/qchemparser.py
+++ b/src/cclib/parser/qchemparser.py
@@ -3,7 +3,7 @@
 # This file is part of cclib (http://cclib.github.io), a library for parsing
 # and interpreting the results of computational chemistry packages.
 #
-# Copyright (C) 2014-2015, the cclib development team
+# Copyright (C) 2014-2016, the cclib development team
 #
 # The library is free software, distributed under the terms of
 # the GNU Lesser General Public version 2.1 or later. You should have
@@ -48,6 +48,10 @@ class QChem(logfileparser.Logfile):
         self.unrestricted = False
         self.is_rohf = False
 
+        # Keep track of whether or not this is a fragment calculation,
+        # so that only the supersystem is parsed.
+        self.is_fragment_section = False
+
         # Compile the dashes-and-or-spaces-only regex.
         self.re_dashes_and_spaces = re.compile('^[\s-]+$')
 
@@ -130,6 +134,22 @@ class QChem(logfileparser.Logfile):
 
+        # Disable/enable parsing for fragment sections.
+        fragment_section_headers = (
+            'Guess MOs from converged MOs on fragments',
+            'CP correction for fragment',
+        )
+        supersystem_section_headers = (
+            'Done with SCF on isolated fragments',
+            'Done with counterpoise correction on fragments',
+        )
+
+        if any(message in line for message in fragment_section_headers):
+            self.is_fragment_section = True
+        if any(message in line for message in supersystem_section_headers):
+            self.is_fragment_section = False
+
+        if not self.is_fragment_section:
             # If the input section is repeated back, parse the $rem and
             # $molecule sections.
             if line[0:11] == 'User input:':
@@ -150,17 +170,6 @@ class QChem(logfileparser.Logfile):
                                     self.norbdisp_beta_aonames = norbdisp_aonames
                                     self.norbdisp_set = True
 
-                # Charge and multiplicity are present in the input file, which is generally
-                # printed once at the beginning. However, it is also prined for fragment
-                # calculations, so make sure we parse only the first occurance.
-                if '$molecule' in line:
-                    line = next(inputfile)
-                    charge, mult = map(int, line.split())
-                    if not hasattr(self, 'charge'):
-                        self.set_attribute('charge', charge)
-                    if not hasattr(self, 'mult'):
-                        self.set_attribute('mult', mult)
-
                     line = next(inputfile)
 
             # Parse the general basis for `gbasis`, in the style used by
@@ -247,7 +256,9 @@ class QChem(logfileparser.Logfile):
 
                 self.atomcoords.append(atomcoords)
 
-            if not hasattr(self, 'atomnos'):
+                # We calculate and handle atomnos no matter what, since in
+                # the case of fragment calculations the atoms may change,
+                # along with the charge and spin multiplicity.
                 self.atomnos = []
                 self.atomelements = []
                 for atomelement in atomelements:
@@ -263,16 +274,25 @@ class QChem(logfileparser.Logfile):
             # Number of electrons.
             # Useful for determining the number of occupied/virtual orbitals.
             if 'Nuclear Repulsion Energy' in line:
-            if not hasattr(self, 'nalpha'):
                 line = next(inputfile)
                 nelec_re_string = 'There are(\s+[0-9]+) alpha and(\s+[0-9]+) beta electrons'
                 match = re.findall(nelec_re_string, line.strip())
-                self.nalpha = int(match[0][0].strip())
-                self.nbeta = int(match[0][1].strip())
+                self.set_attribute('nalpha', int(match[0][0].strip()))
+                self.set_attribute('nbeta', int(match[0][1].strip()))
                 self.norbdisp_alpha += self.nalpha
                 self.norbdisp_alpha_aonames += self.nalpha
                 self.norbdisp_beta += self.nbeta
                 self.norbdisp_beta_aonames += self.nbeta
+                # Calculate the spin multiplicity (2S + 1), where S is the
+                # total spin of the system.
+                S = (self.nalpha - self.nbeta) / 2
+                mult = int(2 * S + 1)
+                self.set_attribute('mult', mult)
+                # Calculate the molecular charge as the difference between
+                # the atomic numbers and the number of electrons.
+                if hasattr(self, 'atomnos'):
+                    charge = sum(self.atomnos) - (self.nalpha + self.nbeta)
+                    self.set_attribute('charge', charge)
 
             # Number of basis functions.
             # Because Q-Chem's integral recursion scheme is defined using
```

langner:
Finally had time to got through this, and wasn't as much as I thought. Looks good!

PR #242, opened 2016-02-25
Psi4: latest version changes header line

berquist:
The first element block also includes an extra column with atomic masses.

See https://github.com/cclib/cclib-data/pull/22.


langner:
Thanks!

PR #244, opened 2016-03-04
Parse gbasis for ORCA

felixplasser:
Some code to parse gbasis for ORCA. As discussed on the mailing list, this seems to work properly.


berquist:
@langner, I'll merge this and add the proper lines for tests, which do pass.

Issue #243, opened 2016-02-29
Gaussian 09 log file crash


ghutchis:
The attached file crashes Git master:

>     File "./energies", line 16, in <module>
    molecule = file.parse()
  File "/usr/local/anaconda/lib/python2.7/site-packages/cclib/parser/logfileparser.py", line 309, in parse
    self.extract(inputfile, line)
  File "/usr/local/anaconda/lib/python2.7/site-packages/cclib/parser/gaussianparser.py", line 802, in extract
    assert HOMO == self.homos[0]

[100.g09.zip](https://github.com/cclib/cclib/files/151549/100.g09.zip)

ghutchis:
ah  - it's a multi-step job.

langner:
Yeah, cclib still has problems with multi step jobs. Would be good to at
least not fail. Can we use your file as a regression test?

ghutchis:
Absolutely you can use this.

ghutchis:
My suggestion would be to add a test for " Initial command" and reset the data at that point. (I'm not sure whether there's an easy way to reset the parsed data.)

The downside is that you lose anything except the final job. I think it's still a better approach than failing.

If the approach sounds good, I'll submit a patch.

langner:
That has been our approach (parse just the largest, usually last,
system/job) until we support multi-job outputs consistently. Your patch
would be very much appreciated. Let me know if you need any help.

PR #246, opened 2016-03-09
Fix crash on multi-job Gaussian output with different orbitals #243.

ghutchis:
Instead of an assertion failure if the HOMO doesn't match, simply
reset the HOMOs.

Fixes issue #243


ghutchis:
I don't know how to add the file as a regression test..

langner:
We keep them in a separate repo (cclib/cclib-data) since they take up quite some space. I'll take care of that in a moment.

PR #249, opened 2016-03-16
add the CSX write support for DALTON

bwang2453:
This commit will allow the cclib write out the CSX file so that it could be published on the chemical semantics portal.


src/cclib/parser/data.py
```diff
@@ -58,10 +61,13 @@ class ccData(object):
         natom -- number of atoms (integer)
         nbasis -- number of basis functions (integer)
         nmo -- number of molecular orbitals (integer)
+        nmriso -- NMR isotropic chemical shieldings (arragy[1], ppm)
+        nmranis -- NMR anisotropic chemical shieldings (arragy[1], ppm)
         nocoeffs -- natural orbital coefficients (array[2])
         nooccnos -- natural orbital occupation numbers (array[1])
         optdone -- flags whether an optimization has converged (Boolean)
         optstatus -- optimization status for each set of atomic coordinates (array[1])
+        package -- name of the quantum chemistry package (string)
```
langner: This is superfluous - use the `logname` attribute.

src/cclib/parser/data.py
```diff
@@ -70,6 +76,8 @@ class ccData(object):
         scftargets -- targets for convergence of the SCF (array[2])
         scfvalues -- current values for convergence of the SCF (list of arrays[2])
         temperature -- temperature used for Thermochemistry (float, kelvin)
+        theory -- name of the quantum chemistry methodology (string)
+        version -- the version number for quantum chemistry package (string)
```
langner: Attributes should be results of the calculations. We definitely don't want to create attributes for specific metadata elements.

src/cclib/parser/data.py
```diff
@@ -28,6 +28,7 @@ class ccData(object):
         atommasses -- atom masses (array[1], daltons)
         atomnos -- atomic numbers (array[1])
         atomspins -- atomic spin densities (dict of arrays[1])
+        basisname -- name of the standard basis set (string)
```
langner: We already have `gbasis` which specifies the basis set numerically. A string is a very vague way of specifying the basis set. We might want this information for convenience, but definitely not as a separate attribute.

src/cclib/parser/data.py
```diff
@@ -43,11 +44,13 @@ class ccData(object):
         fooverlaps -- fragment orbital overlap matrix (array[2])
         fragnames -- names of fragments (list of strings)
         frags -- indices of atoms in a fragment (list of lists)
+        functional -- name of DFT functionals (string)
         gbasis -- coefficients and exponents of Gaussian basis functions (PyQuante format)
         geotargets -- targets for convergence of geometry optimization (array[1])
         geovalues -- current values for convergence of geometry optmization (array[1])
         grads -- current values of forces (gradients) in geometry optimization (array[3])
         hessian -- elements of the force constant matrix (array[1])
+        hofenergies -- heat of formation after semiempirical calculation (array[1], eV)
```
langner: This is not implemented. Besides, wouldn't this be covered by `enthalpy`?

src/cclib/parser/daltonparser.py
```diff
@@ -29,6 +30,7 @@ def __init__(self, *args, **kwargs):
 
         # Call the __init__ method of the superclass
         super(DALTON, self).__init__(logname="DALTON", *args, **kwargs)
+        self.package = "DALTON"
```
langner: Superfluous.

src/cclib/parser/daltonparser.py
```diff
@@ -166,7 +174,8 @@ def extract(self, inputfile, line):
             symmetry_atoms = []
             atommasses = []
             for cols in lines:
-                atomnos.append(self.table.number[cols[0]])
+                cols0 = ''.join([i for i in cols[0] if not i.isdigit()]) #remove numbers
```
langner: If there are numbers here that need to be remove and crash the parser, could you please add a regression and move this to a separate pull request?

langner:
The introduction of `basisname`, `package` and `version` here is a no-go for me. We definitely don't want to create attributes for such things. If we want to parse these, however, perhaps we should create something called `metadata` which would be a dictionary or object and would gather such information. What do others think?

berquist:
> If we want to parse these, however, perhaps we should create something called metadata which would be a dictionary or object and would gather such information. What do others think?

I think a lot of this would be cleared up by the creation of an input file parser. There might still be some metadata-type attributes that it wouldn't catch (such as `version`), but things like the method and basis set name (the latter of which I'd like to have) would be covered in some way by an input file parser.

langner:
> I think a lot of this would be cleared up by the creation of an input file parser. There might still be some metadata-type attributes that it wouldn't catch (such as version), but things like the method and basis set name (the latter of which I'd like to have) would be covered in some way by an input file parser.

Yes. We've been resistant to parsing/writing input files in the past, but I'm not opposed to it in principle. I think this is something we should plan for in a 2.x version, since it might involve some major changes to the API.

In the meantime, @berquist, do you think an attribute gathering metadata would be appropriate?

bwang2453:
I think it is a good idea to gather "package" and "version" into metadata, but "basisname" could be an attribute for convenience. If we are on the same page, I'll go ahead to modify the code according to it.

langner:
The trouble with basis set names is that they can mean different things in different programs.

src/cclib/parser/data.py
bwang2453: I'll remove it for Dalton. It is for semiempirical methods and is different from ethalpy. Some packages, such as Gamess and Gaussian, have the semiempirical methods.

src/cclib/parser/daltonparser.py
bwang2453: I didn't touch this line

bwang2453:
I have generated the metadata attribute, which currently includes "package", "version", "theory", "basisname", "spintype" and "functional".

src/cclib/parser/daltonparser.py
langner: OK. Maybe they were added to cclib/master in the meantime... could you rebase your pull request?

langner:
I like where this is going, but i think we really need to break this up further into three things:
1. Parsing metadata/input as opposed to output data of the calculation. I also created #260 for step size which probably also falls into this category. Whether that would entail a `metadata` attributes or `input` or something else remains to be seen.
2. Parsing the above for DALTON and other programs
3. Adding a CSX writer, independent of DALTON and other programs.

In any case, pushing this back to v1.5

langner:
After chatting with Bing, we decided to shoot for a v1.4.1 release as soon as we can get the CSX output support in master. We'll also include any other interesting bits that get in until then.

langner:
Note: I just noticed the earliest commits here are from before even v1.4, which means they might show up when someone uses that tag and break something. In case that happens we'll need to cherry-pick out those changes on a separate branch.

langner:
Closing as per @bwang2453 who will be separating issues into separate PRs.

dln22:
> After chatting with Bing, we decided to shoot for a v1.4.1 release as soon as we can get the CSX output support in master. We'll also include any other interesting bits that get in until then.

I know this is quite old, but has the CSX output support been integrated after all? I can not find this option in the most recent release.

berquist:
@dln22 No, it was never merged.  Only certain necessary pieces to support it were added.  I think `metadata` was created as a result.

Issue #206, opened 2015-08-10
Q-Chem: `StopIteration` is raised for incomplete SCF


berquist:
Any time an SCF is incomplete (*not* unconverged), the parser doesn't handle it gracefully:

```
~/calc.sgr/krista/batch2/jobs $ l *out
/home/dlambrecht/erb74/calc.sgr/krista/batch2/jobs
-rw-r----- 1 erb74 dlambrecht 825K Aug 10 19:02 CO2_BMIM_PF6.b3lyp.vqz.opt.out
-rw-r----- 1 erb74 dlambrecht 1.5M Aug 10 19:07 CO2_BMIM_TFA.blyp.avtz.opt.out
~/calc.sgr/krista/batch2/jobs $ tail *out
==> CO2_BMIM_PF6.b3lyp.vqz.opt.out <==
 -------------------------------------------------------
 OpenMP BLAS3 based DFT computing Module                
 Release: version 1.0, May 2013, Q-Chem Inc. Pittsburgh 
 -------------------------------------------------------
 ---------------------------------------
  Cycle       Energy         DIIS Error
 ---------------------------------------
    1   -1553.2043340273      1.36E-05
    2   -1553.2044609529      7.13E-07
    3   -1553.2044610939      5.17E-07

==> CO2_BMIM_TFA.blyp.avtz.opt.out <==
Setting SaveMinima to 0
 using 8 threads for integral computing 
 -------------------------------------------------------
 OpenMP BLAS3 based DFT computing Module                
 Release: version 1.0, May 2013, Q-Chem Inc. Pittsburgh 
 -------------------------------------------------------
 ---------------------------------------
  Cycle       Energy         DIIS Error
 ---------------------------------------
    1   -1138.3465451227      8.10E-05
~/calc.sgr/krista/batch2/jobs $ cclib_plot_geometry_convergence.py *.out
Traceback (most recent call last):
  File "/home/dlambrecht/erb74/scripts/cclib_plot_geometry_convergence.py", line 76, in <module>
    main()
  File "/home/dlambrecht/erb74/scripts/cclib_plot_geometry_convergence.py", line 38, in main
    data = job.parse()
  File "/home/dlambrecht/software/python/anaconda3/lib/python3.4/site-packages/cclib/parser/logfileparser.py", line 247, in parse
    self.extract(inputfile, line)
  File "/home/dlambrecht/software/python/anaconda3/lib/python3.4/site-packages/cclib/parser/qchemparser.py", line 347, in extract
[...]

PR #250, opened 2016-03-18
Test for StopIteration during incomplete SCF

berquist:
See https://github.com/cclib/cclib/issues/206 for the original issue and https://github.com/cclib/cclib-data/pull/21 for the regression tests.

This handles the `StopIteration` exception that occurs if the SCF is terminated early. Currently implemented for at least one version of all parsers except ADF, since I'm not sure of the ideal place to "cut" the `dvb_sp.adfout` file, and there are multiple versions of the file as well.

### DIFF
--- a/src/cclib/parser/gamessukparser.py
+++ b/src/cclib/parser/gamessukparser.py
@@ -335,10 +335,14 @@ class GAMESSUK(logfileparser.Logfile):
             scfvalues = []
             line = next(inputfile)
             while line.strip():
+                # e.g. **** recalulation of fock matrix on iteration  4 (examples/chap12/pyridine.out)
                 if line[2:6] != "****":
-            # e.g. **** recalulation of fock matrix on iteration  4 (examples/chap12/pyridine.out)
                     scfvalues.append([float(line[tester-5:tester+6])])
-                line = next(inputfile)
+                try:
+                    line = next(inputfile)
+                except StopIteration:
+                    self.logger.warning('File terminated before end of last SCF! Last tester: {}'.format(line.split()[5]))
+                    break
             self.scfvalues.append(scfvalues)
 
         if line[10:22] == "total energy" and len(line.split()) == 3:
--- a/src/cclib/parser/daltonparser.py
+++ b/src/cclib/parser/daltonparser.py
@@ -581,7 +581,11 @@ class DALTON(logfileparser.Logfile):
 
             while not converged:
 
-                line = next(inputfile)
+                try:
+                    line = next(inputfile)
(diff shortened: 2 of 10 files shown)


src/cclib/parser/gamessukparser.py
```diff
@@ -335,10 +335,14 @@ def extract(self, inputfile, line):
             scfvalues = []
             line = next(inputfile)
             while line.strip():
+                # e.g. **** recalulation of fock matrix on iteration  4 (examples/chap12/pyridine.out)
```
langner: typo?

langner:
Do we have regression files for the other parsers you modified?

berquist:
Yes, they're all in https://github.com/cclib/cclib-data/pull/21.

src/cclib/parser/gamessukparser.py
berquist: It looks like I moved the comment before the line it's referencing. I don't know what the bit in parentheses at the end is, though.

langner:
Looks good. In the future we might want to wrap `next(inputfile)` in something that will handle `StopIteration` gracefully more generally. I can imagine the same issue surfacing in other parts of the parsers.

langner:
Fixes #206

baoilleach:
Why not just add a try except to the caller?

berquist:
> Why not just add a try except to the caller?

Could I do this by implementing `__next__` in `logfileparser.Logfile`? Or is it `logfileparser.FileWrapper`? Looks like the former, since `inputfile` is an instance of `Logfile`, correct me if I'm wrong.

I'd really like to have this, because ultimately it's impossible to add this check everywhere it's necessary for every parser, and a more generic warning could be given than `StopIteration`. For example, if a Q-Chem job dies in the middle of a gradient calculation, it gets raised.

A keyword argument(s) could even be used to give something more specific:
```
# parsing inside SCF
line = next(inputfile, section='SCF')
# or
line = next(inputfile, warning='SCF iterations ended unexpectedly')
# Python 2 would work the same way?
line = inputfile.next(section='SCF')
```
and the global logger instance would be called.

langner:
I'm not sure what Noel meant by caller exactly, and I don't have any implementation preferences. As long as whatever we do generalizes this behavior, I'll be happy.

baoilleach:
I was suggesting that you wrap the whole of the parsing code in a try
except that catches StopIteration and ignores it. This could easily be done
in the caller, i.e. logfileparser.parse. Though I think Eric is suggesting
that the logger reports in which section the parsing failed.

- Noel

PR #259, opened 2016-04-08
Add formatted checkpoints for all test files.

ghutchis:
I added G09-generated fchk files for each of the Gaussian test files.


ATenderholt:
I noticed that some of the files (e.g. ```dvb_sp.fchk```) are adding just 0 lines. Did something go wrong when converting from its original checkpoint file?

langner:
This is cool, thanks! I am wondering whether we want to add these to the library or to cclib-data. Any thoughts?

ghutchis:
@langner Considering there's interest in an fchk parser, I'd have at least some in the library.

@ATenderholt - I just noticed that some had geom=allcheck in the calculation. I'll re-parse the input file so there's something close-to-correct. (I can't guarantee the exact run since I don't have the original checkpoints.)

ghutchis:
BTW, if there's concern about the size, I can gzip them all as a compromise

langner:
My first thought was to keep them all here, but then thought maybe just for one or two unit tests. In the end it's not a lot of space so we can keep all. We don't distribute the files with releases anyway. Also, git will already compress them so there's no point in doing that.

How much are all the fchks togethre?

ghutchis:
It's only 1.8MB for the fchk files.

langner:
I guess we might end up adding *.out files for GAMESS-US and similar formats for other programs. But if they're all similar size to this then it should be fine.

PR #269, opened 2016-04-22
Q-Chem: fix for '****' in geometry optimizations

berquist:
Regression fix in https://github.com/cclib/cclib-data/pull/31.


langner:
Do you want this to go into 1.4?

PR #273, opened 2016-04-27
add metadata for Dalton

bwang2453:
Add the metadata attribute, which currently includes `package`, `version`, `theory`, `basisname`, `spintype` and `functional` and parse those for Dalton


src/cclib/parser/daltonparser.py
```diff
@@ -222,7 +232,11 @@ def extract(self, inputfile, line):
             assert "Total number of atoms:" in line
             self.set_attribute("natom", int(line.split()[-1]))
 
-            self.skip_lines(inputfile, ['b', 'basisname', 'b'])
+            #self.skip_lines(inputfile, ['b', 'basisname', 'b'])
+            line = next(inputfile)
+            line = next(inputfile)
+            self.metadata["basisname"] = re.findall(r'"([^"]*)"', line)[0]
```
langner: Maybe `basis_set_name` is more precise.

src/cclib/parser/daltonparser.py
```diff
@@ -79,6 +83,12 @@ def parse_geometry(self, lines):
 
     def extract(self, inputfile, line):
         """Extract information from the file object inputfile."""
+        # extract the version number first
+        if line[4:30] == "This is output from DALTON":
+            if line.split()[5] == "release" or line.split()[5] == "(Release":
+                self.metadata["version"] = line.split()[6][6:]
```
langner: Perhaps `program_version` is better?

src/cclib/parser/daltonparser.py
```diff
@@ -501,6 +515,9 @@ def extract(self, inputfile, line):
             self.set_attribute("charge", int(line.split()[-1]))
         if "@    Spin multiplicity and 2 M_S" in line:
             self.set_attribute("mult", int(line.split()[-2]))
+        # Dalton only has ROHF, no UHF
+            if self.mult != 1:
+                self.metadata["spintype"] = "ROHF"
```
langner: ROHF is a computation method, so `spintype` is probably not the best choice.

src/cclib/parser/daltonparser.py
```diff
@@ -737,20 +754,36 @@ def extract(self, inputfile, line):
         # @    Final gradient norm:           0.000003746706
         # ...
         #
+        if "Final HF energy" in line and not (hasattr(self, "mpenergies") or hasattr(self, "ccenergies")):
+            self.metadata["theory"] = "HF"
```
langner: I don't like `theory` as a name here, all this is within the same theory of electronic structure. Perhaps `method` would be a good choice. In that case, I would not replace HF with MP2 or something else. MP/CC are post-Hartree-Fock methods that build on top of HF, so perhaps use a different entry?

src/cclib/parser/daltonparser.py
```diff
@@ -737,20 +754,36 @@ def extract(self, inputfile, line):
...
 
+        if "Total CC2   energy:" in line:
+            self.metadata["theory"] = "CCSD"
+            energ = utils.convertor(float(line.split()[-1]), 'hartree', 'eV')
+            if not hasattr(self, "ccenergies"):
+                self.ccenergies = []
+            self.ccenergies.append(energ)
+
         if "Total energy CCSD(T)" in line:
+            self.metadata["theory"] = "CCSD-T"
```
berquist: Is there a reason why we can't use "CCSD(T)" here?

berquist:
I'm fine with this.

src/cclib/parser/daltonparser.py
bwang2453: I'll change it

src/cclib/parser/daltonparser.py
bwang2453: Agree!

src/cclib/parser/daltonparser.py
bwang2453: I think spin_type is more specific. Only three values are allowed: RHF, ROHF, UHF

src/cclib/parser/daltonparser.py
bwang2453: How about "methodology"? It's useful to separate the HF and post-HF methods, but I would do it on the writeCSX.

src/cclib/parser/daltonparser.py
bwang2453: CSX is in XML format now, which doesn't allow parentheses in the element. I know it's annoying. In the future, CSX might use JSON-LD serialization. Then we can CCSD(T)

src/cclib/parser/daltonparser.py
langner: Allowed by whom?

src/cclib/parser/daltonparser.py
langner: I would prefer to have both in cclib, since MP2 can follow other things than HF.

src/cclib/parser/daltonparser.py
langner: Can we use CCSD(T) and have the writer mangle on write?

src/cclib/parser/daltonparser.py
bwang2453: allowed by CSX schema

src/cclib/parser/daltonparser.py
bwang2453: This term is used to identify the computational method conducted in a calculation. It can be semiempirical methods such as "AM1", "PM3" , or multi-configuration methods such as  "CASPT2", "GVB", etc.  I don't think we need to go deep into its theoretical background.

src/cclib/parser/daltonparser.py
bwang2453: OK, I can do that

src/cclib/parser/daltonparser.py
langner: I don't think it's a good idea for cclib attributes to be constrained by CSX. How about we just have one fields called `method` and put all the methods used in the job in there? This would include both Hartree-Fock and post-HF methods. For example:
>>> print data1.metadata['methods']
>>> ['RHF', 'MP2', 'MP3']
>>> print data2.metadata['methods']
>>> ['UHF', 'CCSD(T)']

This would be more inclusive and extendable than creating a multitude of fields for different types of methods.

@berquist any thoughts?

src/cclib/parser/daltonparser.py
langner: Can we include all of these in a `methods` field? I don't think we benefit from any structure here, since the `metadata` dict will not be correlated with any other attributes. However, having a list of methods used you can still output CSX by checking what's inside.

src/cclib/parser/daltonparser.py
ghutchis: I'd suggest some sort of attribute for R, RO, U that's distinct from method. Some programs, for example, call it ROB3LYP or UB3LYP. If not spin type, then something like "open shell"?

But I definitely agree "method" should include all the methods. Consider a composite method like G3 or other thermochemical recipe - there will be multiple methods.

langner: That would be fine, since these are two more or less independent things. That might be a nice thing to call `spintype` (although `open_shell` is fine, too).

bwang2453:
Once everyone is happy about it, please merge this PR as soon as possible. I'll add metadata to other packages.

src/cclib/parser/daltonparser.py
berquist: I'm in agreement with both @langner and @ghutchis. I think it's sufficient to have `self.metadata['methods']` as a list, where only the method and no reference to the spin restriction is made, and then have a boolean for the spin restriction. For example, `self.metadata['unrestricted'] == False` and `len(self.homos == 2)` would indicate restricted open-shell, and a small method would convert this into `"ROHF"`.

src/cclib/parser/daltonparser.py
```diff
@@ -764,9 +770,9 @@ def extract(self, inputfile, line):
         # ...
         #
         if "Final HF energy" in line and not (hasattr(self, "mpenergies") or hasattr(self, "ccenergies")):
-            self.metadata["theory"] = "HF"
+            self.metadata["methods"] = "HF"
         if "Final DFT energy" in line:
-            self.metadata["theory"] = "DFT"
+            self.metadata["methods"] = "DFT"
```
berquist: Is there any support in the CSX format for identifying different functionals?

bwang2453: The metadata has an item called "functional" to be used for exacting the name of the DFT functional. Once the DFT data is published on our portal, it could identify this functional is GGA or hybrid one, or it has dispersion correction or not.

src/cclib/parser/daltonparser.py
langner: To me this also shows how `metadata` can be used as a staging ground for possible attributes, since `unrestricted` is perhaps something we might like to be an attribute(?). Anyway, I like this approach.

src/cclib/parser/daltonparser.py
```diff
@@ -237,7 +247,11 @@ def extract(self, inputfile, line):
             if not self.basislibrary:
                 self.skip_line(inputfile, 'b')
             else:
-                self.skip_lines(inputfile, ['b', 'basisname', 'b'])
+                #self.skip_lines(inputfile, ['b', 'basisname', 'b'])
+                line = next(inputfile)
+                line = next(inputfile)
+                self.metadata["basis_set_name"] = re.findall(r'"([^"]*)"', line)[0]
```
langner: I'm doubtful we need a regex here... can this be done simpler?

src/cclib/parser/daltonparser.py
```diff
@@ -752,20 +769,36 @@ def extract(self, inputfile, line):
...
 
         if "@   = MP2 second order energy" in line:
+            self.metadata["methods"] = "MP2"
             energ = utils.convertor(float(line.split()[-1]), 'hartree', 'eV')
             if not hasattr(self, "mpenergies"):
                 self.mpenergies = []
             self.mpenergies.append([])
             self.mpenergies[-1].append(energ)
 
+        if "Total CC2   energy:" in line:
```
langner: This is a case we didn't catch before. Thanks! Do you have an example file we could use as a regression?

langner:
Overall, I think this one is almost ready to go, with minor changes. @bwang2453 will you have time this week or next to follow up on the minor comments? I can help out.

Also, it would be great if you could compress the commit history into fewer and more descriptive changes.

src/cclib/parser/daltonparser.py
```diff
@@ -83,6 +87,12 @@ def parse_geometry(self, lines):
 
     def extract(self, inputfile, line):
         """Extract information from the file object inputfile."""
+        # extract the version number first
+        if line[4:30] == "This is output from DALTON":
+            if line.split()[5] == "release" or line.split()[5] == "(Release":
+                self.metadata["program_version"] = line.split()[6][6:]
```
langner: Rename this to simply `version` or `package_version` to be consistent.

src/cclib/parser/daltonparser.py
```diff
@@ -250,7 +249,7 @@ def extract(self, inputfile, line):
                 #self.skip_lines(inputfile, ['b', 'basisname', 'b'])
                 line = next(inputfile)
                 line = next(inputfile)
-                self.metadata["basis_set_name"] = re.findall(r'"([^"]*)"', line)[0]
+                self.metadata["basis_set_name"] = line.split()[4][1:-1]
```
langner: What are trying to remove here, apostrophes or something? If so, please use `line.split()[4].strip(...)` or `.replace(...)`.

src/cclib/parser/daltonparser.py
berquist: It needs to be changed to `Total CCSD`; CC2 is an approximation to CCSD and isn't the same thing.

langner:
Pushed back to 1.5

langner:
Please see bwang2453/cclib#1

PR #275, opened 2016-05-04
Parse atombasis for ADF

felixplasser:
"atombasis" is parsed for ADF in a sense that it is consistent with "mocoeffs" and produces correct population analysis data. This only works without symmetry and without fragment DFT. Otherwise the atombasis attribute is deleted.

The data is parsed in a way that the different ordering of the atoms and fragments is taken into account properly

### DIFF
--- a/src/cclib/parser/adfparser.py
+++ b/src/cclib/parser/adfparser.py
@@ -740,6 +740,7 @@ class ADF(logfileparser.Logfile):
 
             self.fonames = []
             self.start_indeces = {}
+            self.atombasis = [[] for frag in self.frags] # parse atombasis in the case of trivial SFOs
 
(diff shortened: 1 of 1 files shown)


langner:
Can you test this?

felixplasser:
What I have is an example for my TheoDORE program where everything looks consistent:
https://sourceforge.net/p/theodore-qc/code/ci/master/tree/EXAMPLES/fa2.adf/QC_FILES/

In that sense I believe that it works correctly

langner:
Thanks! I've confirmed this works for our unit test logfile without symmetry, so I'm merging. However, the logfile with symmetry is what is currently used (this might change) so this won't show up as covered/implemented in the docs. It's still a nice place to start.

PR #276, opened 2016-05-13
email addresses in the README.md have to be in quotes ""

joanwa:
otherwise github assumes this is a webpage. Furthermore added a link to
the mailing list subscription page at sourceforge


langner:
The 2D potential scan already went in as a regression test in cclib/cclib#25 - thanks for the file. I don't think we need the changes in the README file you propose.

joanwa:
did actually anyone ever click on the link? If you click on it it points to

https://github.com/cclib/cclib/blob/master/cclib-users@lists.sourceforge.net <https://github.com/cclib/cclib/blob/master/cclib-users@lists.sourceforge.net>

which obviously can’t exist. Why would anyone want to leave this as is?

langner:
My apologies, you are right. Could modify the PR to just have those changes then?

joanwa:
sure! you mean to just correct the email address link, without adding the link to the mailing list subscription site? And shall I modify it, or you do it? Was not clear from your email ;)

langner:
I meant: could you do it? :D

joanwa:
is changed!

langner:
There's a also a binary (comrpessed log) file there... can you get rid of that?

joanwa:
closing this pull request because of the binary file, opened another one with just README.md updated

PR #277, opened 2016-05-15
Fixes for regression CdtetraM1B3LYP

langner:


langner:
This needs to go in before cclib/cclib-data#33

PR #279, opened 2016-06-04
Implementation of the CJSON writer

Schamnad:
As part of the GSOC2016 project, the first task was to design and implement a standardized cjson format.
The design can be viewed here : https://docs.google.com/document/d/1bgohO_7OljIEEnXIwQmrFYArd8d5KCEapNGdxNWm7-Y/edit?usp=sharing

This spec was used as a starting point for the current implementation. This implementation is able to generate a cjson which has been validated using jsonlint.com

### DIFF
--- a/src/cclib/parser/data.py
+++ b/src/cclib/parser/data.py
@@ -14,6 +14,7 @@
 
 
 import numpy
+from collections import namedtuple
 
 
 class ccData(object):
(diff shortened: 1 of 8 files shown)


langner:
I think this is basically ready to go - I'll merge later today unless someone has further comments.

src/cclib/writer/ccwrite.py
```diff
@@ -28,7 +28,7 @@
 from . import xyzwriter
 
 
-def ccwrite(ccobj, outputtype=None, outputdest=None, returnstr=False,
+def ccwrite(ccobj, outputtype=None, outputdest=None, cjsonterse=False , returnstr=False,
```
langner: Maybe call this simply `terse` in this function which a comment that it only applies to CJSON at this point? Might be useful for other formats in the future.

src/cclib/writer/ccwrite.py
```diff
@@ -67,7 +68,10 @@ def ccwrite(ccobj, outputtype=None, outputdest=None, returnstr=False,
         # Avoid passing multiple times into the main call.
         del kwargs['jobfilename']
 
-    outputobj = outputclass(ccdata, jobfilename=jobfilename, *args, **kwargs)
+    if cjsonterse:
```
langner: Since you are passing kwargs, just add the terse argument to it to avoid two of the same calls. Or change `cjsonterse` into `terse` as noted above and always pass it.

src/cclib/writer/cjsonwriter.py
```diff
@@ -17,97 +17,424 @@
...
+        "etoscs":         'oscillator strength',
+        "etrotats":       'rotatory strength',
+        "etsecs":         'one excited config',
+        "etsyms":         'symmetry',
+        "freeenergy":     'free energy',
+        "fonames":        'orbital names',
+        "fooverlaps":     'orbital overlap',
+        "fragnames":      'names',
+        "frags":          'atom indices',
+        'gbasis':         'TBD',
```
langner: Could you add a comment about what TBD implies here?

src/cclib/writer/cjsonwriter.py
```diff
@@ -17,97 +17,424 @@
...
+import numpy as np
 
 from . import filewriter
 
 
 class CJSON(filewriter.Writer):
     """A writer for chemical JSON (CJSON) files."""
+    
+    # The expected Key names for all supported attributes.
+    _attrkeynames = {
```
langner: This might be something it would make sense to move to parser/data.py and import here. More a note than something to do now - maybe add a TODO comment, we can do this later.

src/cclib/writer/cjsonwriter.py
```diff
@@ -17,97 +17,424 @@
...
         if has_openbabel:
             cjson_dict['smiles'] = self.pbmol.write('smiles')
             cjson_dict['inchi'] = self.pbmol.write('inchi')
             cjson_dict['inchikey'] = self.pbmol.write('inchikey')
             cjson_dict['formula'] = self.pbmol.formula
+            
+        # Incorporate Unit Cell into the chemical JSON
+
+        # Helpers functions which use properties provided by cclib
+        self.generate_properties(cjson_dict)
```
langner: I'm wondering whether this would be easier to maintain if we could separate the structure of the CJSON from the actual generation/writing. Can you think of a data structure that we could use to represent that structure on it's own? Not necessarily to do now, but would be nice.

langner:
A few more comments and questions, if you don't mind :)

src/cclib/writer/ccwrite.py
Schamnad: The 'cjsonterse' argument has been renamed to 'terse' and have gone ahead and included it in the function. Would be useful in the future

src/cclib/writer/ccwrite.py
Schamnad: Have gone with the `terse` option and always pass it

src/cclib/writer/cjsonwriter.py
Schamnad: This has been implemented in this [commit](https://github.com/cclib/cclib/pull/279/commits/b941191fcafe1b6ae37695eefff88c4d50fc3bec)
The data.py file already had a dictionary containing the attribute keys, to avoid redundant code the key:value dictionary has been converted to a key:namedtuple pair. The necessary changes have been made in all files

src/cclib/parser/data.py
```diff
@@ -83,67 +84,68 @@ class ccData(object):
...
+       "etoscs":           Attribute( numpy.ndarray,    'oscillator strength'),
+       "etrotats":         Attribute( numpy.ndarray,    'rotatory strength'),
+       "etsecs":           Attribute( list,             'one excited config'),
+       "etsyms":           Attribute( list,             'symmetry'),
+       "freeenergy":       Attribute( float,            'free energy'),
+       "fonames":          Attribute( list,             'orbital names'),
+       "fooverlaps":       Attribute( numpy.ndarray,    'orbital overlap'),
+       "fragnames":        Attribute( list,             'names'),
+       "frags":            Attribute( list,             'atom indices'),
+       "gbasis'":          Attribute( list,             'TBD'),
```
TDI-tenderholt: Travis CI is failing because of ```gbasis``` when using ```ccData_optdone_bool```. I think that is because there is an extra ```'``` here.

Schamnad: Thank you Adam. I had seen that the Travis CI build was failing, but was in no state to debug it (3 a.m. in the night :) ); had kept the debugging for the morning. I had run the basic sanity testing of checking whether all the file writers were performing as intended, I did not run the whole test suite, will keep this in mind the next time . 
Thanks for helping me out :)

Schamnad:
I've implemented the single dictionary with the namedtuple format, any comments on this implementation.
In the absence of testall.py I am following the test [instructions](https://gist.github.com/berquist/b0c2c5004f1ac25748ce) provided by Eric  but I found that it isn't comprehensive and I miss out on a few tests that Travis CI contains (leading to build failures), is there any way for me to run the whole suite of tests on my local build before I push the code?

berquist:
Sorry I'm late to the discussion, I've been rather busy the past few days. I don't have time to review everything right now, hopefully tomorrow I will. Just wanted to ask:

> it isn't comprehensive and I miss out on a few tests that Travis CI contains (leading to build failures), is there any way for me to run the whole suite of tests on my local build before I push the code?

Can you give me an example? I've been meaning to add this script to the top-level, then have Travis call it to avoid duplication of work, but right now they contain the same Python/shell commands. It could certainly have Python 2/3 compatibility problems, but looking at your last Travis log,

    AttributeError: 'ccData_optdone_bool' object has no attribute '_attrtypes'

should have been caught.

It might be an old bytecode problem, with the \*.pyc files that hang around. Try deleting your `build/` directory, then doing `python setup.py install` again. This *rarely* happens, but it can be really annoying when it does, and I've discovered this during continuous integration too.

Schamnad:
When I ran the tests the code passed all the test cases, but it missed the tests in testScan.py which is where the deprecated _attrtypes reference existed. To implement any change I do to the ccwrite scripts I usually run the following two commands 

1.  python setup.py build
2. sudo python setup.py install
which should take care of any old bytecode problem. I'll verify again in the evening and update the thread

langner:
Two additional points to consider:
1. You can see exactly what Travis is running at https://github.com/cclib/cclib/blob/master/.travis.yml
2. Travis doesn't actually run all tests. It doesn't run the bridge tests, or the method tests.
3. On Linux/Mac, use "grep -R" to find all instances of the string you need to replace.
4. Make sure to run the regression tests as well.

Schamnad:
Thanks for the link Karol, I ran the exact same sequence of commands before the recent commit. It appears that what Eric suggested about old bytecode persisting happened in my case.

I'll make sure to run the tests appropriately the next time around. With regards to the string replacement across the project, it has been completed in the latest commit.

langner:
I'll give Eric a chance to look over this today if he gets the time. I'll merge it in tomorrow if someone doesn't beat me to it.

src/cclib/writer/xyzwriter.py
```diff
@@ -36,7 +36,6 @@ def __init__(self, ccdata, splitfiles=False,
         self.do_lastgeom = lastgeom
         self.do_allgeom = allgeom
 
-        self.generate_repr()
```
berquist: I only have one question before Karol or I merge; if you remove this method call from the writers, how do I use them from Python code?

Schamnad: It's called in the parent script [ccwrite.py](https://github.com/cclib/cclib/blob/master/src/cclib/writer/ccwrite.py#L71) once the filewriter has been instantiated.

berquist: Ah, I see, thank you.

berquist:
I'm getting an error with Python 3, and this works with Python 2. 

```
[eric@copper]$ ccwrite cjson data/Jaguar/basicJaguar8.3/dvb_gopt_hf.out
Attempting to parse data/Jaguar/basicJaguar8.3/dvb_gopt_hf.out
cclib can parse the following attributes from data/Jaguar/basicJaguar8.3/dvb_gopt_hf.out:
  atomcoords
  atomnos
  charge
  coreelectrons
  geotargets
  geovalues
  homos
  moenergies
  mosyms
  mult
  natom
  nbasis
  nmo
  optdone
  scfenergies
  scftargets
  scfvalues
  File "/media/Backup/virtualenvs/3/cclib_berquist/bin/ccwrite", line 93, in <module>
    main()
  File "/media/Backup/virtualenvs/3/cclib_berquist/bin/ccwrite", line 89, in main
    ccwrite(data, outputtype, outputdest, terse, **ccwrite_kwargs)
  File "/media/Backup/virtualenvs/3/cclib_berquist/lib/python3.5/site-packages/cclib/writer/ccwrite.py", line 72, in ccwrite
    output = outputobj.generate_repr()
  File "/media/Backup/virtualenvs/3/cclib_berquist/lib/python3.5/site-packages/cclib/writer/cjsonwriter.py", line 83, in generate_repr
    return json.dumps(cjson_dict, cls=JSONIndentEncoder, sort_keys=True, indent=4)
  File "/usr/lib/python3.5/json/__init__.py", line 237, in dumps
    **kw).encode(obj)
  File "/media/Backup/virtualenvs/3/cclib_berquist/lib/python3.5/site-packages/cclib/writer/cjsonwriter.py", line 370, in encode
    for key, value in o.iteritems():
AttributeError: 'dict' object has no attribute 'iteritems'
```
See [this](http://stackoverflow.com/questions/10458437/what-is-the-difference-between-dict-items-and-dict-iteritems) StackOverflow post for a in-depth discussion. I tend to write the simplest code possible, which would be `dict.items()`, because I'm more focused on Python 3 than 2, but probably not everyone feels this way.

Technically, all the Travis CI tests pass, but this doesn't work for Python 3 from the command line. I already knew this, but we're eventually going to need more test coverage for all the writers. What should we do for now?

Schamnad:
I missed out on due diligence here, I reproduced the error and have rectified it in the latest commit.

In future I think we need to run the Travis CI tests for both Python2 and Python3, python 2.7 is still the default in many linux distributions. A similar bug might be missed .

langner:
We already test multiple Python version in Travis. The problem here is that we don't have sufficient test coverage for the writers.

berquist:
Well, I'm perfectly happy with the way things look; it's a good improvement over what's currently there and I'm pleased to finally see the CJSON specification moving forward.

@langner, do you think it's ok to merge now?

langner:
Yup - let's go!

PR #280, opened 2016-06-13
Updates to built-in methods.

berquist:
* Add fragment spins for population analyses.
* Modify method tests to work with new testing code.
* Run tests using Travis.


test/method/testcda.py
```diff
@@ -11,21 +11,24 @@
...
+import sys
 import os
 import logging
 import unittest
 
 import numpy
 
-from testall import getfile
+sys.path.append("..")
+from test_data import getdatafile
```
berquist: I would welcome a better way of doing this.

langner: Move getfile and getdatafile to the parent dir?

langner:
@berquist is there anything else you want to do here?

test/method/testcda.py
baoilleach: Don't newer pythons support relative imports with a dot notation?

berquist:
I will have to wait until sometime this weekend to take another look at
this.

test/method/testcda.py
berquist: > Don't newer pythons support relative imports with a dot notation?

Yes, but only within modules. If you change this to `from ..test_data import getdatafile` and try `python testcda.py`, this happens:

    SystemError: Parent module '' not loaded, cannot perform relative import

berquist:
Ok, I'm fine with this for now, at least in terms of features.

Issue #228, opened 2016-01-13
PC GAMESS/Firefly parser failes to parse MOs from geometry optimizations?


mkrompiec:
Test case: geometry optimization or single-point calculation on naphthalene (see attached input and output files). The parser reads correctly the single-point calculation, but fails on the geometry optimization. Basis set and all Firefly options are the same in both cases.

[naphtalene.zip](https://github.com/cclib/cclib/files/88720/naphtalene.zip)

Python 2.7.10 (default, Jul 20 2015, 12:26:59)
[GCC 4.3.4 [gcc-4_3-branch revision 152973]] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from cclib.parser import ccopen
>>> ff_in=ccopen("t_0.out")
>>> data=ff_in.parse()
[GAMESS t_0.out WARNING] MO section found but could not be parsed!
[GAMESS t_0.out WARNING] Overwriting previous multipole moments with new values
[GAMESS t_0.out WARNING] This could be from post-HF properties or geometry optimization
>>> ff_in=ccopen("t_0_SP.out")
>>> data=ff_in.parse()
>>>

mkrompiec:
Actually, it is more complicated, dependent on the Firefly binaries.
For most recent binaries (8.1.1), the parser displays the warning message (which is misleading), but actually parses the MOs correctly (there is only one set of MOs in the test file, for the optimised geometry). 
Older binaries (verified for 8.1.0) add the following message in many places of the output file:
 WARNING! THIS VERSION OF FIREFLY IS PROBABLY OUTDATED!
 PLEASE CHECK FIREFLY HOMEPAGE FOR INFORMATION ON UPDATES!
And this message causes the parser to fail as in the following example, whilst displaying the same warning:
>>> ff_in=ccopen("eu3_combo_uks_1_s0.out")
>>> data=ff_in.parse()
[GAMESS eu3_combo_uks_1_s0.out WARNING] MO section found but could not be parsed!
[GAMESS eu3_combo_uks_1_s0.out WARNING] Overwriting previous multipole moments with new values
[GAMESS eu3_combo_uks_1_s0.out WARNING] This could be from post-HF properties or geometry optimization
>>> data.moenergies[0][0]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IndexError: index 0 is out of bounds for axis 0 with size 0

langner:
Thanks for reporting this. Can we use your files as regression tests?

[...]

PR #282, opened 2016-06-19
GAMESS-US/Firefly: fix geometry optimization warning and outdated Firefly crash

berquist:
See https://github.com/cclib/cclib/issues/228.

This currently fixes the "MO section found but could not be parsed!" warning (that also appears for the current test suite geometry optimizations). Waiting on the files that would help fix "WARNING! THIS VERSION OF FIREFLY IS PROBABLY OUTDATED!"

- [x] Ready to go


langner:
Is this ready to go?

berquist:
@mkrompiec hasn't responded with the other files, so I'm fine with it being done.

PR #283, opened 2016-06-21
Added suport for Gaussian BOMD trajectories (1 trajectory per file)

rcrehuet:
This also adds the attribute `time` to the parsed data.
Only energies, times and geometries are parsed, not orbitals.
The main cclib version failed because the integration step sometimes rejects a step and re-calculates, but cclib parsed all of them, creating too many geometries and energies (that did not agree with the number of steps).
Currently it onyl works with files containing a single trajectory. Only tested with DFT runs.


src/cclib/parser/data.py
```diff
@@ -70,6 +70,7 @@ class ccData(object):
         scftargets -- targets for convergence of the SCF (array[2])
         scfvalues -- current values for convergence of the SCF (list of arrays[2])
         temperature -- temperature used for Thermochemistry (float, kelvin)
+        time -- time in molecular dynamic trajectories (array[1], fs)
```
langner: I'm OK with adding this attribute, but perhaps change description to "time in molecular dynamics and other trajectories" since we could use this for any other models of dynamics.

langner:
> The main cclib version failed because the integration step sometimes rejects a step and re-calculates, but cclib parsed all of them, creating too many geometries and energies (that did not agree with the number of steps).

You might want to consider also parsing the rejected steps if they are meaningful.

rcrehuet:
I don't think the rejected steps are useful. It is true that they always convey some information, that may be useful to somebody, but they are physically meaningless: the algorithm rejects those steps and calculates new ones.
In practical terms, it would mean that for a given time frame, we would have two different geometries (but only a valid one). 
My idea for this parser was to return something that is close to a molecular dynamics trajectory that can be analysed afterwards in a similar way as trajectories from Molecular Dynamics software. In other words, a series of geometries associated to a given time.

langner:
Yes, that makes sense.

For your convenience, I merged master into this in rcrehuet/cclib#1 since there were small conflicts. If you can merge than into your branch, I'll then merge in this pull request.

However, could you take care of the remaining issue in cclib/cclib-data#36?

I also opened an issue for general trajectory support in #289

PR #286, opened 2016-07-09
Molpro: natural orbitals

berquist:
Parse natural orbitals from Molpro outputs, and consolidate canonical/natural orbital parsing into a single method.

### DIFF
--- a/src/cclib/parser/molproparser.py
+++ b/src/cclib/parser/molproparser.py
@@ -381,132 +381,10 @@ class Molpro(logfileparser.Logfile):
             else:
                 self.moments[1] == dipole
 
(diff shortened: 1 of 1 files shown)


langner:
Probably put private method before extract()?

berquist:
> Probably put private method before extract()?

At first I didn't think this makes much sense, but it does considering that it's only called by extract().

The Molpro parser doesn't have this, but where should non-class functions go, before or after the parser class?

langner:
> The Molpro parser doesn't have this, but where should non-class (public or private) functions go, before or after the parser class?

There's no recommendation for this in PEP8. Although one could say it is a matter of taste, I think it's most common to define things before they are referenced if possible. In python this is not strictly necessary with classes, since the names needs to be defined only as code is executed.

PR #292, opened 2016-07-28
GAMESS: update unit tests

berquist:
The parsing of *.dat files is a work-in-progress, considering that the parsers aren't designed to read files that vary quite a bit from version to version. These are different enough that they may require their own parser.


berquist:
The *.dat file parsing work hasn't been done, but the updated unit tests should be good to go.

Issue #89, opened 2014-04-07
Switch to atomic units


langner:
We might have had this discussion on the mailing list before, but it would be nice to be able to get all attributes in alternative units. Mostly atomic units for me (hartress, etc), since I find myself constantly converting to those.

It would be useful both in the parsers and ccopen/ccget.

berquist:
Since this is a breaking change, should it go in the v2.0 milestone?

langner:
Setting atomic as the default would be breaking, but adding the ability to switch could be done earlier.

langner:
@berquist Do you think defaulting to atomic is desirable for 2.x?

berquist:
Yes, absolutely.

langner:
As per discussion, we are thinking to just switch for 2.x, but not enabling converting functionality. Store the unit in attribute classes, but don't plan to add convert method for now.

berquist:
Here are all quantities (and perhaps more) that have units attached. I could use help in fleshing this out. What needs to be changed aside from the usual electronic energies?

| attribute          | current units    | future units |
|--------------------|------------------|--------------|
| atomcharges        |                  |              |
| atomcoords         | angstroms        |              |
| atommasses         | daltons          |              |
| atomspins          |                  |              |
| ccenergies         | eV               | hartree      |
| charge             |                  |              |
| dispersionenergies | eV               |              |
| enthalpy           | hartree/particle |              |
| entropy            | particle*kelvin  |              |
| etenergies         | 1/cm             | hartree      |
| etoscs             |                  |              |
| etdips             | ebohr            |              |
| etveldips          | ebohr            |              |
| etmagdips          | ebohr            |              |
| etrotats           | ??               |              |
| freeenergy         | hartree/particle |              |
| geotargets         |                  |              |
| geovalues          |                  |              |
| grads              |                  |              |
| hessian            |                  |              |
| moenergies         | eV               | hartree      |
| moments            | a.u.             |              |
| mpenergies         | eV               | hartree      |
| polarizabilities   | a.u.             | a.u.         |
| pressure           | atm              |              |
| scancoords         | angstroms        |              |
| scanenergies       | eV               | hartree      |
| scanparm           |                  |              |
| scfenergies        | eV               |              |
| scftargets         |                  |              |
| scfvalues          |                  |              |
| temperature        | kelvin           | kelvin       |
| time               | fs               |              |
| transprop          | etoscs           |              |
| vibanharms         | 1/cm             | 1/cm         |
[...]

PR #295, opened 2016-08-01
New fields for Gaussian parser

jaimergp:
I needed additional fields not present in the original parser so I went ahead and extended the .extract() method. I had to also extend the ccData class with the new attributes:
- stoichiometry: str,
- thermalenergies: float,
- zeropointenergies: float
- imaginaryfreqs: int
- alphaelectrons: int
- betaelectrons: int

It also fixes a typo in the free energy parsing line (it was missing an e: `freenergy` instead of `freeenergy`).

I hope you find it useful!


jaimergp:
Quick question: do these energies need unit conversion like those in `scfenergies`?

langner:
We generally prefer not to parse something if it can be calculated. For example, `stoichemistry` can be built from the list of atomic numbers, right?

jaimergp:
Fair enough. Let me know which ones you want to keep!

langner:
Well, let's look at them one at a time. I think `stoichiometry` is easy to calculate - I sent jaimergp/cclib#1 for you to incorporate. Can you check if it works for you? We could also make it a method on the ccData class if that would be more convenient.

langner:
> Quick question: do these energies need unit conversion like those in scfenergies?

Yes, it would be best to be consistent and present all energies in the same units. We would like to be able to switch in the future (cclib/cclib#89).

langner:
Added the charge to the formula in jaimergp/cclib#2. Anything else that needs to be done for the stoichiometry?

langner:
It seems you added carriage returns to data.py in 3a9e4dddbd115d0997da6fd6d1032d592e978329, marking the whole file as changed. Could you please rewrite this commit so that this doesn't happen?

src/cclib/parser/gaussianparser.py
```diff
@@ -1523,12 +1523,24 @@ def extract(self, inputfile, line):
         #Sum of electronic and thermal Energies=              -563.636699
         #Sum of electronic and thermal Enthalpies=            -563.635755
         #Sum of electronic and thermal Free Energies=         -563.689037
-        if "Sum of electronic and thermal Enthalpies" in line:
-            self.set_attribute('enthalpy', float(line.split()[6]))
-        if "Sum of electronic and thermal Free Energies=" in line:
-            self.set_attribute('freenergy', float(line.split()[7]))
         if line[1:12] == "Temperature":
             self.set_attribute('temperature', float(line.split()[1]))
+        if "Sum of electronic and zero-point Energies=" in line:
```
langner: For zero point energy we should probably subtract the electronic energy. @berquist any comments on this?

src/cclib/parser/data.py
```diff
@@ -1,317 +1,328 @@
...
+
+
+    def setattributes(self, *args, **kwargs):
+
+        invalid = super(ccData_optdone_bool, self).setattributes(*args, **kwargs)
+
+        # Reduce optdone to a Boolean, because it will be parsed as a list. If this list has any element,
+        # it means that there was an optimized structure and optdone should be True.
+        if hasattr(self, 'optdone'):
+            self.optdone = len(self.optdone) > 0
```
berquist: Is there a reason why all of these lines have been changed? Did something happen with line endings?

src/cclib/parser/gaussianparser.py
berquist: Yes, it should be subtracted out. Though, why not use the line with `Zero-point correction=`?

src/cclib/parser/data.py
langner: It seems they were added in 3a9e4dd (see my comment above).

berquist: Ok, I see it now. We should have a short page in the official documentation for a few simple guidelines, and one of them would be setting `git config --global core.autocrlf input` for Linux/OS X and `true` for Windows. I think that would avoid the problem of not all text editors obeying existing line endings.

langner:
Pushing the remaining things to do here back to v1.5.1.

src/cclib/parser/gaussianparser.py
```diff
@@ -1523,12 +1523,24 @@ def extract(self, inputfile, line):
...
+            self.set_attribute('zeropointenergies', utils.convertor(float(line.split()[6]), 'hartree', 'eV'))
+        if "Sum of electronic and thermal Energies" in line:
+            self.set_attribute('thermalenergies', utils.convertor(float(line.split()[6]), 'hartree', 'eV'))
+        if "Sum of electronic and thermal Enthalpies" in line:
+            self.set_attribute('enthalpy', utils.convertor(float(line.split()[6]), 'hartree', 'eV'))
+        if "Sum of electronic and thermal Free Energies=" in line:
+            self.set_attribute('freeenergy', utils.convertor(float(line.split()[7]), 'hartree', 'eV'))
+        if "imaginary frequencies (negative Signs)" in line:
+            self.set_attribute('imaginaryfreqs', int(line.split()[1]))
+        if "alpha electrons" in line:
```
langner: This is something we can currently get from `homos`, correct?

jaimergp:
Hi guys, I'm back and ready to fix the line-endings and commit noise I introduced in 3a9e4dd. I'm also ready to implement these features in a cleaner way if they are already available in `homos`, but I'd need some hints there!

jaimergp:
Hi again. After fiddling with the settings of autocrlf and related options, I am not able to reproduce that big diff. I've tried with rebase but, my git-foo is not that powerful... Any help is appreciated!

langner:
I'm going to look into this so we can try to merge it in for v1.5.2, it look like we were only waiting on cosmetic changes.

jaimergp:
Hi guys! Any ETA on this PR merge?

ATenderholt:
It looks like we were primarily waiting on fixes to the line endings for `data.py`. You should also take care of the current conflicts.

jaimergp:
Well, I have invested more time than I will admit with `git rebase` trying to fix the line endings mess, but I cannot get it. There's a conflict between the lines changed on commit c19201a7431ba6b16734fff2460dc6c7afe14642 (you can see the trailing ^M there), and then when the PR is merged, the mess begins. To me, it seems that commit 3a9e4dddbd115d0997da6fd6d1032d592e978329 tried to unify the line-endings, but the damage was done by that time. I am seriously considering starting all over again with proper line-endings in a new PR...

jaimergp:
I am tempted to start all over with this PR and forget about the commit mess... @langner authored the stoichiometry part, but I can (re)submit the changes having to do with:

- thermalenergies
- zeropointenergies
- alphaelectrons
- betaelectrons

Is that OK?

langner:
Well... it seems something's wrong with the current set of changes (only `alphaelectrons` shows up in the parser code now).

If you want to split the PR up into smaller bits, a reasonable way to do it is to have one attribute per PR.

jaimergp:
Yes, @langner, I will get back to this as soon as I can. For the time being, we can close this PR and I will create new separate PRs as needed. By the way, which refactors should I look into?

jaimergp:
Ok, I am ready to get back to this. Before opening a new PR and forgetting about this, I'd like to sum up the pending tasks:

- [x] Be careful with line endings, trailing whitespace and so on.
- [ ] Thermochemistry. Do we really want more ccData attributes for `Sum of electronic and zero-point Energies` and `Sum of electronic and thermal Energies`? If so, which names should we use? `zeropointenergies` and `thermalenergies`? (Plural or singular?)
- [ ] Alpha and beta electrons. I think someone mentioned that it can be computed out of another field already? `homos`?
- [ ] Stoichiometry. @langner implemented that and I wouldn't like to take credit for that. The code is described [here](https://github.com/jaimergp/cclib/pull/2/files).

Depending on the answers, I will start independent PRs for each point, if that's OK.

berquist:
> Thermochemistry. Do we really want more ccData attributes for `Sum of electronic and zero-point Energies` and `Sum of electronic and thermal Energies`? If so, which names should we use? `zeropointenergies` and `thermalenergies`? (Plural or singular?)

I have notes somewhere on this, but maybe you can tell me: can we calculate those numbers, or are they unique fields? If they're unique, I think we should let the user/external program calculate them, and mention this in the Gaussian section of the documentation. We should avoid redundant/non-unique attributes if possible.

For singular/plural, right now all our thermochemistry attributes are singular, because only single values get parsed, and we assume there's only one thermochemistry block per calculation. This will change in the future, because many programs can do multiple isotope runs for a single frequency calculation.

> Alpha and beta electrons. I think someone mentioned that it can be computed out of another field already? `homos`?

Yes. Generally,
``` python
if len(homos) == 1:
    nalpha = nbeta = homos[0] + 1
else:
    nalpha = homos[0] + 1
    nbeta = homos[1]
```

> Stoichiometry. @langner implemented that and I wouldn't like to take credit for that. The code is described [here](https://github.com/jaimergp/cclib/pull/2/files).

You can make a new PR from exactly those commits and all the metadata, like authorship, will be correctly retained.

---

Yes, you should make new PRs. You can use `git cherry-pick <id1> <id2> ...` in a new branch to get the commits you want out of this PR.

langner:
SGTM. Rebasing in this PR is OK with me as well, but new ones would be cleaner. Please just reference this one from the new ones to keep the history of communication.

berquist:
I would definitely prefer new PRs.

jaimergp:
Ok, got it. In the following days I will be posting the new PRs.

langner:
@jaimergp Do you think you will have time in the next week to finish this up before the release?

jaimergp:
Hi @langner! I have doubts on what I am supposed to do :/ 

After reading @berquist's answer to point 2, I think we should start a conversation about what to do with the thermochemistry. Maybe creating an issue for this is enough for now?

So, the only thing left is to calculate the number of alpha and beta electrons with, I guess, the same approach we did for `stoichiometry`: have a method somewhere and then call it with a property from `ccData`? If this is the only thing I have to do, yes, it can be done during this week.

berquist:
> I think we should start a conversation about what to do with the thermochemistry. Maybe creating an issue for this is enough for now?

Sounds exactly right. I'm not comfortable adding attributes for those fields until I can remember how they map to fields in other programs, but I haven't gotten around to it.

> have a method somewhere and then call it with a property from ccData? If this is the only thing I have to do, yes, it can be done during this week.

Maybe inside `method.electrons.Electrons`?

langner:
OK... pushing this back for after the release.

jaimergp:
Since, the relevant PRs have been submitted and I have just create the issue to discuss thermochemistry, I'd say we can dismiss this PR and close it.

PR #297, opened 2016-08-02
Q-Chem: parse gradients/Hessians from correlated calculations

berquist:


src/cclib/parser/qchemparser.py
```diff
@@ -1331,6 +1309,30 @@ def generate_formula_histogram(self):
         return histogram
 
 
+def parse_matrix(inputfile, nparray, ncolsblock):
```
langner: Since this is specific to QChem, why not make it a method of the parser?

src/cclib/parser/qchemparser.py
```diff
@@ -94,6 +94,17 @@ def before_parsing(self):
...
 
+        self.gradient_headers = (
+            'Full Analytical Gradient',
+            'Gradient of SCF Energy',
+            'Gradient of MP2 Energy',
+        )
+
+        self.hessian_headers = (
+            'Hessian of the SCF Energy',
+            'Final Hessian.',
```
langner: Can we add unit tests or regression to cover these cases?

berquist: Yes, I will add some small regression tests; I'd hold off on unit tests simply because we don't have gradients and Hessians implemented for most parsers right now.

src/cclib/parser/qchemparser.py
```diff
@@ -154,6 +154,140 @@ def after_parsing(self):
...
+            if has_spins:
+                spin = self.float(elements[3])
+                spins.append(spin)
+            line = next(inputfile)
+
+        self.atomcharges[chargetype] = numpy.array(charges)
+        if has_spins:
+            self.atomspins[chargetype] = numpy.array(spins)
+
+    @staticmethod
```
berquist: @langner Is this an appropriate use of a static method in Python? I'm slightly more familiar with it in C++.

langner: Yes. But modifying a passed object is not very Pythonic. Is there any reason not to create it in the method and return it?

berquist: I've spent too much time spent in C++ land!

berquist:
Regression tests are located at https://github.com/cclib/cclib-data/pull/42.

src/cclib/parser/qchemparser.py
langner: I know the feeling :)

PR #298, opened 2016-08-03
Add a method for detecting closed shell systems

langner:


src/cclib/method/orbitals.py
```diff
@@ -0,0 +1,59 @@
...
+            assert len(self.data.moenergies) == 1
+            return True
+
+        # If there are beta orbitals, we can assume the system is
+        # closed shell if the orbital energies are identical within
+        # numerical accuracy.
+        diff = numpy.abs(self.data.moenergies[1] - self.data.moenergies[0])
+        precision = 10e-6
+        return any(diff < precision)
+
```
TDI-tenderholt: Maybe use [numpy.allclose](http://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html)?

test/method/testorbitals.py
```diff
@@ -0,0 +1,53 @@
...
+class UnrestrictedCalculationTest(unittest.TestCase):
+    """Check RHF calculation."""
+    def setUp(self):
+        self.data, self.logfile = getdatafile(Gaussian, "basicGaussian09", "dvb_un_sp.log")
+
+    def test_closed_shell(self):
+        self.assertFalse(Orbitals(self.data).closed_shell())
+
+
+# TODO: add a case (regression) with an unrestricted calculation for a closed shell system.
```
TDI-tenderholt: The Mo4OSibdt2 file in the Gaussian/Gaussian03 regressions should work. Probably also worth tracking down a closed-shell broken symmetry example. I should have a Mo tris-dithiolene example somewhere, although there might be a smaller example.

src/cclib/method/orbitals.py
langner: Great catch!

test/method/testorbitals.py
langner: Added a comment. We would need to add this as a regression test (will do after this is in).

berquist:
How do you plan on handling restricted open-shell systems (where `len(data.homos) == 2` but `len(data.mocoeffs) == 1`)? Does the `numpy.allclose` test catch that?

langner:
> How do you plan on handling restricted open-shell systems (where len(data.homos) == 2 but len(data.mocoeffs) == 1)? Does the numpy.allclose test catch that?

The code doesn't use `data.homos`. But it's a good test case to add.

langner:
> > How do you plan on handling restricted open-shell systems (where len(data.homos) == 2 but len(data.mocoeffs) == 1)? Does the numpy.allclose test catch that?

> The code doesn't use data.homos. But it's a good test case to add.

I now understand this. Would add an additional check for len(data.homos) to catch the restricted open-shell case?

berquist:
> I now understand this. Do you think I should add an additional check for len(data.homos) to catch the restricted open-shell case?

Yes, I think that's the first step.

langner:
I added the logic for ROHF (hopefully it's correct) and did some rebasing. Let me know if there's anything left to do here.

berquist:
Not sure how this escaped my notice...

PR #300, opened 2016-08-05
Add an electron counting method and expose it as a property on ccData

langner:
* [x] Ready to merge

### DIFF
--- a/src/cclib/parser/data.py
+++ b/src/cclib/parser/data.py
@@ -301,7 +301,7 @@ class ccData(object):
         return self.write(filename=filename, outputtype='xyz')
 
     @property
-    def electrons(self):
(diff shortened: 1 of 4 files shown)


src/cclib/parser/data.py
```diff
@@ -298,6 +300,10 @@ def writexyz(self, filename=None):
         """Write parsed attributes to an XML file."""
         return self.write(filename=filename, outputtype='xyz')
 
+    @property
+    def electrons(self):
```
berquist: Can we call it `nelectrons` to follow the same convention as `nmo`, `nbasis`, etc., ?

langner: Done

PR #302, opened 2016-08-07
Psi4: update unit tests and handle UHF/ROHF

berquist:
This brings the parser up-to-date with Psi4 1.0.

- [x] Ready for review


langner:
Can you put the beta logfile into the regression suite?

langner:
So is this ready to go as well? Do we want to include it in v1.5?

Also - did we ever have more test logfiles for Psi3?

berquist:
This is done.

There weren't more files for Psi3. I do have a copy but I'm not sure it's worth putting in much effort to support it, considering it's almost 10 years old.

langner:
> This is done.

Great. I'll take one last look today and merge in.

> There weren't more files for Psi3. I do have a copy but I'm not sure it's worth putting in much effort to support it, considering it's almost 10 years old.

I see. There is a version in Debian/Ubuntu, so there might be some use in getting the files in (#319).

PR #304, opened 2016-08-14
Updates for ROHF

berquist:


langner:
It look like this breaks one of the NWChem regression tests.

Issue #293, opened 2016-07-28
Investigate behavior with streams


langner:
@Schamnad has reported FileWrapper might be gobbling up the beginnings of files.

langner:
I don't think I've personally tried using stdin with our FileWrapper, but it is something we should cover. This is a good chance to add some tests to https://github.com/cclib/cclib/blob/master/test/parser/testlogfileparser.py

Schamnad:
exact issue I am having is:
[Here](https://github.com/Schamnad/avogadrolibs/blob/2c3811aaf07d2d70e0ce697893dd9277b60f2309/avogadro/qtplugins/scriptfileformats/formatScripts/cclib/cclibScript.py#L32) I pass sys.stdin to ccopen. When the log.parse() function is called, in different parsers (for example the gaussianparser/gamessparser) the first line passed to the extract() [method](https://github.com/cclib/cclib/blob/master/src/cclib/parser/logfileparser.py#L301) varies, but it seems to miss the data pertaining to the input file.

For example, for the gamess output file [C_bigbasis.out](https://github.com/cclib/cclib/blob/master/data/GAMESS/basicFirefly8.0/C_bigbasis.out), the first line passed to the extract method is line [#188](https://github.com/cclib/cclib/blob/master/data/GAMESS/basicFirefly8.0/C_bigbasis.out#L188) onwards whereas information required about the basis sets is on line [#146](https://github.com/cclib/cclib/blob/master/data/GAMESS/basicFirefly8.0/C_bigbasis.out#L146). This later on causes an attribute error as it tries to access the self.nbasis attribute .
The same error is reproducible with the gaussian output file albeit parsing starting at a different start point.

I have been investigating it in two directions:

1. The Filewrapper might be gobbling up the beginnings of files
2.  Avogadro passes the sys.stdin after skipping the beginning of files

Investigation into Point 1. has resulted in the following:
In my computer Python2 is being used, so the Filewrapper initialization always goes into the '[except](https://github.com/cclib/cclib/blob/master/src/cclib/parser/logfileparser.py#L91)' block but it is unable to set the [size](https://github.com/cclib/cclib/blob/master/src/cclib/parser/logfileparser.py#L93) attribute

Investigation into Point 2:
I printed the first line (basically print(sys.stdin.readline)) before passing it to ccopen, and it printed out the first line of the output file.

This can be taken as confirmation that the Filewrapper is the one that is gobbling up the initial parts.

langner:
[...]

PR #306, opened 2016-08-16
Make all streams work by reading them in at the start

langner:
This will be a fix specifically for #293 

- [x] Ready for review
- [x] Ready for merging

### DIFF
--- a/src/cclib/parser/ccio.py
+++ b/src/cclib/parser/ccio.py
@@ -3,7 +3,7 @@
 # This file is part of cclib (http://cclib.github.io), a library for parsing
 # and interpreting the results of computational chemistry packages.
 #
-# Copyright (C) 2009-2014, the cclib development team
+# Copyright (C) 2009-2016, the cclib development team
 #
 # The library is free software, distributed under the terms of
 # the GNU Lesser General Public version 2.1 or later. You should have
@@ -15,6 +15,7 @@
 
 from __future__ import print_function
 
+import io
 import os
 import sys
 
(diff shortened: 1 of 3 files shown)


langner:
This should fix all our stream woes at the moment, at the cost of reading them all at init time. This might be a problem for really large stream... but let's deal with that problem when it comes up. The code still needs to be cleaned up. In particular, we don't need FileWrapper anymore, and a bunch of other things... but this PR is just about fixing.

src/cclib/parser/logfileparser.py
```diff
@@ -86,20 +86,23 @@ def __init__(self, source, pos=0):
             self.src.seek(0, 2)
             self.size = self.src.tell()
             self.src.seek(pos, 0)
-            self.pos = pos
 
-        except:
+        except (AttributeError, IOError):
```
TDI-tenderholt: Comment above the start of the try block sounds like a test case that throws an ```UnsupportedOperation``` should be found.

src/cclib/parser/ccio.py
```diff
@@ -136,20 +137,29 @@ def ccopen(source, *args, **kargs):
...
+        except (AttributeError, IOError):
+            contents = inputfile.read()
+            inputfile = io.StringIO(contents)
+            inputfile.seek(0, 0)
+
     # Proceed to return an instance of the logfile parser only if the filetype
     # could be guessed. Need to make sure the input file is closed before creating
     # an instance, because parsers will handle opening/closing on their own.
     filetype = guess_filetype(inputfile)
+    inputfile.seek(0, 0)
```
TDI-tenderholt: If you call ```inputfile.seek(0, 0)``` here, why do you need a special case for ```isstream``` other than for the try-except block? Seems like you could get rid of the ```if istream``` block and remove the extra seek here.

src/cclib/parser/logfileparser.py
```diff
@@ -86,20 +86,23 @@ def __init__(self, source, pos=0):
...
-            # Assume the position is what was passed to the constructor.
-            self.pos = pos
+        # Assume the position is what was passed to the constructor.
+        self.pos = pos
 
     def next(self):
         line = next(self.src)
         self.pos += len(line)
+        if self.size < self.pos:
+            self.size = self.pos
```
TDI-tenderholt: I think this will show 100% complete in a progress dialog. Has this been tested?

src/cclib/parser/logfileparser.py
langner: Added.

src/cclib/parser/logfileparser.py
langner: Well, this shouldn't happen now since the stream is read at the start - removed. I think the next step is getting rid of FileWrapper and creating something in ccio that deals with files and streams transparently, because this is starting to get confusing!

langner:
With this change, reading in at the start will happen almost always when using streams. So it'll happen inside Avogadro where Sanjeed has used stdin to pass the logfile contents. This won't be problematic in that case, since the input will be local and fast. This might be an issue for very large HTTP streams that are also slow. Parsing and reading simultaneously might have an advantage there, but I consider that a rare edge case.

langner:
Ugh, had to add a fix for Python2 strings. Should be good to go now.

langner:
Created an issue for further work in #309

PR #307, opened 2016-08-16
CJSON Reader - Removal of cjsonHolder DS

Schamnad:
Highlights of the changes:

1. The reader has been hooked to the ccget script such that a user can parse a cclib attribute in the following way:
`ccget --json <attribute/attribute list> *.cjson`
To use the --list flag with any cjson file, the --json flag must also be provided. For. eg:
`ccget --json --list <attribute/attribute list> *.cjson` 
without the --json flag, the --list flag will throw an error

2. The writers/ folder has been renamed to io/ as it contains both readers and writers
3. The CJSON reader class has not been implemented as a ccData type because it reads a cjson file and is not a log file
4. The changes have been locally tested along with the code change mentioned in https://github.com/cclib/cclib-data/pull/39 . All regression test cases pass. Once this PR is merged, the above PR needs to be merged.
5. This PR contains a few commits referring to a cjsonHolder DS which was deemed superfluous and hence removed, if the repository owners wish not to pollute the commit history with a DS which doesn't exist in the end, then I can squash all commits related to cjsonHolder into one and create a new PR


Schamnad:
Duplicate PR: https://github.com/cclib/cclib/pull/308

Issue #266, opened 2016-04-21
Moments not parsed for Benzeneselenol


langner:
This is a follow-up to #261 concerning the fact that moments don't seem to get parsed for that regression.

berquist:
Just putting the cclib-data link here: https://github.com/cclib/cclib-data/pull/30

langner:
Note: add regression test for moments for that output file once the fix is in.

berquist:
Maybe this was already obvious, but I didn't realize `moments` isn't implemented for Gaussian at all.

langner:
Wasn't to me. Pushing back to 1.5.

berquist:
https://github.com/cclib/cclib-data/pull/48 is a regression to check for this, now that `moments` are implemented.

langner:
This looks all done.

PR #310, opened 2016-08-17
Parse moments attribute for Gaussian, including dipole and quadrupole

ghutchis:
Should solve #266

### DIFF
--- a/src/cclib/parser/gaussianparser.py
+++ b/src/cclib/parser/gaussianparser.py
@@ -268,6 +268,64 @@ class Gaussian(logfileparser.Logfile):
                 else:
                     assert self.moments[2] == quadrupole
 
+        if line[1:41] == "Octapole moment (field-independent basis":
+        # e.g.
+        # Octapole moment (field-independent basis, Debye-Ang**2):
+        #  XXX=              0.0000  YYY=              0.0000  ZZZ=             -0.1457  XYY=              0.0000
+        #  XXY=              0.0000  XXZ=              0.0136  XZZ=              0.0000  YZZ=              0.0000
+        #  YYZ=             -0.5848  XYZ=              0.0000
+            octapole = {}
+            tokens = inputfile.next().split()
+            for i in (0, 2, 4, 6):
+                octapole[tokens[i]] = float(tokens[i+1])
+            tokens = inputfile.next().split()
+            for i in (0, 2, 4, 6):
+                octapole[tokens[i]] = float(tokens[i+1])
+            tokens = inputfile.next().split()
+            for i in (0, 2): # last line only 4 tokens
+                octapole[tokens[i]] = float(tokens[i+1])
+
+            lex = sorted(octapole.keys())
+            octapole = [octapole[key] for key in lex]
+
+            if not hasattr(self, 'moments') or len(self.moments) < 3:
+                self.logger.warning("Found octapole moments but no previous dipole or quadrupole")
+                self.moments = [self.reference, None, None, octapole]
+            else:
+                if len(self.moments) == 3:
+                    self.moments.append(octapole)
+                else:
+                    assert self.moments[3] == octapole
+
+        if line[1:20] == "Hexadecapole moment":
+        # e.g.
+        # Hexadecapole moment (field-independent basis, Debye-Ang**3):
+        # XXXX=             -3.2614 YYYY=             -6.8264 ZZZZ=             -4.9965 XXXY=              0.0000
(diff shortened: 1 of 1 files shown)


ghutchis:
I'll happily add the other moments - just want feedback about formatting, style, etc. first.

ghutchis:
OK, I'll fix the test fail.

src/cclib/parser/gaussianparser.py
```diff
@@ -223,6 +223,109 @@ def extract(self, inputfile, line):
             natom = int(line.split()[1])
             self.set_attribute('natom', natom)
 
+        # Dipole moment
+        # e.g.
+        #  Dipole moment (field-independent basis, Debye):
+        #    X=              0.0000    Y=              0.0000    Z=              0.0930
+        if line[1:39] == "Dipole moment (field-independent basis":
```
berquist: I think that if the ` (field-independent basis` part is removed from all the match checks, all the older regression tests that are failing will pass (rather than adding it to the hexadecapole line; then older logfiles won't have their `moments` parsed at all).

src/cclib/parser/gaussianparser.py
```diff
@@ -223,6 +223,109 @@ def extract(self, inputfile, line):
...
+
+            self.updateprogress(inputfile, "Dipole and Higher Moments", self.fupdate)
+
+            self.reference = [0.0, 0.0, 0.0]
+            self.moments = [self.reference]
+
+            tokens = inputfile.next().split()
```
berquist: Specifying the exact field widths for all the moments will fix the other test failure.

ghutchis: I can do that - I just don't know the Python syntax. In C/C++, I'd use sscanf() with a format string.

src/cclib/parser/gaussianparser.py
ghutchis: Sure.

src/cclib/parser/gaussianparser.py
berquist: Oh, I'm referring to explicit slicing (`dipole = float(line[a:b]), float(line[b:c]), float(line[c:d])`) rather than splitting.

ghutchis: D'oh! I have no idea why that didn't occur to me. I'll do the character counting.

ghutchis:
This should now give moments for all Gaussian files that have them. Some g03 don't seem to always have more than dipole moments, but that's fine.

berquist:
Everything looks good to me. One last thing: can you try removing https://github.com/cclib/cclib/blob/master/test/data/testSP.py#L195 and running the tests?

langner:
> Everything looks good to me. One last thing: can you try removing https://github.com/cclib/cclib/blob/master/test/data/testSP.py#L195 and running the tests?

+1

ghutchis:
Two files error - there are no moments in those files. (I didn't know it was possible to run a Gaussian calculation without at least the dipole moment, so you learn something new.)

- basicGaussian03/dvb_sp.out
- basicGaussian09/dvb_sp.out

Otherwise, it looks clean. If you suggest the appropriate @skipForLogfile bits, I'll add that and push.

```
basicGaussian03/['dvb_sp.out']: Generic restricted single point unites
testmoments (data.testSP.GenericSPTest)Traceback (most recent call last):
  File "/Users/ghutchis/Devel/cclib/test/data/skip.py", line 21, in test wrapper
  File "/Users/ghutchis/Devel/cclib/test/data/testSP.py", line 202, in testaments
    reference = self.data.moments[0]
  File "test_data.py", line 118, in ccdata_getattribute_with_coverage
    return object.__getattribute__(self, attir
AttributeError: 'ccData_optdone_bool' object has no attribute 'moments'

basicGaussian09/['dvb_sp.out']: Generic restricted single point unites
testmoments (data.testSP.GenericSPTest)Traceback (most recent call last):
  File "/Users/ghutchis/Devel/cclib/test/data/skip.py", line 21, in test wrapper
  File "/Users/ghutchis/Devel/cclib/test/data/testSP.py", line 202, in testaments
    reference = self.data.moments[0]
  File "test_data.py", line 118, in ccdata_getattribute_with_coverage
    return object.__getattribute__(self, attir
AttributeError: 'ccData_optdone_bool' object has no attribute 'moments'

```

berquist:
Leaving that `@skipForParser` line the way it was is fine. My first guess was that the calculation input line had `#n` or `#`, but it's `#p`, so who knows. I suspect these will be rerun in the coming months, so it can be fixed then.

PR #312, opened 2016-08-18
Another fix for string not being unicode in Python2

langner:


langner:
This is another quick fix in the spirit of #306. Will merge it in right after tests pass.

PR #318, opened 2016-08-20
Replace NaN's with null to form valid json files

Schamnad:
Previously in files containing orbital coefficients, a few values were not discernible and represented as NaN. While dumping the json from such output files the NaN's, being of the float type, would keep the NaN's as is. This resulted in the generated json being invalid, as NaN's are not a valid json type.

This commit converts the NaN's in the numpy ndarray into None, which the default json encoder converts into the json 'null' type resulting in valid json output files.

### DIFF
--- a/src/cclib/io/cjsonwriter.py
+++ b/src/cclib/io/cjsonwriter.py
@@ -324,13 +324,16 @@ class CJSON(filewriter.Writer):
 
 
 class NumpyAwareJSONEncoder(json.JSONEncoder):
-    def default(self, obj):
+    """A encoder for numpy.ndarray's obtained from the cclib attributes.
+       For all other types the json default encoder is called.
+    """
+    def numpy_ndarray_encoder(self, obj):
         if isinstance(obj, np.ndarray):
             if obj.ndim == 1:
                 nan_list = obj.tolist()
                 return [None if np.isnan(x) else x for x in nan_list]
             else:
-                return [self.default(obj[i]) for i in range(obj.shape[0])]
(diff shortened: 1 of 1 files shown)


src/cclib/io/cjsonwriter.py
```diff
@@ -327,7 +327,8 @@ class NumpyAwareJSONEncoder(json.JSONEncoder):
     def default(self, obj):
```
langner: Can you use a more informative name for this method?

src/cclib/io/cjsonwriter.py
```diff
@@ -327,7 +327,8 @@ class NumpyAwareJSONEncoder(json.JSONEncoder):
...
         if isinstance(obj, np.ndarray):
             if obj.ndim == 1:
-                return obj.tolist()
+                nan_list = obj.tolist()
```
langner: No need to convert to a list, you can do `return obj[~np.isnan(obj)]`.

For reference you can read about [Boolean area indexing](http://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#boolean-array-indexing)

Schamnad: But wouldn't that be equivalent to removing all the NaN values and in turn removing all the 'null' types from the attribute in the Json?

Presently we are passing the attributes as is, and the null values indicates the incomplete calculation

langner: Yeah, you're right. This is fine then.

PR #325, opened 2016-08-23
Update license in scripts, fix README.md in manifest.py

langner:


src/scripts/ccget
```diff
@@ -1,14 +1,9 @@
-#!/usr/bin/env python3
```
berquist: Shouldn't these not be removed, but replaced with `#!/usr/bin/env python`? If there's no shebang, I don't think the shell will know how to execute the script.

langner: That's correct, thanks. I think we had python3 explicitly in order to force Py3 for systems that had both. I don't think it matters much, though I can re-add it if you think it's worthwhile.

berquist: As far as I know, this is the best approach. Python 3 might not be installed, and 2/3 doesn't matter so much here since the scripts aren't meant to be user-modified anyways. It's probably a coincidence that there's never been a problem with this before.

Issue #327, opened 2016-08-23
ORCA output file 'if colums[1] == "Energy":  IndexError: list index out of range' parsing error


skattejag:
Hello,

Upon running the following script in order to parse a partially completed nor a completed ORCA 3.0.3 output file [Partial: d.out, Complete: Orca_test.out - renamed to .txt (See Below) ] I received the following error :
```
  File "C:\Users\Cameron\Anaconda3\lib\site-packages\cclib\parser\orcaparser.py", line 88, in extract
    if colums[1] == "Energy":

IndexError: list index out of range
```
I did not receive this error when I attempted to load the test file: dvb_sp_un.out . Could someone please assist me with this. I am using Spyder 2.3.9 and Python 3.5.2 (Windows, 64-bit)

[d.txt](https://github.com/cclib/cclib/files/433324/d.txt)
[Orca_test.txt](https://github.com/cclib/cclib/files/433326/Orca_test.txt)

**test3.py script:**
```
import cclib
print(cclib.__version__)

from cclib.parser import ORCA
myfile = ORCA("d.out")
data = myfile.parse()

help(data)
print("The number of atoms is %d." % data.natom)"
```

**More detailed output:**

```
runfile('C:/CHPC tests/ga012/Temp/cclib_test/test3.py', wdir='C:/CHPC tests/ga012/Temp/cclib_test')
1.5
Traceback (most recent call last):

  File "<ipython-input-6-55b8a1f3bbac>", line 1, in <module>
    runfile('C:/CHPC tests/ga012/Temp/cclib_test/test3.py', wdir='C:/CHPC tests/ga012/Temp/cclib_test')

  File "C:\Users\Cameron\Anaconda3\lib\site-packages\spyderlib\widgets\externalshell\sitecustomize.py", line 714, in runfile
    execfile(filename, namespace)

  File "C:\Users\Cameron\Anaconda3\lib\site-packages\spyderlib\widgets\externalshell\sitecustomize.py", line 89, in execfile
    exec(compile(f.read(), filename, 'exec'), namespace)

  File "C:/CHPC tests/ga012/Temp/cclib_test/test3.py", line 19, in <module>
    data = myfile.parse()

  File "C:\Users\Cameron\Anaconda3\lib\site-packages\cclib\parser\logfileparser.py", line 301, in parse
    self.extract(inputfile, line)

  File "C:\Users\Cameron\Anaconda3\lib\site-packages\cclib\parser\orcaparser.py", line 88, in extract
[...]

PR #328, opened 2016-08-24
ORCA: simple problem for logic entering energy parsing

berquist:
Reported in https://github.com/cclib/cclib/issues/327.


langner:
I guess we'll get the example into regressions.

PR #336, opened 2016-09-14
Add an attribute to indicate Spherical basis sets

Schamnad:


berquist:
@langner, do we want to make this a method that sets a property on `ccData`?

langner:
Yes, why not.

In the long run we probably want to make the distinction between attributes and properties clearer, somehow. I have no clear idea how at this point, though.

langner:
I think we can go with a property/method for detecting this, which is easier to maintain than parsing, so I'm closing this issue. Let me know if I'm wrong.

Issue #301, opened 2016-08-06
ROHF conventions


berquist:
Reading https://github.com/cclib/cclib/pull/298, I realized there aren't any conventions that I know of for what cclib does in the case of ROHF calculations. Is this true?

These are results for DVB cation, ROHF rather than UHF:

```
Attempting to read dalton.out
homos:
[34]
Attempting to read gamess.out
homos:
[34 33]
Attempting to read gaussian.out
homos:
[34]
Attempting to read molpro.out
homos:
[34 33]
Attempting to read nwchem.out
homos:
[34]
Attempting to read orca.out
homos:
[34]
Attempting to read qchem.out
homos:
[34 33]
```

No Psi4 result because it doesn't handle UHF/ROHF right now, or the latest version of the program; I'm making major changes to the parser and test files.

I think we should have unit tests for restricted open-shell calculations, which should be defined as (to start with):

```
assert len(data.homos) == 2
assert len(data.mocoeffs) == 1
assert len(data.moenergies) == 1
```

langner:
Very good idea. Either unit tests or regressions will work.

bwang2453:
A related question: how do we put ROHF into metadata? Right now, we only have one boolean variable called `unrestricted`.

berquist:
Since https://github.com/cclib/cclib/pull/298 will bring us the property `closed_shell`, my thought is that there could be another property, `spin_restricted`.

For a closed-shell system, `spin_restricted = True` would be RHF and `spin_restricted = False` would be RHF treated as UHF.

For an open-shell system, `spin_restricted = True` would be ROHF and `spin_restricted = False` would be UHF.

The special case of an open-shell singlet would be `spin_restricted = False` and `closed_shell = True`, so just like a doublet, triplet, etc. but the HOMO indices for both alpha and beta would be the same.

berquist:
Related: [for systems with no beta electrons](https://github.com/cclib/cclib-data/blob/master/NWChem/NWChem6.0/hydrogen_atom_UHF_cc-pVDZ.out), for both UHF and ROHF, `homos` should be `[0 -1]`, not `[0 0]` which is an open-shell singlet. It would be up to people running no-beta-electron calculations to recognize that this index is invalid :)

langner:
@berquist Could you explain that (`homos`being `[0 -1]`) in more detail? I don't get it.

berquist:
[...]

PR #338, opened 2016-09-17
Update ROHF conventions

berquist:
See the discussion in https://github.com/cclib/cclib/issues/301. Update so that the hydrogen atom gives `data.homos == [0 None]`.
- [ ] Ready for review

### DIFF
--- a/src/cclib/io/cjsonreader.py
+++ b/src/cclib/io/cjsonreader.py
@@ -26,7 +26,8 @@ class CJSON:
     def read_cjson(self):
         inputfile = self.filename
 
-        json_data = json.loads(open(inputfile).read())
+        with open(inputfile) as fh:
+            json_data = json.loads(fh.read())
 
         # Actual update of attribute dictionary happens here
         self.construct(json_data)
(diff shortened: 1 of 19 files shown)


langner:
The tests generally look good now... just 2 failing regressions I think.

berquist:
My work computer was turned off due to a power outage, it has a bunch of uncommitted work to turn `homos` back into an array, and I'm on vacation until the 3rd, so this will probably have to wait until then.

langner:
That's fine. Have a good holiday break!

langner:
As discussed in #301, can we change the type back to an array for compatibility and adjust the length accordingly? We can explain this in the docs for `homos` (http://cclib.github.io/data_notes.html#homos), where we never mention how long the array is, so this solution feels appropriate.

langner:
Pushing back to v.1.5.3

berquist:
This has been open for several years and weighs on my mind every time I get the chance to work on cclib stuff. Maybe there's some good stuff in here, but it's still unclear in https://github.com/cclib/cclib/issues/301 what we want to do. I'll keep the branch just in case.

langner:
Thanks, Eric. I think it would be good to be more aggressive with closing stale PRs in the future. Nothing is lost if the branches stay around, and it'd be great to have a cleaner state.

PR #341, opened 2016-12-11
Add testall script, and minor clenaups

langner:
Besides adding the testall script, which runs test_data in silent mode, this includes some minor cleanups:
1. Use argv only at bottom
2. Make arguments in getdatafile simpler

### DIFF
--- a/test/testall.py
+++ b/test/testall.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2016, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
+
+"""Run all unit tests for cclib."""
+
+import unittest
+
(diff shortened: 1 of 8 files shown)


TDI-tenderholt:
Looks good to me.

test/testall.py
```diff
@@ -0,0 +1,24 @@
...
+import test_data
+
+from test_bridge import *
+from test_io import *
+from test_method import *
+from test_parser import *
+
+
+if __name__ == "__main__":
+	print "Running unit tests for data..."
```
berquist: Oops, Python 2/3 problem.

ATenderholt: Any reason the testall script wasn't added to the travis file? Presumably it would have caught this problem.

langner: Yeah, we don't run all the tests in Travis CI - we skip bridge, I think, IIRC because I couldn't get openbabel to install cleanly.

Issue #322, opened 2016-08-22
Add a testall script


langner:
Our INSTALL file references such a file, so we should probably have one.

langner:
Added in #341

PR #343, opened 2016-12-17
Fix testall for Python3

langner:


langner:
Follow up to #341 for #322

PR #349, opened 2017-03-04
fix parsing qchem coefficients

srtlg:
Hello,

I noticed qchem output files with a very large AO coefficient somewhere high in the virtual MO orbital space. The current version of cclib raises an exception which makes it not useful.  

This branch uses fixed column parsing to read in lines like the following
958  C 42  s      -20.26525 -72.53084  -4.10297  -0.65820 -12.38880-100.67193

Please consider it for merging.

### DIFF
--- a/src/cclib/parser/qchemparser.py
+++ b/src/cclib/parser/qchemparser.py
@@ -18,6 +18,21 @@ from . import logfileparser
 from . import utils
 
 
+class Splitter:
+    def __init__(self, widths):
+        self.start_indices = [0] + list(itertools.accumulate(widths))[:-1]
+        self.end_indices = list(itertools.accumulate(widths))
+
+    def split(self, line):
+        elements = [line[start:end].strip() for (start, end) in zip(self.start_indices, self.end_indices)]
+        for i in range(1, len(elements)):
+            if elements[-1] == '':
+                elements.pop()
+            else:
+                break
+        return elements
+
+
 class QChem(logfileparser.Logfile):
     """A Q-Chem 4 log file."""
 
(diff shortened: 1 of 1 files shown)


berquist:
Thanks for the PR! I will figure out exactly why the Python 2 test is failing. In the meantime, if it isn't confidential, could you post somewhere the specific output file that's causing this, so we can add it to our regression suite?

srtlg:
Python 2.7 failing is probably itertools.accumulate.  I don't know if I have time the upcoming week, maybe I can check how you parse GAMESS output.  This is also fixed column width.  The culprit output file is quite big and doesn't contain all interesting bits (e.g. two letter elements).

berquist:
> Python 2.7 failing is probably itertools.accumulate.

It is, I've taken care of this problems and I'll make a PR to your branch soon. I like the idea of the splitter, it makes things much cleaner.

It will come in useful later since orbital energies that don't fit in the specified width in the end block get replaced with `****` and sometimes these run together, plus other cases I'm sure.

> The culprit output file is quite big and doesn't contain all interesting bits (e.g. two letter elements).

I noticed that it is quite big since the example is for the 958th AO, but it's fine, we can deal with big files especially since it isn't always obvious how to come up with smaller test cases that reproduce the desired behavior or failure.

srtlg:
I've attached a sample file. 

[sample.zip](https://github.com/cclib/cclib/files/821936/sample.zip)

berquist:
Great, thanks. ~32 MB is fine.

Another problem raised by this file is that there is no matching `$end` for `$rem`, so I am getting `StopIteration` because the `User Input` block is being parsed for something.

src/cclib/parser/qchemparser.py
```diff
@@ -18,6 +18,21 @@
 from . import utils
 
 
+class Splitter:
```
langner: Would it be possible to add a docstring describing the purpose of this class?

berquist: I added one in my PR.

berquist:
For reference, a PR to the PR: https://github.com/srtlg/cclib/pull/1

langner:
@berquist @srtlg Can we get this finished by Sunday? If not, I'm going to move it to v1.5.2.

Eric, if @srtlg is unresponsive then I'll clone his branch and we'll merge your changes in that way.

langner:
@berquist Can we continue in #381 please?

PR #359, opened 2017-03-16
QChem: new section for polarizabilities

berquist:
I added some new polarizability code to Q-Chem so naturally I would like to parse it :)

It also fixes the case (that only now appears) where calculations on a single atom print thermodynamic information, and that section would prevent parsing of anything that came afterward.

This needs to be merged *before* https://github.com/cclib/cclib-data/pull/53.

### DIFF
--- a/src/cclib/parser/qchemparser.py
+++ b/src/cclib/parser/qchemparser.py
@@ -799,33 +799,19 @@ class QChem(logfileparser.Logfile):
             if 'Polarizability (a.u.)' in line:
                 if not hasattr(self, 'polarizabilities'):
                     self.polarizabilities = []
-                polarizability = []
                 while 'Full Tensor' not in line:
                     line = next(inputfile)
                 self.skip_line(inputfile, 'blank')
-                for _ in range(3):
-                    line = next(inputfile)
-                    polarizability.append(line.split())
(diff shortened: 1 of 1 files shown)


src/cclib/parser/qchemparser.py
```diff
@@ -804,6 +808,26 @@ def extract(self, inputfile, line):
...
                 self.polarizabilities.append(numpy.array(polarizability))
 
+            # Static polarizability from finite difference.
+            if line.strip() == 'Static polarizability tensor [a.u.]':
+                if not hasattr(self, 'polarizabilities'):
+                    self.polarizabilities = []
+                polarizability = []
+                for _ in range(3):
+                    line = next(inputfile)
+                    polarizability.append(line.split())
```
langner: Consider using comprehension here:
polarizability = [next(inputfile).split() for _ in range(3)]

src/cclib/parser/qchemparser.py
```diff
@@ -804,6 +808,26 @@ def extract(self, inputfile, line):
...
+                self.polarizabilities.append(numpy.array(polarizability))
+
+            # Polarizability from responseman.
```
langner: This looks exactly like the previous - can you combine them?

PR #361, opened 2017-03-30
Typo in gaussianparser.py

renpj:


berquist:
Thanks!

Looks like we don't have tests for thermochemistry, which would have avoided this.

PR #366, opened 2017-04-29
Add alias cclib.ccopen and restore cclib.parser.ccopen

langner:


berquist:
So now we can write `from cclib import ccopen`, and perhaps in the future `from cclib import ccwrite`, etc?

langner:
That is my proposal - see #369 for the issue at hand. I am open to other suggestions, but can't think of any thing better in the long-term myself.

baoilleach:
Sorry for the delay. If this is accompanied by documentation on the API, then sure. Otherwise, it's just confusing to have multiple ways to access the same function. For example, I would remove similar aliases in any other init.pys and just have them at the top level if desired.

Re docs, you could just put the docstring before your change, and generate Sphinx docs from this. I think.

langner:
> Sorry for the delay. If this is accompanied by documentation on the API, then sure. Otherwise, it's just confusing to have multiple ways to access the same function. For example, I would remove similar aliases in any other init.pys and just have them at the top level if desired.
> 
> Re docs, you could just put the docstring before your change, and generate Sphinx docs from this. I think.

I think we can clean up later. I would like to be backwards compatible again if possible, and I think this change does that. With this change I would release 1.5.1 so that the newest version at least is backwards compatible.

Later we can think about cleaning the aliases up commenting what we want to remove when moving to 2.x.

langner:
> So now we can write from cclib import ccopen, and perhaps in the future from cclib import ccwrite, etc?

Also added cclib.ccwrite here.

Issue #371, opened 2017-05-06
Line ending problems


berquist:
I think https://github.com/cclib/cclib/pull/370 is causing an issue where `src/cclib/parser/data.py` is shown as having changed line endings for the entire file, and `git checkout -- src/cclib/parser/data.py` is not undoing this properly (git is silent).

When I do manage to get a different commit checked out and then switch to yet another commit, there are other line ending changes.

Until this patch, what I kept in my `~/.gitconfig` to avoid line ending problems (at least on Mac and Linux) is

```
[core]
    autocrlf = input
    safecrlf = warn
```

Is there a way to get around this checkout problem?

langner:
Yeah, seems like some of the endings in `data.py` were really weird. Maybe fixed by #372?

berquist:
I don't think https://github.com/cclib/cclib/pull/372 is going to fix it. When I comment out those lines in my gitconfig and clone a fresh copy of the repository, these are all the files that would require a line-ending change:

```
/home/eric/development
[eric@copper]$ \rm -rf cclib
/home/eric/development
[eric@copper]$ git clone https://github.com/cclib/cclib.git
Cloning into 'cclib'...
remote: Counting objects: 14662, done.
remote: Compressing objects: 100% (41/41), done.
remote: Total 14662 (delta 18), reused 1 (delta 1), pack-reused 14619
Receiving objects: 100% (14662/14662), 25.19 MiB | 20.10 MiB/s, done.
Resolving deltas: 100% (9689/9689), done.
/home/eric/development
[eric@copper]$ cd cclib
/home/eric/development/cclib (master|✚62)
[eric@copper]$ git st
## master...origin/master
 M data/GAMESS/basicFirefly8.0/C_bigbasis.out
 M data/GAMESS/basicFirefly8.0/dvb_gopt_a.inp
 M data/GAMESS/basicFirefly8.0/dvb_gopt_a.out
 M data/GAMESS/basicFirefly8.0/dvb_ir.inp
 M data/GAMESS/basicFirefly8.0/dvb_ir.out
 M data/GAMESS/basicFirefly8.0/dvb_raman.inp
 M data/GAMESS/basicFirefly8.0/dvb_raman.out
 M data/GAMESS/basicFirefly8.0/dvb_sp.inp
 M data/GAMESS/basicFirefly8.0/dvb_sp.out
 M data/GAMESS/basicFirefly8.0/dvb_td.inp
[...]

PR #372, opened 2017-05-08
Fix line endings in data.py

langner:


langner:
Closing as per discussion in #371

Issue #373, opened 2017-05-11
ORCA: no etsyms from TDDFT calculation


berquist:
```
  File "/usr/lib/python3.6/site-packages/cclib/parser/orcaparser.py", line 602, in extract
    for x in self.etsyms:
AttributeError: 'ORCA' object has no attribute 'etsyms'
```

Input:

```
! rks pbe def2-svp

%tddft
 nroots 24
 tda false
 end

* xyzfile 0 1 1-ttt_opt.xyz *
```

berquist:
It appears there are several printing differences, both in headers and content, when doing RPA (full TDDFT) compared to TDA.

berquist:
This was 2 separate issues:

1. headers are different for RPA and TDA
2. v4.0.0 doesn't print excitation coefficients, but v3.0.3 does.

Going to reopen this so I can change the empty string to nan in another PR.

langner:
So not are we all done with this?

berquist:
Yes, all done. Let me know if there's anything else that needs to be done this weekend.

PR #374, opened 2017-05-11
ORCA: handle RPA calculations, not just TDA

berquist:
Fixes https://github.com/cclib/cclib/issues/373.

This should be merged __before__ https://github.com/cclib/cclib-data/pull/54.

### DIFF
--- a/src/cclib/parser/orcaparser.py
+++ b/src/cclib/parser/orcaparser.py
@@ -552,7 +552,7 @@ class ORCA(logfileparser.Logfile):
             del self.tmp_atnames
 
         # Read TDDFT information
-        if line[0:18] == "TD-DFT/TDA EXCITED":
+        if any(x in line for x in ("TD-DFT/TDA EXCITED", "TD-DFT EXCITED")):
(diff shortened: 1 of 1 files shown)


src/cclib/parser/orcaparser.py
```diff
@@ -583,7 +583,11 @@ def extract(self, inputfile, line):
                     start = (int(start[:-1]), lookup[start[-1]])
                     end = line[10:17].strip()
                     end = (int(end[:-1]), lookup[end[-1]])
-                    contrib = float(line[35:47].strip())
+                    # Coeffients are not printed for RPA, only
+                    # TDA/CIS.
+                    contrib = line[35:47].strip()
+                    if contrib != '':
+                        contrib = float(contrib)
```
berquist: I am not a fan of this, it's just a temporary placeholder. In the case of no value for `contrib` (which should be a float), should we use `None` or `np.nan`? Then we can check for it in the regression test.

langner: I vote for NaN.

PR #379, opened 2017-05-15
Bump v1.5 to v1.5.1

langner:


langner:
FYI I uploaded to pypi as well.

PR #380, opened 2017-05-20
Separate possibly unspaced polarizability tensor elements

xymaxim:
In some rare cases DALTON yields huge negative polarizability values. ABACUS module prints to the output file following:

```
                           Static polarizabilities (au)
. . .
            Ey       0.000000-7211.390426    0.000000
```

Since here _xy_ and _yy_ tensor elements are stack together, with no spaces, calling parse() method causes `ValueError: setting an array element with a sequence` as expected.

With these changes in place, not only spacing but also minus sign is used to spit elements. Regexp helps us to distinguish between a real number and the exponent when polarizabilities are [written](https://github.com/cclib/cclib-data/blob/master/DALTON/DALTON-2015/trithiolane_polar_abalnr.out#L1183) in E-notation.

### DIFF
--- a/src/cclib/parser/daltonparser.py
+++ b/src/cclib/parser/daltonparser.py
@@ -10,8 +10,6 @@
 
 from __future__ import print_function
 
-import re
-
 import numpy
 
 from . import logfileparser
@@ -1076,7 +1074,7 @@ class DALTON(logfileparser.Logfile):
                 line = next(inputfile)
                 # Separate possibly unspaced huge negative polarizability tensor
                 # element and the left adjacent column from each other.
-                line = re.sub(r'(\d)-', r'\1 -', line)
+                line = line.replace('-', ' -')
                 polarizability.append(line.split()[1:])
             self.polarizabilities.append(numpy.array(polarizability))
 
@@ -1088,7 +1086,6 @@ class DALTON(logfileparser.Logfile):
             self.skip_lines(inputfile, ['d', 'directions', 'b'])
             for _ in range(3):
                 line = next(inputfile)
-                line = re.sub(r'(\d)-', r'\1 -', line)
                 polarizability.append(line.split()[1:])
             self.polarizabilities.append(numpy.array(polarizability))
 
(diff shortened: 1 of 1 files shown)


berquist:
Would you be able to provide us with an example output file that shows this problem so we can run regression tests against it?

xymaxim:
[huge_neg_polar_stat.txt](https://github.com/cclib/cclib/files/1017172/huge_neg_polar_stat.txt)
[huge_neg_polar_freq.txt](https://github.com/cclib/cclib/files/1017171/huge_neg_polar_freq.txt)

Please find attached output files for the static and dynamic polarizability calculations (at this moment I have access only to the DALTON2016.02 release). 

As you can see, ABACUS printed numbers in different ways. I was just looking through the original code and found (omitting file names and line numbers) the use of the following descriptors:
- `3F12.6` for **PROPE .POLARI
- `3(G14.7,1X)` for **PROPE **ABALNR

In the first case, we never get a number formatted in scientific notation and there is no need to use regexp here. Actually, such formatting allows to successfully parse numbers only in the range of -10^2 to 10^3 without this pull request. After merging the proposed changes, the range will also contain negative values up to, but not including, -10^4.  In the second case, enclosing `1X` eliminates the existence of unspaced tensor elements. Correct me if I am wrong.

Considering the above, I made a new commit. Despite the fact that my example is not very general and contains problems with calculation, personally I would prefer that the end-user has ability to parse and get these abnormal values rather than parser exception.

berquist:
First of all, thank you for the PR; it's good to know that someone other than me is using the DALTON parser for molecular properties. If there is any new functionality you'd like to see, feel free to make a feature request.

I will add these two files to our regression test suite. Don't worry about version coverage, since this part of their codebase hasn't changed in quite some time, and it is best to use the latest version for production anyway. Also don't worry about the warnings, which are due to DALTON not displaying all the MO coefficients after printing the total before SCF starts.

> In the second case, enclosing 1X eliminates the existence of unspaced tensor elements.

I thought it inserted a space (my Fortran is not so great), but the point is moot since it's easier to write parsing code in Python than formatting code in...almost any other language :)

PR #381, opened 2017-05-21
fix parsing coefficients

langner:


berquist:
We will need to pull in the changes from https://github.com/srtlg/cclib/pull/1.

langner:
> We will need to pull in the changes from srtlg#1.

Done.

Issue #377, opened 2017-05-12
Writing cjson using ccwrite not working for all test outputs


gaursagar:
[ccwrite_fails.txt](https://github.com/cclib/cclib/files/996040/ccwrite_fails.txt)

`ccwrite` is failing to write `cjson` for a number of test files due to various different reasons (listed in the attached file).
A fix for error specific to Molpro files is [here](https://github.com/gaursagar/cclib/pull/1). If this looks fine, I would issue a PR.

Steps to reproduce the error for Molpro:
`ccwrite cjson https://raw.githubusercontent.com/cclib/cclib/master/data/Molpro/basicMolpro2012/dvb_gopt.log`

Although ccwrite showed an error, I was not able to get unittests to fail by making changes in [testcjsonwriter.py](https://github.com/cclib/cclib/blob/master/test/io/testcjsonwriter.py)

pasenor:
There are some problems with the DALTON files.
apart from some typos (PR #466) the following falls

    $ python ../src/scripts/ccwrite cjson ./DALTON/basicDALTON-2013/C_bigbasis.out
    Attempting to parse ./DALTON/basicDALTON-2013/C_bigbasis.out
    ...
    Traceback (most recent call last):
      File "../src/scripts/ccwrite", line 101, in <module>
        main()
      File "../src/scripts/ccwrite", line 97, in main
        **ccwrite_kwargs)
      File "/home/gosha/venvs/py36/lib/python3.6/site-packages/cclib/io/ccio.py", line 339, in ccwrite
        outputobj = outputclass(ccdata, jobfilename=jobfilename, terse=terse, *args, **kwargs)
      File "/home/gosha/venvs/py36/lib/python3.6/site-packages/cclib/io/cjsonwriter.py", line 33, in __init__
        super(CJSON, self).__init__(ccdata, terse=terse, *args, **kwargs)
      File "/home/gosha/venvs/py36/lib/python3.6/site-packages/cclib/io/filewriter.py", line 56, in __init__
        self.obmol, self.pbmol = self._make_openbabel_from_ccdata()
      File "/home/gosha/venvs/py36/lib/python3.6/site-packages/cclib/io/filewriter.py", line 70, in _make_openbabel_from_ccdata
        obmol = makeopenbabel(self.ccdata.atomcoords,
    AttributeError: 'ccData_optdone_bool' object has no attribute 'atomcoords'
    
    
[...]

PR #382, opened 2017-05-26
Fix for `TypeError` in Molpro cjsonwriter

gaursagar:
```python
TypeError: Object of type 'int64' is not JSON serializable
``` 
Issue #377


langner:
I assume we'll cover this with the writer regression tests that will be added?

berquist:
How should we start doing regression tests for parts of the code other than the parsers?

berquist:
How should we start doing regression tests for parts of the code other than the parsers?

PR #386, opened 2017-06-01
Update ccread docstring

langner:


berquist:
Does the order in which the filenames are passed matter, in the case of attributes being overwritten?

langner:
> Does the order in which the filenames are passed matter, in the case of attributes being overwritten?

Yes, I suppose the order can matter. But I don't think this something we need to mention in the docstring. It seems more appropriate for the docs.

PR #387, opened 2017-06-14
Replaced ntpath with os.path in cjsonwriter.

gaursagar:
Patch based on @langner's  suggestion to use `os.path` in place of `ntpath`.


ATenderholt:
Looks good to me.

PR #388, opened 2017-06-16
Tests for filewriters

gaursagar:
Changes made:
1. Added a custom exception `MissingAttributeError` in `io.filewriter`. This ensures a sane error is raised in the absence of some critical attribute. 

Try running `ccwrite xyz data/MOPAC/h2o.out` with the current build. It currently fails with 
`AttributeError: 'ccData_optdone_bool' object has no attribute 'atomnos'`,
and with this update the following exception is raised
`cclib.io.filewriter.MissingAttributeError: Could not parse necessary attributes to write file.`

2. Added basic tests for `moldenwriter`. (More would be added).
3. Made changes in `moldenwriter.py` class to raise exceptions if some critical attribute is missing.

### DIFF
--- a/src/cclib/io/filewriter.py
+++ b/src/cclib/io/filewriter.py
@@ -26,9 +26,6 @@ class MissingAttributeError(Exception):
 
 class Writer(object):
     """Abstract class for writer objects.
-
-    Subclasses defined by cclib:
-        CJSON, CML, XYZ, MOLDEN
     """
 
     def __init__(self, ccdata, jobfilename=None, terse=False,
@@ -48,7 +45,10 @@ class Writer(object):
         self.terse = terse
 
         self.pt = PeriodicTable()
-        self.elements = [self.pt.element[Z] for Z in self.ccdata.atomnos]
+
+        # Check if required attributes are present.
+        if hasattr(self, 'required_attrs'):
+            self._check_required_attributes()
 
         # Open Babel isn't necessarily present.
         if has_openbabel:
@@ -92,6 +92,12 @@ class Writer(object):
                                         obbond.GetEndAtom().GetIndex(),
                                         obbond.GetBondOrder()))
         return bond_connectivities
+
+    def _check_required_attributes(self):
+        """Check if required attributes are present in ccdata."""
+        missing = ', '.join([x for x in self.required_attrs if not hasattr(self.ccdata, x)])
+        if len(missing) > 0:
+            raise MissingAttributeError('Could not parse required attributes to write file: ' + missing)
 
 
 if __name__ == "__main__":
--- a/src/cclib/io/filewriter.py
+++ b/src/cclib/io/filewriter.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2016, the cclib development team
+# Copyright (c) 2017, the cclib development team
 #
 # This file is part of cclib (http://cclib.github.io) and is distributed under
 # the terms of the BSD 3-Clause License.
@@ -18,6 +18,10 @@ except ImportError:
 from math import sqrt
 
 from cclib.parser.utils import PeriodicTable
+
+
+class MissingAttributeError(Exception):
+    pass
 
 
 class Writer(object):
--- a/src/cclib/io/filewriter.py
+++ b/src/cclib/io/filewriter.py
@@ -24,7 +24,7 @@ class Writer(object):
     """Abstract class for writer objects.
 
     Subclasses defined by cclib:
-        CJSON, CML, XYZ
+        CJSON, CML, XYZ, MOLDEN
     """
 
     def __init__(self, ccdata, jobfilename=None, terse=False,
--- a/src/cclib/io/moldenwriter.py
+++ b/src/cclib/io/moldenwriter.py
@@ -10,19 +10,11 @@
 import os.path
 
 from . import filewriter
-from .filewriter import MissingAttributeError
 from cclib.parser import utils
 
 
 class MOLDEN(filewriter.Writer):
     """A writer for MOLDEN files."""
-
-    def _check_required_attributes(self, required):
-        """Check if required attributes are present in ccdata"""
-        missing = [ x for x in required if not hasattr(self.ccdata, x) ]
-        if len(missing) > 0:
-            missing = ' '.join(missing)
-            raise MissingAttributeError('Could not parse required outputs to write molden file: ' + missing)
 
     def _title(self, path):
         """Return filename without extension to be used as title."""
@@ -32,7 +24,8 @@ class MOLDEN(filewriter.Writer):
     def _coords_from_ccdata(self, index):
         """Create [Atoms] section using geometry at the given index."""
         self._check_required_attributes(['atomcoords', 'atomnos', 'natom'])
-
+        
+        elements = [self.pt.element[Z] for Z in self.ccdata.atomnos]
         atomcoords = self.ccdata.atomcoords[index]
         atomnos = self.ccdata.atomnos
         nos = range(self.ccdata.natom)
@@ -40,7 +33,7 @@ class MOLDEN(filewriter.Writer):
         # element_name number atomic_number x y z
         atom_template = '{:2s} {:5d} {:2d} {:12.6f} {:12.6f} {:12.6f}'
         lines = []
-        for element, no, atomno, (x, y, z) in zip(self.elements, nos, atomnos,
+        for element, no, atomno, (x, y, z) in zip(elements, nos, atomnos,
                                                   atomcoords):
             lines.append(atom_template.format(element, no + 1, atomno,
                                               x, y, z))
--- a/src/cclib/io/moldenwriter.py
+++ b/src/cclib/io/moldenwriter.py
@@ -10,6 +10,7 @@
 import os.path
 
 from . import filewriter
+from .filewriter import MissingAttributeError
 from cclib.parser import utils
 
 
@@ -23,6 +24,11 @@ class MOLDEN(filewriter.Writer):
 
     def _coords_from_ccdata(self, index):
         """Create [Atoms] section using geometry at the given index."""
+
+        if not (hasattr(self.ccdata, 'atomcoords') and
+                hasattr(self.ccdata, 'atomnos') and
+                hasattr(self.ccdata, 'natom')):
+                raise MissingAttributeError('Could not parse necessary outputs to write molden file.')
 
         atomcoords = self.ccdata.atomcoords[index]
         atomnos = self.ccdata.atomnos
@@ -127,8 +133,9 @@ class MOLDEN(filewriter.Writer):
         molden_lines = ['[Molden Format]']
 
         # Title of file.
-        molden_lines.append('[Title]')
-        molden_lines.append(self._title(self.jobfilename))
+        if self.jobfilename is not None:
+            molden_lines.append('[Title]')
+            molden_lines.append(self._title(self.jobfilename))
 
         # Coordinates for the Electron Density/Molecular orbitals.
         # [Atoms] (Angs|AU)
--- a/src/cclib/io/moldenwriter.py
+++ b/src/cclib/io/moldenwriter.py
@@ -5,7 +5,7 @@
 # This file is part of cclib (http://cclib.github.io) and is distributed under
 # the terms of the BSD 3-Clause License.
 
-"""A writer for MOLDEN format files."""
+"""A writer for Molden format files."""
 
 import os.path
 
@@ -13,8 +13,8 @@ from . import filewriter
 from cclib.parser import utils
 
 
-class MOLDEN(filewriter.Writer):
-    """A writer for MOLDEN files."""
+class MoldenWriter(filewriter.Writer):
+    """A writer for Molden files."""
 
     required_attrs = ['atomcoords', 'atomnos', 'natom']
 
@@ -124,7 +124,7 @@ class MOLDEN(filewriter.Writer):
         return lines
 
     def generate_repr(self):
-        """Generate the MOLDEN representation of the logfile data."""
+        """Generate the Molden representation of the logfile data."""
 
         molden_lines = ['[Molden Format]']
 
--- a/src/cclib/io/moldenwriter.py
+++ b/src/cclib/io/moldenwriter.py
@@ -17,6 +17,13 @@ from cclib.parser import utils
 class MOLDEN(filewriter.Writer):
     """A writer for MOLDEN files."""
 
+    def _check_required_attributes(self, required):
+        """Check if required attributes are present in ccdata"""
+        missing = [ x for x in required if not hasattr(self.ccdata, x) ]
+        if len(missing) > 0:
+            missing = ' '.join(missing)
+            raise MissingAttributeError('Could not parse required outputs to write molden file: ' + missing)
+
     def _title(self, path):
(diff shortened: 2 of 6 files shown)


src/cclib/io/filewriter.py
```diff
@@ -20,11 +20,15 @@
...
 
 
+class MissingAttributeError(Exception):
+    """Raised when some necessary attribute is not available."""
+
+
 class Writer(object):
     """Abstract class for writer objects.
 
     Subclasses defined by cclib:
```
langner: Let's not refer to subclasses in the parent.

src/cclib/io/filewriter.py
```diff
@@ -44,7 +48,10 @@ def __init__(self, ccdata, jobfilename=None, terse=False,
         self.terse = terse
 
         self.pt = PeriodicTable()
-        self.elements = [self.pt.element[Z] for Z in self.ccdata.atomnos]
+        if hasattr(self.ccdata, 'atomnos'):
```
langner: Why is `atomnos` necessarily a required attribute?

src/cclib/io/moldenwriter.py
```diff
@@ -127,8 +133,9 @@ def generate_repr(self):
         molden_lines = ['[Molden Format]']
 
         # Title of file.
-        molden_lines.append('[Title]')
-        molden_lines.append(self._title(self.jobfilename))
+        if self.jobfilename is not None:
```
langner: So, in Molden is [Title] optional?

test/io/testfilewriter.py
```diff
@@ -20,15 +20,30 @@
 
 class WriterTest(unittest.TestCase):
 
+    def setUp(self):
+        self.MissingAttributeError = cclib.io.filewriter.MissingAttributeError
```
langner: If you want to make an alias for an exception in another module, better at module level. But I also see nothing wrong with using the entire path.

src/cclib/io/moldenwriter.py
gaursagar: Yes, it is optional.

test/io/testfilewriter.py
```diff
@@ -20,15 +20,30 @@
...
+
     def test_init(self):
```
langner: There's no init specific to MoldenWriter, so why do we need to test this?

src/cclib/io/filewriter.py
gaursagar: Any written file must have the coordinates. If the coordinates could not be parsed, there would be nothing to write.

test/io/testfilewriter.py
```diff
@@ -20,15 +20,30 @@
...
         """Does the class initialize correctly?"""
-        fpath = os.path.join(__datadir__, "data/ADF/basicADF2007.01/dvb_gopt.adfout")
+        fpath = os.path.join(__datadir__,
```
langner: You don't need to actually parse a file just to test the initializer. You can use an empty ccData object for that, right?

test/io/testmoldenwriter.py
```diff
@@ -0,0 +1,50 @@
...
+
+__filedir__ = os.path.dirname(__file__)
+__filepath__ = os.path.realpath(__filedir__)
+__datadir__ = os.path.join(__filepath__, "..", "..")
+
+
+class MOLDENTest(unittest.TestCase):
+
+    def setUp(self):
+        self.molden = cclib.io.MOLDEN
```
langner: Also here no need for object level aliases.

test/io/testmoldenwriter.py
```diff
@@ -0,0 +1,50 @@
...
+        self.MissingAttributeError = cclib.io.filewriter.MissingAttributeError
+
+    def test_init(self):
```
langner: MoldenWriter does not have it's own initializer, so do you need to test this again?

src/cclib/io/moldenwriter.py
```diff
@@ -24,6 +25,11 @@ def _title(self, path):
     def _coords_from_ccdata(self, index):
         """Create [Atoms] section using geometry at the given index."""
 
+        if not (hasattr(self.ccdata, 'atomcoords') and
+                hasattr(self.ccdata, 'atomnos') and
+                hasattr(self.ccdata, 'natom')):
+                raise MissingAttributeError('Could not parse necessary outputs to write molden file.')
+
```
ATenderholt: Would be helpful to know which attribute is missing. Perhaps something like:

```
required = [ 'atomcoords', 'atoms', 'atom']
missing = [ x for x in required if not hasattr(self.ccdata, x) ]
if len(missing) > 0:
   raise MissingAttributeError('Could not parse necessary outputs to write molden file:" + missing)
```

src/cclib/io/filewriter.py
langner: While that might be true for Molden, is it necessarily true for all future writers?

test/io/testfilewriter.py
gaursagar: This unit-test is common across all file writers, so I used it as is.

src/cclib/io/filewriter.py
gaursagar: It may not be. As you have suggested, it is better each writer takes care whether all necessary attributes are present.

test/io/testmoldenwriter.py
gaursagar: I guess not. This can be removed.

src/cclib/io/moldenwriter.py
gaursagar: Fair point. I would make the change.

test/io/testfilewriter.py
langner: Can you fix that then?

gaursagar: Sure. I would make a seperate PR.

src/cclib/io/filewriter.py
gaursagar: This calculation has been moved to individual `filewriters`. `filewriter.__init__` should not depend upon any ccData attributes. PR: https://github.com/cclib/cclib/pull/389

test/io/testfilewriter.py
gaursagar: Removed this line. The exception is now being imported from the filewriter class https://github.com/cclib/cclib/pull/388/commits/b7a09f6219fb345918d942b6e2ca3baa58be1b3d#diff-b77b9f83bdfaba6902ee0d0d99d8ac7dR14

test/io/testmoldenwriter.py
gaursagar: Removed.

test/io/testmoldenwriter.py
gaursagar: Removed.

src/cclib/io/moldenwriter.py
gaursagar: Made the change here https://github.com/cclib/cclib/pull/388/files#diff-599bcb2b83c5ef046682dfaa37fd74e1R20.

src/cclib/io/moldenwriter.py
```diff
@@ -10,19 +10,28 @@
...
 
 from . import filewriter
+from .filewriter import MissingAttributeError
 from cclib.parser import utils
 
 
 class MOLDEN(filewriter.Writer):
     """A writer for MOLDEN files."""
 
+    def _check_required_attributes(self, required):
```
TDI-tenderholt: I was thinking this should go in the base ```Writer``` class, and then the required attributes could be passed into the constructor:

```
def __init__(self, ccdata, jobfilename=None, terse=False,
                 required_attrs=[], *args, **kwargs):
```

You'll have to make sure that this doesn't break anything downstream though.

gaursagar: Yes, this would be more useful. I'll make the change. Though I don't think it would break anything.

gaursagar: I moved this function to the base class. For now, I am avoiding passing the list of required attributes to base class though.

src/cclib/io/filewriter.py
langner: It looks like you still did not remove reference to subclasses here.

langner: It looks like you still did not remove reference to subclasses here.

src/cclib/io/filewriter.py
```diff
@@ -89,6 +92,13 @@ def _make_bond_connectivity_from_openbabel(self, obmol):
                                         obbond.GetBondOrder()))
         return bond_connectivities
 
+    def _check_required_attributes(self, required):
+        """Check if required attributes are present in ccdata"""
```
langner: Please end sentences in periods.

src/cclib/io/filewriter.py
```diff
@@ -89,6 +92,13 @@ def _make_bond_connectivity_from_openbabel(self, obmol):
                                         obbond.GetBondOrder()))
         return bond_connectivities
 
+    def _check_required_attributes(self, required):
```
langner: `required` should be a class variable, so you don't need to pass it to this method.

src/cclib/io/filewriter.py
```diff
@@ -89,6 +92,13 @@ def _make_bond_connectivity_from_openbabel(self, obmol):
...
+        """Check if required attributes are present in ccdata"""
+        missing = [x for x in required if not hasattr(self.ccdata, x)]
+        if len(missing) > 0:
+            missing = ' '.join(missing)
```
langner: It's bad practive to re-use the same variable name. I think it's safe to include the joining inline in the next line anyway.

src/cclib/io/moldenwriter.py
```diff
@@ -16,22 +16,35 @@
...
+    def __init__(self, ccdata, *args, **kwargs):
+        """Initialize the MOLDEN writer object.
+
+        Inputs:
+          ccdata - An instance of ccData, parsed from a logfile.
+        """
+        # Call the __init__ method of the superclass
+        super(MOLDEN, self).__init__(ccdata, *args, **kwargs)
+
+        required_attrs = ['atomcoords', 'atomnos', 'natom']
```
langner: As we discussed and agreed, this should be a class variable.

src/cclib/io/moldenwriter.py
```diff
@@ -16,22 +16,35 @@
...
+        # Check if all required attributes are present in ccData.
+        self._check_required_attributes(required_attrs)
```
langner: As discussed, this should be called from the initializer of the base class.

src/cclib/io/moldenwriter.py
```diff
@@ -16,22 +16,35 @@
 class MOLDEN(filewriter.Writer):
     """A writer for MOLDEN files."""
 
+    def __init__(self, ccdata, *args, **kwargs):
```
langner: There's no need currently for a custom initializer in this class, right?

test/io/testmoldenwriter.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class MOLDENTest(unittest.TestCase):
+
+    def test_missing_attributes(self):
+        """Check if MissingAttributeError is raised as expected."""
+        fpath = os.path.join(__datadir__,
+                             "data/ADF/basicADF2007.01/dvb_gopt.adfout")
+        data = cclib.io.ccopen(fpath).parse()
+        del data.atomcoords
```
langner: Could you do this for all required attributes?

src/cclib/io/filewriter.py
gaursagar: Removed for sure now : ) https://github.com/cclib/cclib/pull/388/commits/3b5e1457f922383f9bf9090db4c94fef0c3c03b9

src/cclib/io/moldenwriter.py
gaursagar: We had agreed to put the attributes check within init.

src/cclib/io/moldenwriter.py
gaursagar: Ah.. I had remembered a class constant. Will change it.

src/cclib/io/moldenwriter.py
gaursagar: Okay. I had misunderstood (even after your explanation)

src/cclib/io/moldenwriter.py
langner: Doesn't matter to me if it's a variable or constant (there are no constants in Python), but a something attached to the class, not the object.

src/cclib/io/moldenwriter.py
langner: Yes, but in the base class.

test/io/testmoldenwriter.py
gaursagar: Done.

src/cclib/io/moldenwriter.py
gaursagar: Fixed.

src/cclib/io/moldenwriter.py
gaursagar: Removed the init, made `required_attrs` a class constant, and moved the check to base class.

src/cclib/io/filewriter.py
gaursagar: Fixed.

src/cclib/io/filewriter.py
gaursagar: Fixed.

test/io/testmoldenwriter.py
ATenderholt: @langner : did you mean this approach or three separate tests?

langner: Yeah, what I had in mind is testing each one separately. That covers this case but is broader.

gaursagar: By separately did you mean a single test with incrementally removing,
```python
def test_missing():
    ...
    del data.atomnos 
    # Test with one attr deleted.
    with self.assertRaises(MissingAttributeError):
        cclib.io.moldenwriter.MoldenWriter(data)
    del data.natom
    # Test with two attrs deleted.
    with self.assertRaises(MissingAttributeError):
        cclib.io.moldenwriter.MoldenWriter(data)
    ...
```
or different tests each removing one required attribute.
```python
def test_missing_attr1():
    ...
    del data.natom
    # Test with natom deleted.
    with self.assertRaises(MissingAttributeError):
        cclib.io.moldenwriter.MoldenWriter(data)

def test_missing_attr2():
    ...
    del data.atomnos 
    # Test with atomnos deleted.
    with self.assertRaises(MissingAttributeError):
        cclib.io.moldenwriter.MoldenWriter(data)
    ...
```

ATenderholt: I think there should be different tests that each remove one attribute. If it's just a single test, for the subsequent ```assertRaises```, you don't know if the exception is being raised due to the attribute you've just deleted (e.g. ```data.natom```) or a previous one (e.g. ```data.atomnos```).

langner: +1 to Adam's comment

gaursagar: Seperated the tests as follows:
```python
def test_missing_attribute_error(self):
    ...
    required_attrs = ['atomcoords', 'atomnos', 'natom']
    for attr in required_attrs:
        data = cclib.io.ccopen(fpath).parse()
        delattr(data, attr)
        with self.assertRaises(MissingAttributeError):
            cclib.io.moldenwriter.MOLDEN(data)
```

PR #389, opened 2017-06-21
Moving elements list calculation from init to individual filewriters.

gaursagar:
Calculations based on `ccData` should be done in respective filewriters. 

This would help with testing filewriters with empty `ccData` objects (Ref. https://github.com/cclib/cclib/pull/388#pullrequestreview-44604158), 
and better handling required attributes at filewriter level (Ref. https://github.com/cclib/cclib/pull/388#pullrequestreview-44602730).

### DIFF
--- a/src/cclib/io/cmlwriter.py
+++ b/src/cclib/io/cmlwriter.py
@@ -43,16 +43,17 @@ class CML(filewriter.Writer):
         if self.jobfilename is not None:
             d['id'] = self.jobfilename
         _set_attrs(molecule, d)
-
+        
         # Form the listing of all the atoms present.
         atomArray = ET.SubElement(molecule, 'atomArray')
-        if hasattr(self.ccdata, 'atomcoords'):
+        if hasattr(self.ccdata, 'atomcoords') and hasattr(self.ccdata, 'atomnos'):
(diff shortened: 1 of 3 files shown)


TDI-tenderholt:
Looks good to me. Should ```self.pt``` also be moved to individual classes?

gaursagar:
`self.pt` can stay because it does not depend on any parsed data.

src/cclib/io/cmlwriter.py
```diff
@@ -43,6 +43,9 @@ def generate_repr(self):
         if self.jobfilename is not None:
             d['id'] = self.jobfilename
         _set_attrs(molecule, d)
+        
+        if hasattr(self.ccdata, 'atomnos'):
+            elements = [self.pt.element[Z] for Z in self.ccdata.atomnos]
```
TDI-tenderholt: Is ```elements``` only used when writing the ```atomArray```? If so, move it just above the ```for atomid ...``` and combine the two ```if hasattr...``` statements.

gaursagar: @TDI-tenderholt yes, thats the only place it was used. I made the change and squashed the commit.

Issue #390, opened 2017-06-26
New attribute scfinfo for microiteration data.


gaursagar:
For the Molden writer, energies after each scf convergence step are required.

It is the `TOTAL ENERGY` column in `RHF SCF CALCULATION` section in GAMESS  outputs. 
```
 ITER EX DEM     TOTAL ENERGY        E CHANGE  DENSITY CHANGE    DIIS ERROR
   1  0  0      -74.7981539269   -74.7981539269   0.585814622   0.000000000
   2  1  0      -74.9499878493    -0.1518339224   0.180197673   0.000000000
   3  2  0      -74.9626905270    -0.0127026777   0.060203035   0.000000000
   4  3  0      -74.9640834596    -0.0013929326   0.020782027   0.000000000
   5  4  0      -74.9642853920    -0.0002019324   0.007719362   0.000000000
   6  0  0      -74.9643205282    -0.0000351362   0.005106732   0.000000000
   7  1  0      -74.9643287842    -0.0000082560   0.000126896   0.000000000
   8  2  0      -74.9643287911    -0.0000000069   0.000045747   0.000000000
   9  3  0      -74.9643287923    -0.0000000011   0.000017697   0.000000000
  10  4  0      -74.9643287925    -0.0000000002   0.000007301   0.000000000
  11  5  0      -74.9643287925    -0.0000000000   0.000003237   0.000000000
  12  6  0      -74.9643287925    -0.0000000000   0.000001437   0.000000000
  13  7  0      -74.9643287925    -0.0000000000   0.000000639   0.000000000
                       ^                ^              ^
                    Needed;      Could be used;     Parsed
                  not parsed      not parsed      as scfvalues
```
(from [data/GAMESS/basicGAMESS-US2014/water_mp2.out](https://github.com/cclib/cclib/blob/master/data/GAMESS/basicGAMESS-US2014/water_mp2.out#L323)). These were initially omitted from the GAMESS parser because it uses density change for convergence.

Adam and Karol have proposed parsing these into a new attribute `scfinfo`.

langner:
The new attribute route looks the best, but other options include:
1. Extending scftargets/scfvalues to include not only what is used to determine convergence (I don't think we reflect that accurately right now anyway). Then again, we won't have target value for this that are not used, and this would not be backwards compatible.
2. Extend only scfvalues, maybe at the end of the array so it's clear it's not associated with targets.
3. ?

gaursagar:
[...]

PR #391, opened 2017-06-27
Diff tests for molden writer.

gaursagar:
This PR adds:
- tests for Molden writer,
- some fixes to the Molden writer to match formatting,
- a Molden reformatter in utils,
- reference outputs from the Molden program.

`[SCFCONV]` section is currently omitted from the output and while checking until https://github.com/cclib/cclib/issues/390 is resolved.

(PS. the branch used for this PR is derived from https://github.com/cclib/cclib/pull/388, hence the extra commits.)


src/cclib/io/moldenwriter.py
```diff
@@ -16,23 +17,53 @@
...
+        required_attrs = ['atomcoords', 'atomnos', 'natom']
+        # Check if all required attributes are present in ccData.
+        self._check_required_attributes(required_attrs)
+
     def _title(self, path):
         """Return filename without extension to be used as title."""
         title = os.path.basename(os.path.splitext(path)[0])
         return title
 
+    def _round_molden(self, num, p=6):
```
ATenderholt: Does ```p``` always equal ```6```? If so, I don't see a reason for the default argument.

Also would be good to see this decorated with ```@staticmethod``` and some unit testing on it to handle cases you've seen in the molden output.

gaursagar: I chose `p` based on the diffs, it could be a different. For the files tested so far 6 seems to be working fine.

src/cclib/parser/utils.py
```diff
@@ -194,6 +195,72 @@ def split(self, line, truncate=True):
...
+
+            # Exclude SCFCONV section until issue #390 is resolved.
+            # https://github.com/cclib/cclib/issues/390
+            if '[scfconv]' in line.lower():
+                break
+
+            # Although Molden format specifies Sym in [MO] section,
+            # the Molden program does not print it.
+            if 'sym' in line.lower():
+                # Skip the line.
```
langner: Unnecessary comment.

src/cclib/io/moldenwriter.py
```diff
@@ -78,6 +109,32 @@ def _scfconv_from_ccdata(self):
 
         return lines
 
+    def _rearrange_mocoeffs(self, mocoeffs):
```
gaursagar: The ordering issue occurs in only `f` orbitals. `d`, `g` and others were fine. (I had exhaustively checked the files in `data` folder.)

src/cclib/io/moldenwriter.py
langner: This feels like a function not really specific to the writer, so should be module level.

src/cclib/io/filewriter.py
```diff
@@ -20,11 +20,15 @@
...
+class MissingAttributeError(Exception):
+    pass
+
+
 class Writer(object):
     """Abstract class for writer objects.
 
     Subclasses defined by cclib:
-        CJSON, CML, XYZ
+        CJSON, CML, XYZ, MOLDEN
```
langner: Let's not mention subclasses in the parent class.

src/cclib/io/filewriter.py
```diff
@@ -89,6 +92,13 @@ def _make_bond_connectivity_from_openbabel(self, obmol):
                                         obbond.GetBondOrder()))
         return bond_connectivities
 
+    def _check_required_attributes(self, required):
```
langner: As discussed earlier, `required` should be a class attribute.

src/cclib/io/filewriter.py
```diff
@@ -89,6 +92,13 @@ def _make_bond_connectivity_from_openbabel(self, obmol):
...
+        """Check if required attributes are present in ccdata"""
+        missing = [x for x in required if not hasattr(self.ccdata, x)]
+        if len(missing) > 0:
+            missing = ' '.join(missing)
+            raise MissingAttributeError('Could not parse required outputs to write molden file: ' + missing)
```
langner: Don't mention molden in the base class.

src/cclib/io/moldenwriter.py
```diff
@@ -8,31 +8,62 @@
 """A writer for MOLDEN format files."""
 
 import os.path
+import math
 
 from . import filewriter
 from cclib.parser import utils
 
 
+def round_molden(num, p=6):
```
langner: Please add unit test for this function.

src/cclib/io/moldenwriter.py
```diff
@@ -8,31 +8,62 @@
...
+    """Molden style number rounding in [Atoms] section."""
+    # Digit at pth position after dot.
+    p_digit = math.floor(abs(num) * 10 ** p) % 10
+    # If the 6th digit after dot is greater than 5, but not 7.
+    if p_digit > 5 and p_digit != 7:
+        # Round the number upto 6th place.
+        return round(num, p)
+    # Else truncate at 6th digit after dot.
```
langner: nit: combine comment on lines 21, 24 and 25 into one longer comment.

src/cclib/io/moldenwriter.py
```diff
@@ -8,31 +8,62 @@
...
     """A writer for MOLDEN files."""
 
+    def __init__(self, ccdata, *args, **kwargs):
+        """Initialize the MOLDEN writer object.
+
+        Inputs:
+          ccdata - An instance of ccData, parsed from a logfile.
+        """
+        # Call the __init__ method of the superclass
+        super(MOLDEN, self).__init__(ccdata, *args, **kwargs)
```
langner: There are only two keyword arguments in the base class initializer - any reason not to just expose them in MOLDEN as well?

src/cclib/parser/utils.py
```diff
@@ -194,6 +195,70 @@ def split(self, line, truncate=True):
         return elements
 
 
+class MoldenReformatter:
```
langner: Perhaps this is better placed in the moldenwriter module?

test/io/testmoldenwriter.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2017, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
+
+"""Unit tests for writer moldenwriter module."""
```
langner: Remove 'writer'.

src/cclib/io/moldenwriter.py
gaursagar: Are you referring to `jobfilename=None`, and `terse=False`? These are used by other writers, I see no particular problem in changing this however. (This can be done a separate PR)

langner: Please do it here. If you think molden does need them, then do not pipe them through at all (they do have default values).

gaursagar: `jobfilename` is needed. `terse` is very specific to `CJSON writer`, I suggest moving it from the writer itself.

langner: Yeah, that's reasonable.

src/cclib/parser/utils.py
gaursagar: Removed and moved this class to `moldenwriter`

src/cclib/io/filewriter.py
gaursagar: Removed all references to subclasses.

src/cclib/io/filewriter.py
gaursagar: Addressed. `required_attrs` is now created as a class attribute in the [subclass](https://github.com/cclib/cclib/pull/391/files#diff-599bcb2b83c5ef046682dfaa37fd74e1R33).

src/cclib/io/filewriter.py
gaursagar: Removed. New error message would look something like 
`Could not parse required outputs to write file: gbasis, natom`

src/cclib/io/moldenwriter.py
gaursagar: Tests added [here](https://github.com/cclib/cclib/pull/391/files#diff-b77b9f83bdfaba6902ee0d0d99d8ac7dR79) to cover relevant cases of negative, positive integers and floats.

src/cclib/io/moldenwriter.py
gaursagar: Merged to a single comment:
```python
    # If the 6th digit after dot is greater than 5, but is not 7,
    # round the number upto 6th place.
    # Else truncate at 6th digit after dot.
```

src/cclib/io/moldenwriter.py
gaursagar: These variables are assigned non-default values in [`ccio`](https://github.com/cclib/cclib/blob/master/src/cclib/io/ccio.py#L337) only when called via `ccwrite`. 

Initialising any subclass of the `filewriter` directly assigns these arguments their default values. So it is not necessary to expose them in the subclasses.

src/cclib/parser/utils.py
gaursagar: Done.

test/io/testmoldenwriter.py
gaursagar: Removed.

src/cclib/io/filewriter.py
```diff
@@ -53,6 +52,10 @@ def __init__(self, ccdata, jobfilename=None, terse=False,
             self.obmol, self.pbmol = self._make_openbabel_from_ccdata()
             self.bond_connectivities = self._make_bond_connectivity_from_openbabel(self.obmol)
 
+        # Check if required attributes are present.
```
langner: You can also set `required_attrs` in the base class to the empty list, and then you don't need to do this check.

src/cclib/io/filewriter.py
```diff
@@ -89,6 +92,14 @@ def _make_bond_connectivity_from_openbabel(self, obmol):
                                         obbond.GetBondOrder()))
         return bond_connectivities
 
+    def _check_required_attributes(self):
+        """Check if required attributes are present in ccdata"""
```
langner: nit: add period to end of docstring.

src/cclib/io/moldenwriter.py
```diff
@@ -8,31 +8,53 @@
 """A writer for MOLDEN format files."""
 
 import os.path
+import math
 
-from . import filewriter
+from decimal import Decimal
```
langner: Perhaps just import the whole module `decimal`, in the block above?

src/cclib/io/moldenwriter.py
```diff
@@ -8,31 +8,53 @@
...
+    if num >= 0:
+        return math.floor(num * 10 ** p) / 10 ** p
+    else:
+        return math.ceil(num * 10 ** p) / 10 ** p
 
 
 class MOLDEN(filewriter.Writer):
     """A writer for MOLDEN files."""
 
+    required_attrs = ['atomcoords', 'atomnos', 'natom']
```
langner: Since this is not supposed to change, perhaps tuple instead of list?

src/cclib/io/moldenwriter.py
```diff
@@ -8,31 +8,53 @@
...
         atomnos = self.ccdata.atomnos
         nos = range(self.ccdata.natom)
 
         # element_name number atomic_number x y z
         atom_template = '{:2s} {:5d} {:2d} {:12.6f} {:12.6f} {:12.6f}'
         lines = []
-        for element, no, atomno, (x, y, z) in zip(self.elements, nos, atomnos,
+        for element, no, atomno, (x, y, z) in zip(elements, nos, atomnos,
                                                   atomcoords):
+            x = round_molden(x)
```
langner: You can avoid x, y and z in the loop definition here by doing `x, y, z = [round_molden(c) for c in coord]` or functioanlly `x, y, z = map(round_molden, coord)`.

src/cclib/io/moldenwriter.py
langner: Could you extend the docstring with this information?

src/cclib/io/moldenwriter.py
```diff
@@ -78,6 +100,32 @@ def _scfconv_from_ccdata(self):
 
         return lines
 
+    def _rearrange_mocoeffs(self, mocoeffs):
+        """Rearrange cartesian F functions in mocoeffs."""
+        # Molden's order:
```
langner: Include these comments in the docstring.

src/cclib/io/filewriter.py
gaursagar: `filewriter` is the base class. Setting it to an empty list here would clear the values set in derived classes. One way around is to set it in the `__init__` of the derived class, but we had earlier agreed not to do so.

src/cclib/io/moldenwriter.py
gaursagar: Done.

src/cclib/io/moldenwriter.py
gaursagar: Done.

src/cclib/io/moldenwriter.py
gaursagar: Rewrote to use `x, y, z  =  map(round_molden, coord)`

src/cclib/io/moldenwriter.py
gaursagar: Done.

src/cclib/io/moldenwriter.py
gaursagar: Done.

src/cclib/io/filewriter.py
langner: The child would overwrite the class member of the base class. Like so:
```python
class A(object):
  x = []
class B(A):
  x = [1,2,3]
print B().x
[1, 2, 3]
```

gaursagar:
```python
class A(object):
    def __init__(self):
        self.var = ()
        self.foo()
        
class B(A):
    var = ('aloha')

obj = B()
>()
```
I was doing this way. Anyways, fixed now :)

langner: Well, in your example A.var is not a class attribute.

src/cclib/io/filewriter.py
```diff
@@ -52,6 +54,9 @@ def __init__(self, ccdata, jobfilename=None, terse=False,
             self.obmol, self.pbmol = self._make_openbabel_from_ccdata()
             self.bond_connectivities = self._make_bond_connectivity_from_openbabel(self.obmol)
 
+        # Check if required attributes are present.
```
langner: This comment is redundant.

Issue #258, opened 2016-04-08
Support for molecular point group symmetry


ghutchis:
While mosyms and vibsyms exist, the molecular point group symmetry "sym" is not parsed. This is extremely useful to know if a calculation was performed in a reduced symmetry.

bwang2453:
Maybe we could put the molecular symmetry in the metadata.

langner:
Sure, I can see that as being useful.

berquist:
I have started parsing the full and Abelian/reduced point group from every program. I like the idea of putting both in the metadata, once it's available in all the parsers.

langner:
Cool, will you finish before the 1.5 release (mid August)?

berquist:
Thinking about v2.0, I would really like it if this was extended to more than just a string `d2h`, `c2v`, `cs`, and so on. In addition to determining the point group, there should be the ability to parse the number of doubly-occupied/singly-occupied/empty/total irreps, or calculate them if enough information is available.

This would enable, for the future, proper handling of symmetry-adapted orbitals for `mocoeffs` and MCSCF calculations.

ghutchis:
I'd recommend optional integration with https://github.com/mcodev31/libmsym

libmsym has a small Python wrapper and can produce much of this, including symmetry-projected orbitals.

ghutchis:
Not sure how much of your suggestion would need to be in cclib vs. already in libmsym, but certainly handling the irreps, etc. is already implemented.

berquist:
Yes, I'd definitely like that. I think it would be a good candidate as a bridge, rather than a hard dependency. I don't think any point group math needs to be added to cclib directly, since programs are usually nice enough to print the irrep ordering for you, but if we offer any symmetry-based methods in the future, libmsym would probably be needed.

berquist:
The first part is done in https://github.com/cclib/cclib/pull/839.

PR #392, opened 2017-07-03
Add point group detection

berquist:
This isn't currently full symmetry handling for all appropriate attributes, just the string representing the point group.

As mentioned in https://github.com/cclib/cclib/issues/258, these are currently in the metadata. `symmetry_full` is the real or molecular point group, and `symmetry_abelian` is the point group that the calculation is actually using. There are a few cases where only one is printed; right now, this PR is _not_ smart about fixing this.

If symmetry is not parsed, it assumes that both point groups are C1, so no symmetry is present.

Tests were added to `test/data/testSP.py`.

What things are called and so on is of course open for discussion, just figured this would be a good starting place.

The regression PR is https://github.com/cclib/cclib-data/pull/56, which needs to go in before this will pass.

---

Additionally,

- `metadata['package_version']` is added for ADF and GAMESS-UK. MOPAC is the only one missing this, as it is missing `metadata` entirely.
- `metadata['basis_set']` now defaults to the empty string.
- `mosyms` are now parsed from ORCA.
- All parsers now test to see if the `metadata` fields have been filled out (except for MOPAC).

### DIFF
--- a/src/cclib/parser/adfparser.py
+++ b/src/cclib/parser/adfparser.py
@@ -105,9 +105,6 @@ class ADF(logfileparser.Logfile):
         SCFCNV, SCFCNV2 = list(range(2))  # used to index self.scftargets[]
         maxelem, norm = list(range(2))  # used to index scf.values
 
-    def after_parsing(self):
-        super(ADF, self).after_parsing()
-
     def extract(self, inputfile, line):
         """Extract information from the file object inputfile."""
 
--- a/src/cclib/parser/adfparser.py
+++ b/src/cclib/parser/adfparser.py
@@ -146,10 +146,13 @@ class ADF(logfileparser.Logfile):
             info = line.split()
             if info[1] == "NOSYM":
                 self.nosymflag = True
-                point_group = "c1"
+                point_group_full = "c1"
             else:
-                point_group = info[1].replace('(', '').replace(')', '').lower()
-            self.metadata['symmetry_full'] = point_group
+                point_group_full = info[1].replace('(', '').replace(')', '').lower()
+            # TODO
+            point_group_abelian = point_group_full
+            self.metadata['symmetry_full'] = point_group_full
+            self.metadata['symmetry_abelian'] = point_group_abelian
 
         # Use this to read the subspecies of irreducible representations.
         # It will be a list, with each element representing one irrep.
--- a/src/cclib/parser/adfparser.py
+++ b/src/cclib/parser/adfparser.py
@@ -106,7 +106,7 @@ class ADF(logfileparser.Logfile):
         maxelem, norm = list(range(2))  # used to index scf.values
 
     def after_parsing(self):
-        super().after_parsing()
+        super(ADF, self).after_parsing()
 
     def extract(self, inputfile, line):
         """Extract information from the file object inputfile."""
--- a/src/cclib/parser/adfparser.py
+++ b/src/cclib/parser/adfparser.py
@@ -140,6 +140,8 @@ class ADF(logfileparser.Logfile):
             info = line.split()
             if info[1] == "NOSYM":
                 self.nosymflag = True
+            else:
+                point_group = info[1].replace('(', '').replace(')', '').lower()
 
         # Use this to read the subspecies of irreducible representations.
         # It will be a list, with each element representing one irrep.
--- a/src/cclib/parser/adfparser.py
+++ b/src/cclib/parser/adfparser.py
@@ -105,6 +105,9 @@ class ADF(logfileparser.Logfile):
         SCFCNV, SCFCNV2 = list(range(2))  # used to index self.scftargets[]
         maxelem, norm = list(range(2))  # used to index scf.values
 
+    def after_parsing(self):
+        super().after_parsing()
+
     def extract(self, inputfile, line):
         """Extract information from the file object inputfile."""
 
@@ -136,12 +139,17 @@ class ADF(logfileparser.Logfile):
             while line[:5] != "title":
                 line = inputfile.next()
 
+        if "Amsterdam Density Functional" in line:
+            self.metadata['package_version'] = line.split()[5]
+
         if line[1:10] == "Symmetry:":
             info = line.split()
             if info[1] == "NOSYM":
                 self.nosymflag = True
+                point_group = "c1"
             else:
                 point_group = info[1].replace('(', '').replace(')', '').lower()
+            self.metadata['symmetry_full'] = point_group
 
         # Use this to read the subspecies of irreducible representations.
         # It will be a list, with each element representing one irrep.
--- a/src/cclib/parser/gamessparser.py
+++ b/src/cclib/parser/gamessparser.py
@@ -89,7 +89,7 @@ class GAMESS(logfileparser.Logfile):
         self.scftype = "none"  # Type of SCF calculation: BLYP, RHF, ROHF, etc.
 
     def after_parsing(self):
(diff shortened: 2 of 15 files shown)


ATenderholt:
I've merged [cclib/cclib-data#56](https://github.com/cclib/cclib-data/pull/56). How do we convince Travis to retest?

berquist:
I don't think you can do it automatically. If you click on "Details", there's a restart build button on the Travis page.

screen shot 2017-07-07 at 5 43 52 pm

src/cclib/parser/adfparser.py
```diff
@@ -136,10 +139,20 @@ def extract(self, inputfile, line):
             while line[:5] != "title":
                 line = inputfile.next()
 
+        if "Amsterdam Density Functional" in line:
```
langner: Update the PR description since this is not just about symmetry now.

src/cclib/parser/adfparser.py
```diff
@@ -136,10 +139,20 @@ def extract(self, inputfile, line):
...
             info = line.split()
             if info[1] == "NOSYM":
                 self.nosymflag = True
+                point_group_full = "c1"
+            else:
+                point_group_full = info[1].replace('(', '').replace(')', '').lower()
+            # TODO
+            point_group_abelian = point_group_full
+            self.metadata['symmetry_full'] = point_group_full
+            self.metadata['symmetry_abelian'] = point_group_abelian
```
langner: What is the point of having symmetry_abelian if it's never different here?

Also, shouldn't it be a Boolean, since a group is simply Abelian or not? And since it is a property of the group itself, do we really need to parse it?

src/cclib/parser/gamessparser.py
```diff
@@ -25,26 +25,28 @@ class GAMESS(logfileparser.Logfile):
...
-            'CCTC': 'cc-pCVTZ', \
-            'CCQC': 'cc-pCVQZ', \
-            'CC5C': 'cc-pCV5Z', \
-            'CC6C': 'cc-pCV6Z', \
-            'ACCDC': 'aug-cc-pCVDZ', \
-            'ACCTC': 'aug-cc-pCVTZ', \
-            'ACCQC': 'aug-cc-pCVQZ', \
-            'ACC5C': 'aug-cc-pCV5Z', \
-            'ACC6C': 'aug-cc-pCV6Z'}
+    dunningbas = {
```
langner: Is this from some automatic formatter? If so, perhaps in a separate commit?

src/cclib/parser/gamessparser.py
```diff
@@ -86,12 +88,16 @@ def before_parsing(self):
...
+        super(GAMESS, self).after_parsing()
+
     def extract(self, inputfile, line):
         """Extract information from the file object inputfile."""
         
         # extract the version number first
         if line.find("GAMESS VERSION") >= 0:
-            self.metadata["package_version"] = line.split()[4] + line.split()[5] + line.split()[6]
+            tokens = line.split()
+            self.metadata["package_version"] = ''.join(tokens[4:7])
```
langner: Reduce to one line?

test/data/testSP.py
```diff
@@ -241,23 +240,23 @@ def testmoments(self):
             for m in moment32:
                 self.assertEquals(m, 0.0)
 
-    @skipForParser('ADF', 'Does not support metadata yet')
```
langner: Looks like this was overdue? Could you enable these tests in a separate commit?

src/cclib/parser/adfparser.py
berquist: Doing it this way shows both

1. what the maximum point group detected was (`full`), and
2. what the calculation was actually run in (`abelian`).

Take benzene for example. If these are set to `"d6h"` and `False`, it doesn't tell you if the calculation was run in `"d2h"`, `"c2v"`, `"c1"`, etc.

If you choose to have `symmetry_full` be the point group the calculation was run in, and not the full molecular point group, these will be set to `"d2h"` and `True`. You can have benzene or something like it _actually_ be D2h, so some information has been lost.

Pinging @ghutchis to see what he'd like implemented.

src/cclib/parser/adfparser.py
```diff
@@ -105,6 +105,9 @@ def before_parsing(self):
         SCFCNV, SCFCNV2 = list(range(2))  # used to index self.scftargets[]
         maxelem, norm = list(range(2))  # used to index scf.values
 
+    def after_parsing(self):
```
berquist: Done.

src/cclib/parser/adfparser.py
berquist: Done.

src/cclib/parser/gamessparser.py
berquist: Done.

src/cclib/parser/gamessparser.py
berquist: Done.

test/data/testSP.py
berquist: Commit or PR? I'd argue that this is intimately tied to the symmetry detection/testing.

I also think this should be a squash merge at this point.

berquist:
> I'm wondering whether it makes sense to standardize the possible symmetry groups somehow, maybe with an Enum?

I don't know how they work in Python, so I'll take a look.

I'll clean up after my TODOs right now with at least some explanations.

test/data/testSP.py
langner: Yeah, maybe just rebase this PR so that the metadata tests are activated in one commit and the symmetry stuff is added in another.

src/cclib/parser/adfparser.py
langner: Ah, no I understand. So maybe `symmetry_abelian` is not the best choice of name?

berquist: You're right. `symmetry_used`?

langner: Hmm, not very happy with that but it's better, I think.

berquist: What about `symmetry_calculated`? I'd rather try and get this right the first time.

langner: I'm OK with that, too. I don't have any better ideas. Maybe `symmetry_actual`.

berquist: I thought about that, but it could mean either

1. the actual symmetry that was used in the calculation, or
2. the actual symmetry that the system has.

langner: I think `symmetry_used` is was the best. What do you think now?

langner:
@berquist Looks like we'll manage to clean this up for next week and make it in for v1.5.2?

berquist:
Yes, I'll work on it.

langner:
Needs to be reworked due to recent changes. @berquist I think this is worth breaking up into several changes, so we can merge stuff in faster. Besides doing cleanups separately, you can break up the new features into several steps:
1. Add the generic bits, namely the stuff in `logfile_parser.py`, if any, and tests. That way we have our intent documented in the code without actually implementing it.
2. A separate PR for each program, activating the appropriate test. I'm sure we can make better progress by consider one program at a time.

langner:
@berquist were planning on following up here with new PRs?

berquist:
I am reworking this. Do you still want 1 PR/program?

Also going to make these changes:
- `symmetry_full -> symmetry_detected`
- `symmetry_abelian -> symmetry_used`

PR #393, opened 2017-07-05
Parses thermo from ORCA

jevandezande:
Parses temperature, pressure, ZPE, enthalpy, entropy, and free energy.

I've added all my energies in eV. The specified the thermodynamic units of `hartree/particle` (which perhaps should be just `hartree`) is inconsistent with scf and mp energies, which are in `eV`. Is this intentional? It would perhaps be nicer if most energies used the same units (though vibfreqs should obviously be `cm^-1`).

I've added simple unit tests. Should I add more complex ones, and if so, what should they include?

Other recommendations for my code? I hope to add more ORCA related stuff in the future.

Note: ORCA 3 had a bug in the IR intensities for SCF and MP2. This has been fixed in ORCA 4, but it explains why the results were not good (as noted in testvib).

### DIFF
--- a/src/cclib/parser/orcaparser.py
+++ b/src/cclib/parser/orcaparser.py
@@ -569,35 +569,35 @@ class ORCA(logfileparser.Logfile):
             while line[:17] != 'Electronic energy':
                 line = next(inputfile)
             self.zpe = next(inputfile).split()[4]
-            thermal_vibrational_correction = next(inputfile).split()[4]
-            thermal_rotional_correction = next(inputfile).split()[4]
-            thermal_translational_correction = next(inputfile).split()[4]
+            thermal_vibrational_correction = float(next(inputfile).split()[4])
+            thermal_rotional_correction = float(next(inputfile).split()[4])
+            thermal_translational_correction = float(next(inputfile).split()[4])
             next(inputfile)
-            total_thermal_energy = next(inputfile).split()[3]
+            total_thermal_energy = float(next(inputfile).split()[3])
 
             # Enthalpy
             line = next(inputfile)
             while line[:17] != 'Total free energy':
                 line = next(inputfile)
-            thermal_enthalpy_correction = next(inputfile).split()[4]
+            thermal_enthalpy_correction = float(next(inputfile).split()[4])
             next(inputfile)
-            self.enthalpy = next(inputfile).split()[3]
+            self.enthalpy = float(next(inputfile).split()[3])
 
             # Entropy
             line = next(inputfile)
             while line[:18] != 'Electronic entropy':
                 line = next(inputfile)
-            electronic_entropy = line.split()[3]
-            vibrational_entropy = next(inputfile).split()[3]
-            rotational_entropy = next(inputfile).split()[3]
-            translational_entropy = next(inputfile).split()[3]
+            electronic_entropy = float(line.split()[3])
+            vibrational_entropy = float(next(inputfile).split()[3])
+            rotational_entropy = float(next(inputfile).split()[3])
+            translational_entropy = float(next(inputfile).split()[3])
             next(inputfile)
-            self.entropy = next(inputfile).split()[4]
+            self.entropy = float(next(inputfile).split()[4])
 
             line = next(inputfile)
             while line[:25] != 'Final Gibbs free enthalpy':
                 line = next(inputfile)
-            self.freeenergy = line.split()[5]
+            self.freeenergy = float(line.split()[5])
 
         # Read TDDFT information
         if any(x in line for x in ("TD-DFT/TDA EXCITED", "TD-DFT EXCITED")):
--- a/src/cclib/parser/orcaparser.py
+++ b/src/cclib/parser/orcaparser.py
@@ -551,6 +551,55 @@ class ORCA(logfileparser.Logfile):
                 self.gbasis.append(gbasis_tmp[bas_atname])
             del self.tmp_atnames
 
+        """
+        --------------------------
+        THERMOCHEMISTRY AT 298.15K
+        --------------------------
+        """
+        if 'THERMOCHEMISTRY AT' == line[:18]:
+            conv = lambda x: utils.convertor(float(x), "hartree", "eV")
+
+            next(inputfile)
+            next(inputfile)
+            self.temperature = float(next(inputfile).split()[2])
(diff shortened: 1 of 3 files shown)


berquist:
First of all, thanks for the patch. It's always good to have new contributors, and for different packages.

> `hartree/particle` (which perhaps should be just `hartree`)

It's written this way because those are the units Gaussian prints in, and that was probably the first parser to have thermochemistry attributes added to. I think they say "hartree/particle" and not "hartree" just to differentiate from "kcal/mol" and "kJ/mol", but they should be the same.

I think we're all in agreement that the units are a bit of a mess right now. Thermodynamic energies should probably be in wavenumbers or kcal/mol, and everything else should be in hartrees. My only suggestion here would be to not use eV, but use hartree/particle, so no conversion should be necessary anymore. We'll probably have a discussion about changing the units for cclib v2.0.

> I've added simple unit tests. Should I add more complex ones, and if so, what should they include?

What's there is a good start, but can you turn them into comparisons against reference data?

> Note: ORCA 3 had a bug in the IR intensities for SCF and MP2. This has been fixed in ORCA 4, but it explains why the results were not good (as noted in testvib).

This is a good reminder for me to update the ORCA unit test data files. Maybe I'll do that now...

src/cclib/parser/orcaparser.py
```diff
@@ -551,6 +551,55 @@ def extract(self, inputfile, line):
...
+        --------------------------
+        THERMOCHEMISTRY AT 298.15K
+        --------------------------
+        """
+        if 'THERMOCHEMISTRY AT' == line[:18]:
+            conv = lambda x: utils.convertor(float(x), "hartree", "eV")
+
+            next(inputfile)
+            next(inputfile)
+            self.temperature = float(next(inputfile).split()[2])
```
langner: We don't do this kind of alignment. Could you use one space at each side of all binary operators?

src/cclib/parser/orcaparser.py
```diff
@@ -551,6 +551,55 @@ def extract(self, inputfile, line):
                 self.gbasis.append(gbasis_tmp[bas_atname])
             del self.tmp_atnames
 
+        """
+        --------------------------
+        THERMOCHEMISTRY AT 298.15K
+        --------------------------
+        """
+        if 'THERMOCHEMISTRY AT' == line[:18]:
+            conv = lambda x: utils.convertor(float(x), "hartree", "eV")
```
langner: Perhaps call this hartree2ev or similar for clarity?

src/cclib/parser/orcaparser.py
```diff
@@ -551,6 +551,55 @@ def extract(self, inputfile, line):
                 self.gbasis.append(gbasis_tmp[bas_atname])
             del self.tmp_atnames
 
+        """
+        --------------------------
```
langner: A longer fragment of example output might be useful here.

test/data/testvib.py
```diff
@@ -81,6 +81,27 @@ class OrcaIRTest(GenericIRTest):
     # but in the meanwhile we will expect to parse this value.
     max_IR_intensity = 215
 
+    def testtemperature(self):
+        """Is the temperature 298.15 K?"""
+        print(self.data)
```
langner: Leftover from debugging?

test/data/testvib.py
```diff
@@ -81,6 +81,27 @@ class OrcaIRTest(GenericIRTest):
...
+        """Is the temperature 298.15 K?"""
+        print(self.data)
+        self.assertAlmostEqual(298.15, self.data.temperature)
+
+    def testpressure(self):
+        """Is the pressure 1 atm?"""
+        self.assertAlmostEqual(1, self.data.pressure)
+
+    def testenthalpy(self):
+         """Is the enthalpy defined"""
```
langner: nit: missing question mark? (also in other test methods)

test/test_data.py
```diff
@@ -163,6 +163,7 @@ def testall(self):
             parser = self.parsers[td['parser']]
             test = getattr(module, td['class'])
 
+            description = ''
```
langner: Is this needed?

jevandezande: If self.silent is True, then lines 196 and 197 fail as description is undefined.

src/cclib/parser/orcaparser.py
jevandezande: Orca has a lot of filler in the thermochemistry output and breaks it up into a bunch of sections (some of this filler changed upon release of version 4; doesn't affect my parsing). It would take >100 lines, so I don't think it is worthwhile to include a fragment. If you still think it is worthwhile, I can go ahead and add a fragment.

jevandezande:
I've made some test cases for Orca, but ran into a problem where the Raman and IR test cases have different geometries and thus slightly different thermochemistry. If I write an IR test that checks the freeenergy, it runs for both outputs, and thus any hardcoded values are off by up to a millihartree. Should I check the free energy of each output separately (if so, what would be the best way to do this), change the geometry of the Raman test case to match the IR test case, or just allow large error bars on the thermochemistry (need large ones if regression tests are included)?

langner:
> I've made some test cases for Orca, but ran into a problem where the Raman and IR test cases have different geometries and thus slightly different thermochemistry. If I write an IR test that checks the freeenergy, it runs for both outputs, and thus any hardcoded values are off by up to a millihartree. Should I check the free energy of each output separately (if so, what would be the best way to do this), change the geometry of the Raman test case to match the IR test case, or just allow large error bars on the thermochemistry?
> 
> Furthermore, I just found out about the regression tests the hard way; they break even worse on hardcoded values.

I think you went the right course, but a bit too far with the loosening.

langner:
> I've made some test cases for Orca, but ran into a problem where the Raman and IR test cases have different geometries and thus slightly different thermochemistry. If I write an IR test that checks the freeenergy, it runs for both outputs, and thus any hardcoded values are off by up to a millihartree. Should I check the free energy of each output separately (if so, what would be the best way to do this), change the geometry of the Raman test case to match the IR test case, or just allow large error bars on the thermochemistry?
> 
> Furthermore, I just found out about the regression tests the hard way; they break even worse on hardcoded values.
> I think you went the right course, but a bit too far with the loosening.

Ahh, now I see why you set places to -1, it's because of the failing regressions. That's fine for now - I'll do something about that later.

Thanks for the contribution!

PR #394, opened 2017-07-07
Orca: update unit test files to use v4.0

berquist:
This also moves the Travis testing commands to a separate shell script which makes testing locally much easier.


ATenderholt:
Should we retire the older Orca 2.9 to regressions? I vaguely recall we had decided to not keep test files for *every* version, but don't remember if we decided it was official policy.

berquist:
Yes, I think we should. I also don't recall an official policy. Maybe a good one would be to keep one prior version, and all others go to regressions?

data/.gitignore
```diff
@@ -1 +1,21 @@
 regression
+
+# ORCA
```
langner: I don't see why we need to ignore so many files here. If you run calcs in these directories locally, perhaps ignore them only locally, see https://git-scm.com/docs/gitignore for details.

.travis.yml
```diff
@@ -11,11 +11,4 @@ install:
...
 script:
-  - cd test
-  - python test_utils.py
-  - python test_method.py
-  - python test_parser.py
-  - python test_io.py
-  - python test_data.py --status --terse
-  - cd ../data && bash regression_download.sh
-  - cd ../test && python run_regressions.py --status --traceback
+  - sh run_tests.sh
```
langner: Since this is specific to travis in some way, perhaps call it travis_tests or similar?

langner:
> Yes, I think we should. I also don't recall an official policy. Maybe a good one would be to keep one prior version, and all others go to regressions?

I think this is a reasonable balance (leaving one previous vresion).

.travis.yml
berquist: I should have been more clear. Since I always run the same Travis commands myself when I'm developing, I just moved them to a script. This way, anyone can do `$ sh ./run_tests.sh`. You do run the tests before you commit, don't you? 😉 I'll add a comment in the shell script and update the PR comment, because I think this is more general than Travis.

data/.gitignore
berquist: Done.

berquist:
Let me know if the testing script should be in another PR.

.travis.yml
langner: Well, I usually run test_all.py which includes more than Travis does. But I don't always run it with multiple Python version (in fact, almost never), which I see as the main benefit of Travis when I do my own commits.

langner:
> Let me know if the testing script should be in another PR.

I think this can stay here, but just rebase to a logical set of commits.

data/.gitignore
langner: I meant that maybe we don't need all these ignores at all. I imagine the types of temporary files can change over time, and these ignores are specific to your local setup where you run calcs in this directory. Could you keep the ignores local for these reasons?

berquist:
I didn't know about testall. I am thinking it should go in another PR with a switch to the new Travis testing infrastructure, if possible: https://docs.travis-ci.com/user/trusty-ci-environment/

langner:
> I didn't know about testall. I am thinking it should go in another PR with a switch to the new Travis testing infrastructure, if possible: https://docs.travis-ci.com/user/trusty-ci-environment/

Sounds good.

berquist:
I've changed my mind, I won't combine the test script with the Travis infrastructure update.

langner:
> I've changed my mind, I won't combine the test script with the Travis infrastructure update.

I'd prefer separate PRs, but whatever.

data/.gitignore
langner: So could remove all these temporary files from .gitignore?

berquist:
No, I'll do it properly, Git was just making me cranky yesterday...

gaursagar:
Can please these two files be used instead of the ones in this PR? The coordinates used here are being used in testing the `wfx writer`

[dvb_sp.inp](https://github.com/cclib/cclib/files/1252194/dvb_sp.inp.txt)
[dvb_sp.out](https://github.com/cclib/cclib/files/1252193/dvb_sp.out.txt)

berquist:
These are the newly-standardized coordinates, right? If so I'll make sure all of them are correct.

gaursagar:
Yes these coordinates are from the standardised set.

langner:
Do we want this to go into v1.5.2?

berquist:
Yes, I'm almost done with it. However, the scan calculation (not the test, the actual calculation) with the updated coordinates is failing.

langner:
@berquist Is this still an issue? The file I see in the PR has "normally terminated" at the end. Is that for old coordinates?

Also, could you please use "Adam's hack", namely have one commit that copies the 3.0 inputs and outputs, and then update them in subsequent commits? That way we'll see what the changes between versions are.

berquist:
That's for old coordinates; I didn't add the new one to the PR. Not sure what to do about it.

I don't understand what you mean by "one commit that copies the old inputs and outputs." You mean copy 3.0 to 4.0, then actually replace fake 4.0 with real 4.0?

langner:
> That's for old coordinates; I didn't add the new one to the PR. Not sure what to do about it.

I think it's OK to just not submit it and add an exception in the unit test until we figure out how to get it to work. We do still have the 3.0 output that is tested.

> I don't understand what you mean by "one commit that copies the old inputs and outputs." You mean copy 3.0 to 4.0, then actually replace fake 4.0 with real 4.0?

Yes, exactly. Commit 1: copy inputs/outputs from 3.0 to 4. Commit 2: replace old files with updated ones. Now looking at commit 2 shows you the diff.

langner:
Looks much more readable! Did you actually add the 2.9 logfiles to regressions? I don't see a PR.

berquist:
See https://github.com/cclib/cclib-data/pull/57.

langner:
I think this is ready to be merged, but not entirely sure. @berquist could you comment?

berquist:
Looks good to me.

langner:
Seems good, but conflicts need to be resolved. Pushing back for now.

langner:
Instead of merging, you rebase your changes on top of of `master. From you current branch state, you would need to rewind a little, so:
```
git reset --hard HEAD~1
git rebase cclib/master
```
(replace cclib with whatever upstream is called locally in you terminal)

PR #395, opened 2017-07-13
QChem: fix for trying to parse more MOs than basis functions

berquist:
The regression PR is https://github.com/cclib/cclib-data/pull/58.

This needs to be merged before the regression PR.

### DIFF
--- a/src/cclib/parser/qchemparser.py
+++ b/src/cclib/parser/qchemparser.py
@@ -66,7 +66,8 @@ class QChem(logfileparser.Logfile):
         # aoname.
         self.re_atomindex = re.compile('(\d+)_')
 
-        # A maximum of 6 columns per block when printing matrices.
+        # A maximum of 6 columns per block when printing matrices. The
+        # Fock matrix is 4.
         self.ncolsblock = 6
 
         # By default, when asked to print orbitals via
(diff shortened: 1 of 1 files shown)


src/cclib/parser/qchemparser.py
```diff
@@ -481,6 +483,12 @@ def extract(self, inputfile, line):
             if 'basis functions' in line:
                 if not hasattr(self, 'nbasis'):
                     self.set_attribute('nbasis', int(line.split()[-3]))
+                    # We can't display more MOs than there are basis
```
langner: By we do you mean QChem? Could you be more explicit - does QChem print more MOs than there are basis functions and we parse only a subset?

berquist: The problem is the parser needs to keep track of how many MOs are going to be displayed. By default, this is all the occupied MOs and the first 5 virtual MOs. However, it isn't known how many MOs this could be until the number of electroons and the number of basis functions are encountered.

In the process of figuring out the display number, it may become larger than the number of MOs that are physically present. This would happen if there are fewer than 5 virtual MOs.

I've tried changing the comment to clarify this.

berquist:
I'm seeing that other PRs are failing now that this hasn't been merged, but the regression PR has.

langner:
> I'm seeing that other PRs are failing now that this hasn't been merged, but the regression PR has.

Sorry about doing it in the wrong order.

berquist:
Ah, it's ok, I'll restart the CI tests.

PR #399, opened 2017-07-26
Cleans up methods.

jevandezande:
It looks like sections of some methods had been incorrectly copy-pasted. This fixes the string formatting for them and removes some trailing whitespace that has crept in over time


langner:
Thanks!

PR #400, opened 2017-07-27
wfx writer

gaursagar:
This PR adds the capability of writing `.wfx` files using `ccwrite`. 
For more information about the format please refer to the [AIMALL](http://aim.tkgristmill.com/wfxformat.html) website.

### DIFF
--- a/src/cclib/io/wfxwriter.py
+++ b/src/cclib/io/wfxwriter.py
@@ -1,0 +1,153 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2017, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
+
+"""A writer for wfx format files."""
+
+import os.path
+
+from . import filewriter
+
+class WFXWriter(filewriter.Writer):
+    """A writer for wfx files."""
+
+    required_attrs = ['atomcoords', 'atomnos', 'gbasis', 'charge', 'homos']
+    keyword = 'GTO'
+
+    @staticmethod
+    def _section(section_name, section_data):
+        """Add opening/closing section_name tags to data."""
+        opening_tag = ['<' + section_name + '>']
+        closing_tag = ['</' + section_name + '>']
+        
+        if isinstance(section_data, list):
+            return opening_tag + section_data + closing_tag
+        elif isinstance(section_data, str):
+            return opening_tag + section_data.split() + closing_tag
+        elif isinstance(section_data, int) or isinstance(section_data, float):
+            return opening_tag + [str(section_data)] + closing_tag
+        else:
+            return opening_tag + closing_tag
+
+    def _title(self):
+        """Return filename without extension to be used as title."""
+        if self.jobfilename is not None:
+            title = os.path.basename(os.path.splitext(self.jobfilename)[0])
+            return title
+
+    def _keywords(self):
+        """Return one of GTO, GIAO, CSGT keyword."""
+        # AIMALL currently supports on GTO keyword.
+        return self.keyword
+
+    def _no_of_nuclei(self):
+        """Number of nuclei in the molecule."""
+        no_nuclei = len(self.ccdata.atomnos)
+        return no_nuclei
+
+    def _no_of_prims(self):
+        """Number of Primitives."""
+        nprims = 0
+        for atom in self.ccdata.gbasis:
+            for prims in atom:
+                nprims += len(prims)
+        return nprims
+
+    def _no_of_mos(self):
+        """Number of occupied MOs."""
+        return int(max(self.ccdata.homos))
+
+    def _no_of_perturbations(self):
+        """Number of Perturbation."""
+        # This is usually zero.  For GIAO it should be 3 (corresponding to Lx, Ly and Lz) and for
+        # CSGT it should be 6 (corresponding to Lx, Ly, Lz, Px, Py and Pz).
+        nperturb = 0
+        if self.keyword == 'GIAO':
+            nperturb = 3
+        elif self.keyword == 'CSGT':
+            nperturb = 6
+        return nperturb
+
+    def _nuclear_names(self):
+        """Names of nuclei present in the molecule."""
+        # O1 H2 H3 etc.
+        return [self.pt.element[Z]+str(i) for i, Z in enumerate(self.ccdata.atomnos)]
+
+    def _atomic_nos(self):
+        """Atomic numbers of elements."""
+        return [str(Z) for Z in self.ccdata.atomnos]
+
+    def _nuclear_charges(self):
+        """Nuclear charges."""
+        return ['{:.14e}'.format(Z) for Z in self.ccdata.atomnos]
+
+    def _nuclear_coords(self):
+        """Nuclear charges."""
+        coord_template = '%.14E %.14E %.14E'        
+        nuc_coords = [coord_template % tuple(coord) for coord in self.ccdata.atomcoords[-1]]
+        return nuc_coords
+
+    def _net_charge(self):
+        """Net charge on molecule."""
+        return '%.14E' % self.ccdata.charge
+
+    def _no_electrons(self):
+        """Number of electrons in molecule."""
+        homos = self.ccdata.homos
+        charge = self.ccdata.charge
+        print (sum(homos), charge)
+        if len(homos) > 1:
+            return int(sum(homos))
+        elif charge > 0:
+            return int(homos[0]*2 - (charge % 2))
+
+    def generate_repr(self):
+        """Generate the wfx representation of the logfile data."""
+
+        sections = [
+            (self._title, "Title"),
+            (self._keywords, "Keywords"),
+            (self._no_of_prims, "Number of Primitives"),
+            (self._no_of_mos, "Number of Occupied Molecular Orbitals"),
+            (self._no_of_perturbations, "Number of Perturbations"),
+            (self._nuclear_names, "Nuclear Names"),
+            (self._atomic_nos, "Atomic Numbers"),
+            (self._nuclear_charges, "Nuclear Charges"),
+            (self._nuclear_coords, "Nuclear Cartesian Coordinates"),
+            (self._net_charge, "Net Charge"),
+            (self._no_electrons, "Number of Electrons"),
+            # (self._no_alpha_electrons, "Number of Alpha Electrons"),
+            # (self._no_beta_electrons, "Number of Beta Electrons"),
+            # (self._spin_mult, "Electronic Spin Multiplicity"),
+            # (self._model, "Model"),
+            # (self._prim_centers, "Primitive Centers"),
+            # (self._prim_types, "Primitive Types"),
+            # (self._prim_exps, "Primitive Exponents"),
+            # (self._mo_occup_nos, "Molecular Orbital Occupation Numbers"),
+            # (self._mo_energies, "Molecular Orbital Energies"),
+            # (self._mo_spin_types, "Molecular Orbital Spin Types"),
+            # (self._mo_prim_coeffs, "Molecular Orbital Primitive Coefficients"),
+            # (self._energy, "Energy"),
+            # (self._nuc_energy_gradients, "Nuclear Cartesian Energy Gradients"),
+            # (self._nuc_virial, "Nuclear Virial of Energy-Gradient-Based Forces on Nuclei, W"),
+            # (self._virial_ratio, "Full Virial Ratio"),
+        ]
+
+        wfx_lines = []
+
+        for section_module, section_name in sections:
+            try:
+                section_data = section_module()
+            except:
+                section_data = None
+            wfx_lines.extend(self._section(section_name, section_data))
+
+        wfx_lines.append('')
+        return '\n'.join(wfx_lines)
+
+
+if __name__ == "__main__":
+    pass
--- a/src/cclib/io/wfxwriter.py
+++ b/src/cclib/io/wfxwriter.py
@@ -13,78 +13,119 @@ from cclib.parser import utils
 
 from . import filewriter
 
+
+# Number of orbitals of type key.
+# There are 3 p type, 6 d type orbitals etc.
+ORBITAL_COUNT = {'S':1, 'P':3, 'D':6, 'F':10, 'G':15, 'H':21}
+
+# Index of first orbital of type key in a list of orbitals.
+# The first s orbital has index 1, first p orbital has index 2, and first d
+# has index 5.
+ORBITAL_INDICES = {'S': 1}
+ORBITAL_NAMES = 'SPDFGH'
+for idx, name in enumerate(ORBITAL_NAMES[1:], start=1):
+    prev_orbital_name = ORBITAL_NAMES[idx - 1]
+    prev_orbital_count = ORBITAL_COUNT[prev_orbital_name]
+    prev_orbital_index = ORBITAL_INDICES[prev_orbital_name]
+    ORBITAL_INDICES[name] = prev_orbital_count + prev_orbital_index
+
+PI_CUBE_INV = (2.0 / numpy.pi) ** 3
+
+# Float formatting template.
+WFX_FIELD_FMT = '%22.11E'
+
+# Precomputed values for l+m+n to be used in MO normalization.
+_L = dict(
+    [(prim_type, 0) for prim_type in range(1, 2)] +   # s
+    [(prim_type, 1) for prim_type in range(2, 5)] +   # p
+    [(prim_type, 2) for prim_type in range(5, 11)] +  # d
+    [(prim_type, 3) for prim_type in range(11, 21)] + # f
+    [(prim_type, 4) for prim_type in range(21, 36)]   # g
+)
+
+# Precomputed values for ((2l-1)!! * (2m-1)!! * (2n-1)!!).
+_M = dict(
+    [(L, 1) for L in range(1, 5)] +
+    [(L, 9) for L in range(5, 8)] +
+    [(L, 1) for L in range(8, 11)] +
+    [(L, 225) for L in range(11, 14)] +
+    [(L, 9) for L in range(14, 20)] +
+    [(L, 1) for L in range(20, 21)] +
+    [(L, 11025) for L in range(21, 24)] +
+    [(L, 225) for L in range(24, 30)] +
+    [(L, 81) for L in range(30, 33)] +
+    [(L, 9) for L in range(33, 36)]
+)
+
+
+def _section(section_name, section_data):
+    """Add opening/closing section_name tags to data."""
+    opening_tag = ['<' + section_name + '>']
+    closing_tag = ['</' + section_name + '>']
+
+    section = None
+    if isinstance(section_data, list):
+        section = opening_tag + section_data + closing_tag
+    elif isinstance(section_data, str):
+        section = opening_tag + (' ' + section_data).split('\n') + closing_tag
+    elif isinstance(section_data, int) or isinstance(section_data, float):
+        section = opening_tag + [' ' + str(section_data)] + closing_tag
+    return section
+
+
+def _list_format(data, per_line, style=WFX_FIELD_FMT):
+    """Format lists for pretty print."""
+    template = style * per_line
+    leftover = len(data) % per_line
+    # Template for last line.
+    last_template = style * leftover
+
+    pretty_list = [template%tuple(data[i:i+per_line])
+                    for i in range(0, len(data) - leftover, per_line)]
+    if leftover:
+        return pretty_list + [last_template%tuple(data[-1*leftover:])]
+    return  pretty_list
+
+
 class WFXWriter(filewriter.Writer):
     """A writer for wfx files."""
 
-    required_attrs = ('atomcoords', 'atomnos', 'gbasis', 'charge', 'homos',
-                      'mult')
-    keyword = 'GTO'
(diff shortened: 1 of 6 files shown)


src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+
+        prim_types = []
+        for atom in self.ccdata.gbasis:
+            for prims in atom:
+                prim_orb = []
+                for i in range(self.num_orb[prims[0]]):
+                    prim_orb += [(self.idx_orb[prims[0]]  + i)]\
+                                * len(prims[1])
+                prim_types += prim_orb
+        return prim_types
```
ATenderholt: Shouldn't ```_prim_types()``` be based on ```aonames``` instead of ```gbasis``` since we have no guarantee that each QM program uses the same order of primitives?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+
+    def _prim_types_section(self):
+        prim_types = self._prim_types()
+        # GAMESS specific reordering.
+        if self.ccdata.metadata['package'] == 'GAMESS':
+            prim_types = self._rearrange_modata(prim_types)
```
ATenderholt: Do we know of any programs besides GAMESS that need special ordering?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+        alpha = self._no_alpha_electrons()
+        beta = self._no_beta_electrons()
+        if len(self.ccdata.homos) == 1:
+            occup += ['{:22.11E}'.format(2)] * int(electrons / 2) +\
+                        ['{:22.11E}'.format(1)] * (electrons % 2)
+        else:
+            occup += ['{:22.11E}'.format(1)] *  + alpha +\
+                        ['{:22.11E}'.format(1)] * beta
+        return occup
+
```
ATenderholt: Can excited-state calculations be handled?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+        beta = self._no_beta_electrons()
+        if len(self.ccdata.homos) == 1:
+            spin_types += ['Alpha and Beta'] * int(electrons / 2) +\
+                            ['Alpha'] * (electrons % 2)
+        else:
+            spin_types += ['Alpha'] * alpha +\
+                            ['Beta'] * beta
+        return spin_types
+
+    pi3 = (2.0 / numpy.pi) ** 3
```
ATenderholt: ```pi3``` is only used in ```normalize_()```, so I think it should be moved there for clarity.

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+        norm_four = self.pi3 * 2**(4*L) * alpha**(3+2*L) / M
+        norm = numpy.power(norm_four, 1/4.0)
+        return norm
+
+    def _rearrange_mocoeffs(self, mocoeffs):
+        """Rearrange cartesian F functions in mocoeffs.
+        Expected order:
+        xxx, yyy, zzz, xyy, xxy, xxz, xzz, yzz, yyz, xyz
+        cclib's order:
+        XXX, YYY, ZZZ, XXY, XXZ, YYX, YYZ, ZZX, ZZY, XYZ
```
ATenderholt: If I remember correctly, cclib just parses the order in the log file so this doesn't seem like a robust solution to me.

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+                    prim_orb += [norb + i]
+                    alpha += [prim[0] for prim in prims[1]]
+                    prim_coeff += [prim[1] for prim in prims[1]]
+
+        # GAMESS specific reordering.
+        if self.ccdata.metadata['package'] == 'GAMESS':
+            prim_type = self._rearrange_modata(self._prim_types())
+            alpha = self._rearrange_modata(alpha)
+            prim_coeff = self._rearrange_modata(prim_coeff)
+
```
ATenderholt: Again, what about other packages?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+        21 GXXXX, 22 GYYYY, 23 GZZZZ, 24 GXXXY, 25 GXXXZ, 26 GXYYY,
+        27 GYYYZ, 28 GXZZZ,
+        29 GYZZZ, 30 GXXYY, 31 GXXZZ, 32 GYYZZ, 33 GXXYZ, 34 GXYYZ,
+        35 GXYZZ,
+        36 HZZZZZ, 37 HYZZZZ, 38 HYYZZZ, 39 HYYYZZ,
+        40 HYYYYZ, 41 HYYYYY, 42 HXZZZZ, 43 HXYZZZ, 44 HXYYZZ,
+        45 HXYYYZ, 46 HXYYYY, 47 HXXZZZ, 48 HXXYZZ, 49 HXXYYZ,
+        50 HXXYYY, 51 HXXXZZ, 52 HXXXYZ, 53 HXXXYY, 54 HXXXXZ, 55 HXXXXY,
+        56 HXXXXX
+        """
```
ATenderholt: What about spherical basis functions?

src/cclib/io/wfxwriter.py
gaursagar: `normalize` is called very frequently, once for each `mocoeff` .Hence, I had put it as a class attribute to avoid recalculation. How about making it a global variable?

src/cclib/io/wfxwriter.py
gaursagar: This particular reordering is for `GAMESS`. I got it by doing a bit of reverse engineering on `wfn` files (which are the old style `wfx` files) generated using `GAMESS`.

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+        return '%22.11E' % self.ccdata.charge
+
+    def _no_electrons(self):
+        """Number of electrons in molecule."""
+        homos = self.ccdata.homos
+        charge = self.ccdata.charge if hasattr(self.ccdata, 'charge') else 0.0
+        if len(homos) > 1:
+            return int(sum(homos)) + 2
+        else:
+            return int((homos[0] + 1)*2 - (charge % 2))
```
gaursagar: According to the official .wfx specification given on the AIMALL [website](http://aim.tkgristmill.com/wfxformat.html)

> When an ECP is used, the written nuclear charge is less than the real nuclear charge by the number of core electrons modeled by the ECP.  Thus, for Cl and a valency-only ECP, the atomic number is 17 but the nuclear charge is 7.

I would need to consider `ccdata.coreelectrons` here.

src/cclib/io/wfxwriter.py
gaursagar: The `.wfx` specification on AIMALL [website](http://aim.tkgristmill.com/wfxformat.html) doesn't specify how to handle spherical basis. I would get back on this one after a bit of reverse engineering how AIMALL writes `.wfx` files using Gaussian's checkpoint files.

src/cclib/io/wfxwriter.py
gaursagar: I would agree with you on this one. Right now there might be a need to do parser specific reordering.

src/cclib/io/wfxwriter.py
gaursagar: So far I know of `GAMESS` and by extension `Firefly`.

src/cclib/io/wfxwriter.py
gaursagar: One thing for sure that if `aonames` are used, the code would have to consider the basis set being used.
For instance, `aonames` for water molecule in STO-3G basis are:
`['O1_S', 'O1_S', 'O1_X', 'O1_Y', 'O1_Z', 'H2_S', 'H3_S']`

for cartesian basis this section is to be written as:
```xml
<Primitive Types>
1 1 1 1 1 1 2 3 4 2 
3 4 2 3 4 1 1 1 1 1 
1 
</Primitive Types>
```

If `gbasis` is used, these is easier to generate.

src/cclib/io/wfxwriter.py
gaursagar: _*Not a thorough investigation_, 
for `MoOCl4` molecule in spherical basis :
(values from ccget on `Gaussian`)
```
'Mo1_1S', 'Mo1_2S',
'Mo1_3PX', 'Mo1_3PY', 'Mo1_3PZ', 
'Mo1_4PX', 'Mo1_4PY', 'Mo1_4PZ',
'Mo1_5D 0', 'Mo1_5D+1', 'Mo1_5D-1', 'Mo1_5D+2', 'Mo1_5D-2'
```
, the primitives are numbered as:
(values from `.wfx` written for `Gaussian` by AIMALL)
```
1 1
2 3 4 
2 3 4 
5 6 7 8 9 10
```

and in cartesian basis:
(values from ccget on `GAMESS`)
```
'Mo1_S', 'Mo1_S', 'Mo1_S',
'Mo1_X', 'Mo1_Y', 'Mo1_Z', 
'Mo1_X', 'Mo1_Y', 'Mo1_Z', 
'Mo1_X', 'Mo1_Y', 'Mo1_Z', 
'Mo1_XX', 'Mo1_YY', 'Mo1_ZZ', 'Mo1_XY', 'Mo1_XZ', 'Mo1_YZ', 
'Mo1_XX', 'Mo1_YY', 'Mo1_ZZ', 'Mo1_XY', 'Mo1_XZ', 'Mo1_YZ'
```
the primitve types look something like:
(values from `.wfn` written for `GAMESS`)
```
1 1 1 1 1 1 1 1 2 2 
2 3 3 3 4 4 4 2 2 3 
3 4 4 2 3 4 5 5 5 6 
6 6 7 7 7 8 8 8 9 9 
9 10 10 10 5 6 7 8 9 10 
```
The repetitions are due to basis set involved, which are handled using `gbasis` in the writer.

ATenderholt: Seems like there is some conversion from spherical to cartesian orbitals. If you create a WFX file for ```MoOCl4``` using both your writer and AIMALL, are there any other differences?

src/cclib/io/wfxwriter.py
ATenderholt: Oh, I was misunderstanding what the primitive types are supposed to be. I think your approach makes more sense.

src/cclib/io/wfxwriter.py
ATenderholt: Ah, good point. I think it can stay here then.

langner: Perhaps you could think of a better name?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+from . import filewriter
+
+class WFXWriter(filewriter.Writer):
+    """A writer for wfx files."""
+
+    required_attrs = ('atomcoords', 'atomnos', 'gbasis', 'charge', 'homos',
+                      'mult')
+    keyword = 'GTO'
+    # Number of orbitals of type key.
+    num_orb = {'s':1, 'p':3, 'd':6, 'f':10, 'g':15, 'h':21,
```
langner: Do you need to have both lower and upper cases here?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+               'S':1, 'P':3, 'D':6, 'F':10, 'G':15, 'H':21}
+    # Index of first orbital of type key in a list of orbitals.
+    idx_orb = {'s':1, 'p':2, 'd':5, 'f':11, 'g':21, 'h':36,
+               'S':1, 'P':2, 'D':5, 'F':11, 'G':21, 'H':36}
+
+    @staticmethod
```
langner: Any reason not to make this a module level function?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,475 @@
...
+
+        if isinstance(section_data, list):
+            return opening_tag + section_data + closing_tag
+        elif isinstance(section_data, str):
+            return opening_tag + (' ' + section_data).split('\n') + closing_tag
+        elif isinstance(section_data, int) or isinstance(section_data, float):
+            return opening_tag + [' ' + str(section_data)] + closing_tag
+
+
+    @staticmethod
```
langner: Any reason not to make this a module level function?

src/cclib/io/wfxwriter.py
gaursagar: Its a small hack to avoid explicitly making sure letters are all lowercase.

src/cclib/io/wfxwriter.py
gaursagar: No there isn't. I would move the methods.

src/cclib/io/wfxwriter.py
gaursagar: I could test that for `GAMESS` and there isn't any difference. And `.wfx` files written for other programs (which use cartesian basis) are same as those of `GAMESS` .

As of now writer isn't expected to work for programs which use spherical basis. However, the files which do get written have ordering differences in primitive types, and therefore different values of normalized MO coefficients. 
Though they all pass the strict consistency checks by AIMALL program, shouldn't be considered correct.

src/cclib/io/wfxwriter.py
langner: Can't you just covert before looking up?

test/io/testwfxwriter.py
```diff
@@ -0,0 +1,168 @@
...
+            'NWChem': "data/NWChem/basicNWChem6.5/dvb_sp_hf.out",
+            'Psi': "data/Psi/basicPsi4.0/dvb_sp_rhf.out",
+            'GAMESS-UK': "data/GAMESS-UK/basicGAMESS-UK8.0/dvb_sp_hf.out",
+            'Firefly': "data/GAMESS/basicFirefly8.0/dvb_sp.out",
+            'Molpro': "data/Molpro/basicMolpro2012/dvb_sphf.out"
+        }
+        fpath_ref = os.path.join(__datadir__, ref_file)
+        data_ref = cclib.io.ccopen(fpath_ref).parse()
+        wfx_ref = cclib.io.wfxwriter.WFXWriter(data_ref)
+        norm_mat_ref = wfx_ref._norm_mat()[0]
```
TDI-tenderholt: Why test ```_norm_mat()```, which isn't directly used by the writer, instead of ```_mo_prim_coeffs()``` or maybe ```_normalized_mocoeffs()```?

TDI-tenderholt:
Is there a reason there are data files in this PR? Are they different than #412?

src/cclib/io/wfxwriter.py
gaursagar: writer now considers `core electrons`

src/cclib/io/wfxwriter.py
gaursagar: Official description of wfx format does not talk about any special treatment for excited states.

src/cclib/io/wfxwriter.py
gaursagar: Renamed to `pi_cube_inv`

src/cclib/io/wfxwriter.py
gaursagar: Moved.

src/cclib/io/wfxwriter.py
gaursagar: Moved to module level.

test/io/testwfxwriter.py
gaursagar: I couldn't check `_normalized_mocoeffs` directly, because it also depends upon `ccdata.mocoeffs`. `mocoeffs` are different across programs, for same set of input coordinates and basis set.
Hence, I am checking `_norm_mat` instead. It ensures that the other three data used in normalization i.e,  `prim_type`, `alpha`, `prim_coeff` are being correctly computed.

ATenderholt:
I think this can now be merged. Any additional comments by @langner and @berquist ?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,488 @@
...
+        return [self.pt.element[Z]+str(i) for i, Z in
+                enumerate(self.ccdata.atomnos, start=1)]
+
+    def _atomic_nos(self):
+        """Atomic numbers of elements."""
+        return [str(Z) for Z in self.ccdata.atomnos]
+
+    def _nuclear_charges(self):
+        """Nuclear charges."""
+        if hasattr(self.ccdata, 'coreelectrons'):
```
langner: Please calculate the charges and then return at end.

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,488 @@
...
+
+    def _no_beta_electrons(self):
+        """Number of beta electrons in molecule."""
+        return self._no_electrons() - self._no_alpha_electrons()
+
+    def _spin_mult(self):
+        """Electronic Spin Multiplicity"""
+        return self.ccdata.mult
+
+    def _prim_centers(self):
```
langner: Can you get this in an easier way from `atombasis`?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,488 @@
...
+        prim_centers = []
+        for nuc_num, atom in enumerate(self.ccdata.gbasis, start=1):
+            for prims in atom:
+                prim_centers += [nuc_num] * num_orb[prims[0]]\
+                                * len(prims[1])
+
+        return _list_format(prim_centers, 10, '%d ')
+
+    def _rearrange_modata(self, data):
+        """Rearrange MO related data."""
```
langner: Can you please provide a more detailed explanation in the docstring of what this does and why?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,488 @@
...
+            data = data.tolist()
+
+        pos_yyx = [key for key, val in enumerate(prim_types)
+                   if val == 17]
+        pos_yyz = [key for key, val in enumerate(prim_types)
+                   if val == 16]
+
+        if pos_yyx:
+            for pos in pos_yyx:
+                data.insert(pos-3, data.pop(pos))
```
langner: Consider not mutating arguments, or stating clearly in the docstring that this is the case.

src/cclib/io/wfxwriter.py
langner: Could you please comment in some appropriate place that spherical basis set representations are not supported?

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,488 @@
...
+        if pos_yyz:
+            for pos in pos_yyz:
+                data.insert(pos+3, data.pop(pos + 1))
+
+        return data
+
+
+    def _prim_types(self):
+        """Primitive type numbering starts at 1.
```
langner: Please start the docstring with a description of what the method does.

src/cclib/io/wfxwriter.py
```diff
@@ -0,0 +1,488 @@
...
+        for atom in self.ccdata.gbasis:
+            for prims in atom:
+                prim_orb = []
+                for i in range(num_orb[prims[0]]):
+                    prim_orb += [(idx_orb[prims[0]]  + i)]\
+                                * len(prims[1])
+                prim_types += prim_orb
+        return prim_types
+
+    def _prim_types_section(self):
```
langner: I'm not sure if this corresponds to a section in the WFX file as well? Some of these method do, but don't have this _section suffix. Could you be consistent in some way so it's easy to tell? Please also include a comment in the class docstring how to tell which methods correspond to a section in the output file.

src/cclib/io/wfxwriter.py
gaursagar: Docstring updated to explain the need of rearranging.

src/cclib/io/wfxwriter.py
gaursagar: Refactored to now have a single return.

src/cclib/io/wfxwriter.py
gaursagar: `atombasis`. does not directly translate to what is expected to be written in the section.
For instance for water_mp2:
`atombasis` is 
```python
[[0, 1, 2, 3, 4], [5], [6]]
```
and the expected output is
```xml
<Primitive Centers>
1 1 1 1 1 1 1 1 1 1 
1 1 1 1 1 2 2 2 3 3 
3 
</Primitive Centers>
```
The conversion would require knowing the name of orbitals, and hence using `ccData.gbasis`.

src/cclib/io/wfxwriter.py
gaursagar: Docstring updated to explain the need of such rearrangements.

src/cclib/io/wfxwriter.py
gaursagar: Docstring updated.

src/cclib/io/wfxwriter.py
gaursagar: Method docstrings returning data to be written have been updated to include 
`Section: Section Name`.
```python
    def _prim_types(self):
        """Section: Primitive Types."""
        ...
```
Other methods are helper functions.

src/cclib/io/wfxwriter.py
gaursagar: Docstring updated with more details.

PR #402, opened 2017-07-28
Bring ORCA IR unit test precisions back to sane values

langner:
This is a follow up to cclib/cclib#393 and cclib/cclib#397 and needs to be merged AFTER cclib/cclib-data#59.


berquist:
The `push` build is failing because this PR is on top of the old problem code (see https://github.com/cclib/cclib/pull/395), and the `pr` build is from the sucessful (non-conflicting) merge of the branch with current master. At least, I think that's the behavior.

See https://github.com/travis-ci/travis-ci/issues/3241 for the root of this behavior.

Because of this, I'm going to merge it.

Issue #136, opened 2014-10-05
Non-stationary point used in vib unit tests


langner:
The coordinates we use for vibrational analysis, at least in some cases, is not a stationary point. While this probably does not influence the performance of parsers, it would be nice to do these calculations at startionary points. I'm not sure which programs check this and report it, but GAMESS-US does:
https://github.com/cclib/cclib/blob/master/data/GAMESS/basicGAMESS-US2012/dvb_ir.out#L11303

(I dug this up from an old email on the ML)

langner:
@berquist What do you think about this? Is there something specific action items we can identify here to make this concrete?

berquist:
I'm confused because other than an old ORCA unit test that's probably a bad calculation or a bug, an ORCA regression, and two ADF unit tests, none of IR outputs present negative frequencies in the regular output (not talking about the "low frequencies" section because not all programs have that).

* The first step is to identify the problematic files. There is some inconsistency in which structure is used for each program (see https://github.com/cclib/cclib/issues/407).
* If a new geometry should be used, it can't be the same as the one used for geometry optimizations unless you're ok with a geo opt test only taking 1-2 steps.
* Programs do different things with symmetry and reorientation of inputs, which is a problem because the DFT grid is not invariant to this reorientation. The effect should be small for B3LYP but I don't have proof yet. Disabling reorientation is a possibility but this may affect testing `vibsyms`. I am against turning off lots of features selectively like this because it isn't the default usage, which is the most important case for us to cover; that should be for regressions.

A partial fix to the solution is

2. use that program's (+ version) converged geometry for a frequency calculation with all default symmetry enabled.

The most complete solution is to

0. switch to Hartree-Fock for most tests,
1. pick a uniform non-stationary starting geometry for most calculation types across all programs,
2. use that program's (+ version) converged geometry for a frequency calculation with all default symmetry enabled.

langner:
I like the complete solution most - let's coordinate after v1.5.3 is out.

PR #403, opened 2017-07-28
Updates ORCA geometry parsing.

jevandezande:
Simplifies geometry parsing by combining single point and optimization geometry parsing. Adds parsing for atommasses and a test case.

Notes
-----
- Apparently different programs have different weights for atoms, necessitating large error bars.

Questions
---------
- [x] Is there a way to run a test only for a set of parsers (e.g. with a decorator like `runForParser` akin to `skipForParser`)?
- [x] I had to use the ugly test `self.assertTrue(numpy.isclose(atommass, H, 3) or numpy.isclose(atommass, C, 1))` in order to pass the regressions as some regressions have the atoms ordered differently. Could the regressions be changed to have the same starting geometries and atom order?

Status
------
- [ ] Regression tests need to be merged, PR cclib/cclib-data#60

### DIFF
--- a/test/data/testGeoOpt.py
+++ b/test/data/testGeoOpt.py
@@ -123,6 +123,18 @@ class GenericGeoOptTest(unittest.TestCase):
         msg = "len(atomcoords) is %d but len(geovalues) is %d" % (count_coords, count_geovalues)
         self.assertEquals(count_geovalues, count_coords, msg)
 
+    @skipForParser("GAMESSUK", "Not implemented.")
+    @skipForParser("GAMESS", "Not implemented.")
+    @skipForParser("Jaguar", "Not implemented.")
+    @skipForParser("Molpro", "Not implemented.")
+    @skipForParser("NWChem", "Not implemented.")
+    @skipForParser("QChem", "Not implemented.")
+    def testatommasses(self):
+        """Are atommasses correct?"""
+        H, C = 1.008, 12.011
+        for atommass in self.data.atommasses:
+            self.assertTrue(numpy.isclose(atommass, H, 3) or numpy.isclose(atommass, C, 1))
+
     def testgeovalues_scfvalues(self):
         """Are scfvalues consistent with geovalues?"""
         count_scfvalues = len(self.data.scfvalues) - self.extrascfs
--- a/test/data/testGeoOpt.py
+++ b/test/data/testGeoOpt.py
@@ -123,18 +123,6 @@ class GenericGeoOptTest(unittest.TestCase):
         msg = "len(atomcoords) is %d but len(geovalues) is %d" % (count_coords, count_geovalues)
         self.assertEquals(count_geovalues, count_coords, msg)
 
-    @skipForParser("GAMESSUK", "Not implemented.")
-    @skipForParser("GAMESS", "Not implemented.")
-    @skipForParser("Jaguar", "Not implemented.")
-    @skipForParser("Molpro", "Not implemented.")
-    @skipForParser("NWChem", "Not implemented.")
-    @skipForParser("QChem", "Not implemented.")
-    def testatommasses(self):
-        """Are atommasses correct?"""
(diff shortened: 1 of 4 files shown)


berquist:
> Apparently different programs have different weights for atoms, necessitating large error bars.

I think `atommasses` was originally intended to hold the isotopic masses used in frequency calculations. This may explain part of it (where the masses were originally taken from), plus the printing precision. Due to the low precision it's hard to tell exactly what ORCA is printing here, but I assume it's most common isotope, which is the most important thing.

> Is there a way to run a test only for a set of parsers (e.g. with a decorator like runForParser akin to skipForParser)?

Maybe something like this (invert the run/skip condition from `skipForParser`):

```python
def runForParser(parser, msg):
    """Return a decorator that runs the test for only the specified parser."""
    def testdecorator(testfunc):
        def testwrapper(self, *args, **kwargs):
            if self.logfile.logname == parser:
            else:
                self.skipTest(msg)
        return testwrapper
    return testdecorator
```

We probably don't want to do this; by forcing opt-out as the default behavior, it's more clear when reading the tests what specific problems are with specific parsers. This would make it harder to go back and update tests so that they run for all parsers.

> I had to use the ugly test self.assertTrue(numpy.isclose(atommass, H, 3) or numpy.isclose(atommass, C, 1)) in order to pass the regressions as some regressions have the atoms ordered differently. Could the regressions be changed to have the same starting geometries and atom order?

Yes, the geometries are inconsistent between programs, plus they have [other problems](https://github.com/cclib/cclib/issues/136). Fixing this in general will be a difficult undertaking because no one has access to ADF, GAMESS-UK, or Jaguar right now.

For now, it would be better to test both elements (so use `and` rather than `or`) by searching for the first instance of each.

langner:
Yes, `atommasses` was meant for the common isotope.

I agree with Eric, and would prefer to skip with skipping parsers. If you want, you could create a `skipForPrasers` decorator to save a few lines.

Although we make an effort to standardize unit test logfiles, there will be no such effort for old unit tests in regressions. However, we can skip single tests or make tweaks to old unit tests in regressions as needed.

langner:
> Yes, atommasses was meant for the common isotope.

Correction - it was to hold whatever the program prints. So, you might request some rare isotope if you're studying its effect on something.

test/data/testGeoOpt.py
```diff
@@ -123,6 +123,18 @@ def testgeovalues_atomcoords(self):
         msg = "len(atomcoords) is %d but len(geovalues) is %d" % (count_coords, count_geovalues)
         self.assertEquals(count_geovalues, count_coords, msg)
 
+    @skipForParser("GAMESSUK", "Not implemented.")
```
langner: So, do we need this test given we do an equivalent test in `testSP`? Are the masses formatted any differently for geometry optimization?

Also, please take the time time to enable this test for ORCA in `testSP`. Thanks!

test/data/testSP.py
```diff
@@ -101,15 +101,17 @@ def testatombasis(self):
...
     @skipForLogfile('Psi/basicPsi4.0b5', 'atommasses not implemented yet')
     @skipForParser('QChem', 'atommasses not implemented yet')
     def testatommasses(self):
         """Do the atom masses sum up to the molecular mass?"""
         mm = 1000*sum(self.data.atommasses)
-        msg = "Molecule mass: %f not %f +- %fmD" % (mm, self.molecularmass, self.mass_precision)
-        self.assertAlmostEquals(mm, self.molecularmass, delta=self.mass_precision, msg=msg)
+        molecularmass = self.molecularmass
+        if self.logfile.logname == 'ORCA':
+            molecularmass = 130190.0
```
berquist: @langner Would it be better to create a `ORCASPTest` class? There is precedent: see `Psi3SPTest`, where only the total energy is different.

langner: Sure... that's the convention we've been using and this fits the use case.

PR #409, opened 2017-08-04
Updates to Travis CI testing infrastructure

berquist:
* Move the testing commands into a shell script.
* Move to the new sudo-less container-based testing infrastructure. Not that this matters much for us, but tests should start faster.


langner:
No need for different command for python3?

berquist:
> No need for different command for python3?

Nope, everything is done in virtualenvs that are automatically created and sourced.

```
$ python --version
Python 3.6.2
$ pip --version
pip 9.0.1 from /home/travis/virtualenv/python3.6.2/lib/python3.6/site-packages (python 3.6)
```

PR #410, opened 2017-08-04
Update copyright year to 2017

berquist:
Only 7 months or so late...

Also changes a version number to 1.5.1 in `src/cclib/__init__.py`

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2016, the cclib development team
+Copyright (c) 2017, the cclib development team
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/data/regression_download.sh
+++ b/data/regression_download.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (c) 2016, the cclib development team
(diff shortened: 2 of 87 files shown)


langner:
I dunno... aren't we supposed to have the first year of publication here? https://softwareengineering.stackexchange.com/questions/130478/what-copyright-date-for-an-update-to-an-open-source-project-from-last-year

berquist:
At some point they were all changed over from that format: 1a2a2f22

langner:
> At some point they were all changed over from that format: 1a2a2f2

I know. What do you think we should do?

berquist:
According to that thread, it isn't even required to have that notification there with the date, but I don't think we should remove it completely.

I have no preference for either of:

1. The current year
2. The year when the file was created - the current year

It probably isn't worthwhile to go back to comma-separated years, where only the years the files were modified are present. That's what `git blame` is for.

langner:
> According to that thread, it isn't even required to have that notification there with the date, but I don't think we should remove it completely.
> 
> I have no preference for either of:
> 
> The current year
> The year when the file was created - the current year
> It probably isn't worthwhile to go back to comma-separated years, where only the years the files were modified are present. That's what git blame is for.

I'm fine with what you did.

PR #411, opened 2017-08-07
Fixes SCF and geometry parsing for heavy elements.

jevandezande:
Heavy elements can cause problems with the SCF printing, as a space was omitted between some scf values. Also fixes problem with atommass parsing when using ECPs.

We should probably add a regression that has ECPs for Orca.

As for the SCF printing, I've talked with the developers and they are hesitant to change anything as some other codes may use fixed format parsing. However, the SCF may undergo a rewrite in the near future (giving cclib even more headaches).

Questions
---------
- [x] Do all regression tests have to have a successfully completed output file? I often use cclib to check currently running and failed jobs, which is how I found this bug.

### DIFF
--- a/src/cclib/parser/orcaparser.py
+++ b/src/cclib/parser/orcaparser.py
@@ -350,6 +350,8 @@ class ORCA(logfileparser.Logfile):
             line = next(inputfile)
             self.atommasses = []
             while len(line) > 1:
+                if line == '* core charge reduced due to ECP\n':
+                    break
                 no, lb, za, frag, mass, x, y, z = line.split()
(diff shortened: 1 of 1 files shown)


langner:
> Do all regression tests have to have a successfully completed output file? I often use cclib to check currently running and failed jobs, which is how I found this bug.

No, regression don't need to be successfully completed output files. We expect cclib not to choke on non-standard logfiles.

Please do supply the file where you encountered this, though - thanks!

jevandezande:
I added the file to the regression tests, PR cclib/cclib-data#61.

langner:
Thanks!

PR #414, opened 2017-08-13
Added reading chelpg charges to ORCA

richardjgowers:
This adds reading chelpg charges from ORCA.  I've got a test file which has these in, but I couldn't figure out where to add a test just for the ORCA parser?  I'll put up a PR for the docs on the doc repo

### DIFF
--- a/src/cclib/parser/orcaparser.py
+++ b/src/cclib/parser/orcaparser.py
@@ -794,6 +794,39 @@ class ORCA(logfileparser.Logfile):
             if has_spins:
                 self.atomspins["lowdin"] = spins
 
+        #CHELPG Charges            
+        #--------------------------------
+        #  0   C   :       0.363939
+        #  1   H   :       0.025695
+        # ...
+        #--------------------------------
+        #Total charge:    -0.000000
+        #--------------------------------
+        if line.startswith('CHELPG Charges'):
+            has_spins = 'AND SPIN POPULATIONS' in line
+
+            if not hasattr(self, "atomcharges"):
+                self.atomcharges = {}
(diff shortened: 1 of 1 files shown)


berquist:
Thanks for the patch!

>  I've got a test file which has these in, but I couldn't figure out where to add a test just for the ORCA parser?

If it's the usual test molecule that we use (divinylbenzene), then what you can do is add a test in `test/data/testSP.py`, probably as a method under `GenericSPTest`, then add `@skipForParser` decorators for all parsers except ORCA. That way we can use the test in the future for Gaussian and Q-Chem, both of which also support CHELPG charges.

If it's a one-off calculation, then we have a separate repository for regression and contributed tests: https://github.com/cclib/cclib-data.

Also, would it be possible to turn parsing all 3 charge methods (Mulliken, Lowdin, CHELPG) into a single class function? The code is basically identical between the 3. See the Q-Chem parser below:

https://github.com/cclib/cclib/blob/e7cc329adcdcf1347c686b826cab37bf07dd0a0e/src/cclib/parser/qchemparser.py#L162

richardjgowers:
@berquist thanks, I refactored the charge reading, I'll look at adding the test to the other repo

src/cclib/parser/orcaparser.py
```diff
@@ -842,6 +809,48 @@ def extract(self, inputfile, line):
                 polarizability.append(line.split())
             self.polarizabilities.append(numpy.array(polarizability))
 
+    def parse_charge_section(self, line, inputfile, chargestype):
```
berquist: Can you add a docstring?

berquist:
Great, thank you!

Issue #200, opened 2015-05-07
Test bridges with Travis


langner:
It's not trivial to get  the whole Python/OpenBabel thing working to test the openbabel bridge in cclib, currnetly we're not testing bridges automatically. See #190 for the details. It would be nice to figure it out in the future, of course.

PR #415, opened 2017-08-13
Test bridges with travis

berquist:
Implements https://github.com/cclib/cclib/issues/200.

Currently, the only bridge that has tests implemented is Open Babel. This PR installs all possible dependencies (Open Babel, BioPython, and PyQuante) so that tests can be written in the future.

Notes:

* The original PyQuante (v1.6.5) only works under Python 2.7. A future bridge for pyquante2, which is under active development, would be nice.
* BioPython never supported Python 3.2.
* This uses an older version of the Open Babel library, but the latest PyPI version of the Python bindings.


langner:
Openbabel doesn't seem to be installing successfully.

berquist:
The build in my branch is using Ubuntu 14.04, and the PR is using 12.04, leading to very different versions of Open Babel.

langner:
Thanks!

PR #408, opened 2017-08-03
Make ccData_optdone_bool persistent during type check, and test explicitly

langner:

PR #417, opened 2017-08-17
Make ccData_optdone_bool persistent during type check, and test explicitly

langner:
This is a reincarnation of #408


langner:
Note that cclib/cclib-data#64 needs to go in before this (to unbreak regresssions).

Issue #420, opened 2017-08-20
Parsing multiple files after each other for polarizability could be ambiguous


xymaxim:
Parsing a DALTON output file which doesn't contain all polarizability components (`**RESPONSE .DIPLNX/Y/Z` keyword) just before another one with `.DIPLEN` keyword results in the output containing ambiguous numbers.

**How to reproduce this?**

See a [Gist](https://gist.github.com/xymaxim/b0a78e9fa2e545b8163f07ef6f60a21f) I've prepared to illustrate the issue. The first output file contains `.DIPLNX` keyword in the input—in this case *A* and *B* operators are set only to the *x* dipole operator. 

```python
    from cclib.parser import DALTON

    print(DALTON('Trp_polar_response_diplnx.out').parse().polarizabilities)
    print(DALTON('Trp_polar_response.out').parse().polarizabilities)
    print(DALTON('Trp_polar_response_diplnx.out').parse().polarizabilities)
```

And what went wrong? Notice that the last parsing output contains meaningful numbers from the previous file. Surprisingly. The source of that is the line below:

https://github.com/cclib/cclib/blob/73aeee78ad2a45963928747eb7b3723b17366d34/src/cclib/parser/daltonparser.py#L1102

NumPy's reference [warns](https://docs.scipy.org/doc/numpy/reference/generated/numpy.empty.html) users about using `empty()` with caution. Look at the example:

```python
    >>> a = np.empty(2); a[0] = 3.14
    >>> print(a)
    array([  3.14000000e+000,   9.65124433e-317])
    >>> print(np.empty(2))
    array([  3.14000000e+000,   9.65124433e-317])
```

**Real-life example**
Imagine that someone (inattentive or not) wants to find the trace of a tensor and calculate isotropic polarizability value for a bunch of output files. With the mentioned cuckoo-like behavior from `cclib` it could be potentially a ticking time bomb. 

**Nested issue and solution**
Actually this issue is not about how we parse the output nor shamelessly modifying the user data (it's not so). While parsing a single file with `.DIPLNX/Y/Z` run, the diagonal components of polarizability mixed up with meaningless off-diagonal numbers and that's fine (for attentive users).

Probably the better idea is to fill an empty array with `NaN` values: `a = np.empty(shape) * np.nan`. Should I create a pull request that can be used as a patch? However, some of other parsers can also be affected by this issue.

berquist:
[...]

PR #421, opened 2017-08-20
Fill an empty polarizability diplen array with np.nan values

xymaxim:


langner:
It'd be great if we could reproduce this problem and add a test to make sure it stays fixed. Do you have something at hand?

berquist:
It could be adding `Trp_polar_response_diplnx.out` to regressions and comparing to one of the other DALTON polarizability outputs there, all of which have the full tensor.

xymaxim:
Agreed. I'll take care of it. @berquist have reasonably pointed out about filling the missing values with  `np.nan`. And it should be conventional. Since this issue only refers to programs that provide ability to selectively calculate a tensor components, would it be superfluous to enlarge this pull request for other parsers?

berquist:
> Since this issue only refers to programs that provide ability to selectively calculate a tensor components, would it be superfluous to enlarge this pull request for other parsers?

Searching for `empty` returns the matches in ADF and DALTON, plus GAMESS and Q-Chem, but not all for polarizabilities. It should be fine to add this in all cases.

xymaxim:
ADF parser doesn't experience the issue #420 since the ADF program sets the missing components to zero and print these values in matrix format. In this case, should I add a test for ADF parser to ensure that  `np.nan` values are correctly replaced? If so, there was a discussion about standardized inputs and I'm wondering if there are any requirements at the moment (especially for xc-functional)?

langner:
> ADF parser doesn't experience the issue #420 since the ADF program sets the missing components to zero and prints these values in matrix format. In this case, should I add a test for ADF parser to ensure that np.nan values are correctly replaced?

No, I think we should stick to parsing the values that the program prints. Even if those values are not what we would expect based on the method/theory.

Issue #360, opened 2017-03-17
Cannot parse gaussian IRC log


bertadenes:
I encountered an error trying to parse IRC outputs. The structure of the output is slightly different, thus when gaussianparser.py tries to parse the step, (line 822 in the latest version) it fails to cast a string. It is due to the fact that IRC has an additional "Pt X" in the beginning on the line.
Eventually I think it can be just fixed by just adding a couple lines, see the following example.
The same change does work with version 1.4 as well.
[gaussianparser.py.zip](https://github.com/cclib/cclib/files/850454/gaussianparser.py.zip)

bertadenes:
[gaussianparser.py.zip](https://github.com/cclib/cclib/files/850635/gaussianparser.py.zip)
This one can handle IRC over 100 points as well.

langner:
Hi, would you like to create a pull request, or should we just patch it for you?

Would it be possible for you to also provide us with the logfile to test this?

bertadenes:
I could create a pull request with the patched file, but it would require add me in order to create a new branch, wouldn't it?
On the other hand, I cannot give my first problematic output away, I will create another relevant example. Thank you for your understanding.

langner:
@bertadenes If you manage to create another relevant example, please reopen this and we can continue. Meanwhile I'll close this since some time has passed. Please let us know if you have any problems or questions.

bertadenes:
@langner Sorry, this just slipped through my table lately. Yes, I do have a relevant example which was properly processed by using the modified code I uploaded earlier.
[IRCparse-example.out.zip](https://github.com/cclib/cclib/files/1003390/IRCparse-example.out.zip)

langner:
Thanks Denes! I will take a look soon.

langner:
The file attached here being added to regressions in cclib/cclib-data#67.

PR #424, opened 2017-08-23
Fix IRC parsing with Gaussian

LaTruelle:
Refers to Issue #360 
Simple fix that allows proper parsing of IRCs.


langner:
Confirmed this patch works for the file attached in #360

PR #425, opened 2017-08-23
Parses gradients, metadata, and vibfreqs for ORCA

jevandezande:
Adds parsing of gradients and metadata and updates parsing of vibfreqs of linear molecules. A regression was added in cclib/cclib-data#66 that needs to be pulled before these regressions can pass.

Also updates spelling and some comments.

### DIFF
--- a/src/cclib/parser/orcaparser.py
+++ b/src/cclib/parser/orcaparser.py
@@ -792,18 +792,23 @@ NAME = input.dat
 
             self.skip_lines(inputfile, ['d', 'b'])
 
-            self.vibfreqs = numpy.zeros((3 * self.natom,), "d")
+            vibfreqs = numpy.zeros(3*self.natom)
 
-            for i in range(3 * self.natom):
-                line = next(inputfile)
-                self.vibfreqs[i] = float(line.split()[1])
+            for i, line in zip(range(3*self.natom), inputfile):
+                vibfreqs[i] = line.split()[1]
 
-            if numpy.any(self.vibfreqs[0:6] != 0):
-                msg = "Modes corresponding to rotations/translations "
-                msg += "may be non-zero."
+            nonzero = numpy.nonzero(vibfreqs)[0]
+            if len(nonzero) > 3*self.natom - 6:
+                msg = "Modes corresponding to rotations/translations may be non-zero."
+                if len(nonzero) == 3*self.natom - 5:
+                    msg += '\n You can ignore this if the molecule is linear.'
                 self.logger.warning(msg)
+            elif len(nonzero) < 3*self.natom - 6:
+                self.logger.warning('Too few vibrational modes for molecule.')
+            # Take all modes after first
+            # Mode between imaginary and real modes could be 0
+            self.vibfreqs = vibfreqs[nonzero[0]:]
 
-            self.vibfreqs = self.vibfreqs[6:]
 
         if line[0:12] == "NORMAL MODES":
             """ Format:
--- a/src/cclib/parser/orcaparser.py
+++ b/src/cclib/parser/orcaparser.py
@@ -62,23 +62,21 @@ class ORCA(logfileparser.Logfile):
         if "Program Version" in line:
             self.metadata["package_version"] = line.split()[2]
 
-        """
-================================================================================
-                                        WARNINGS
-                       Please study these warnings very carefully!
-================================================================================
-
-Warning: TCutStore was < 0. Adjusted to Thresh (uncritical)
-
-WARNING: your system is open-shell and RHF/RKS was chosen
-  ===> : WILL SWITCH to UHF/UKS
-
-
-INFO   : the flag for use of LIBINT has been found!
-
-================================================================================
-        """
-        if "WARNINGS" == line.strip():
+        # ================================================================================
+        #                                         WARNINGS
+        #                        Please study these warnings very carefully!
+        # ================================================================================
+        #
+        # Warning: TCutStore was < 0. Adjusted to Thresh (uncritical)
+        #
+        # WARNING: your system is open-shell and RHF/RKS was chosen
+        #   ===> : WILL SWITCH to UHF/UKS
+        #
+        #
+        # INFO   : the flag for use of LIBINT has been found!
+        #
+        # ================================================================================
+        if line.strip() == "WARNINGS":
             next(inputfile)
             self.skip_lines(inputfile, ['equals', 'blank'])
 
@@ -100,23 +98,21 @@ INFO   : the flag for use of LIBINT has been found!
                         line = next(inputfile)
                 line = next(inputfile)
 
-        """
-================================================================================
-                                       INPUT FILE
-================================================================================
-NAME = input.dat
-|  1> %pal nprocs 4 end
-|  2> ! B3LYP def2-svp
-|  3> ! Grid4
-|  4>
-|  5> *xyz 0 3
-|  6>     O   0   0   0
-|  7>     O   0   0   1.5
-|  8> *
-|  9>
-| 10>                          ****END OF INPUT****
-================================================================================
-"""
+        # ================================================================================
+        #                                        INPUT FILE
+        # ================================================================================
+        # NAME = input.dat
+        # |  1> %pal nprocs 4 end
+        # |  2> ! B3LYP def2-svp
+        # |  3> ! Grid4
+        # |  4>
+        # |  5> *xyz 0 3
+        # |  6>     O   0   0   0
+        # |  7>     O   0   0   1.5
+        # |  8> *
+        # |  9>
+        # | 10>                          ****END OF INPUT****
+        # ================================================================================
         if "INPUT FILE" in line:
             next(inputfile)
             name = next(inputfile).split()[-1]
@@ -363,15 +359,13 @@ NAME = input.dat
                 self.geotargets_names.append(name)
                 self.geotargets.append(target)
 
-        """
-        ------------------
-        CARTESIAN GRADIENT
-        ------------------
-
-        1   H   :    0.000000004    0.019501450   -0.021537091
-        2   O   :    0.000000054   -0.042431648    0.042431420
-        3   H   :    0.000000004    0.021537179   -0.019501388
-        """
+        # ------------------
+        # CARTESIAN GRADIENT
+        # ------------------
+        #
+        # 1   H   :    0.000000004    0.019501450   -0.021537091
+        # 2   O   :    0.000000054   -0.042431648    0.042431420
+        # 3   H   :    0.000000004    0.021537179   -0.019501388
         if line == 'CARTESIAN GRADIENT\n':
             self.skip_lines(inputfile, ['dashes', 'blank'])
 
@@ -444,15 +438,14 @@ NAME = input.dat
 
             self.geovalues.append(newvalues)
 
-        """ Grab Cartesian coordinates
-        ---------------------------------
-        CARTESIAN COORDINATES (ANGSTROEM)
-        ---------------------------------
-        H      0.000000    0.000000    0.000000
-        O      0.000000    0.000000    1.000000
-        H      0.000000    1.000000    1.000000
-        """
-        if line[0:33] == "CARTESIAN COORDINATES (ANGSTROEM)":
+        # Grab Cartesian coordinates
+        # ---------------------------------
+        # CARTESIAN COORDINATES (ANGSTROEM)
+        # ---------------------------------
+        # H      0.000000    0.000000    0.000000
+        # O      0.000000    0.000000    1.000000
+        # H      0.000000    1.000000    1.000000
+        if line[:33] == "CARTESIAN COORDINATES (ANGSTROEM)":
             next(inputfile)
 
             atomnos = []
@@ -470,16 +463,15 @@ NAME = input.dat
                 self.atomcoords = []
             self.atomcoords.append(atomcoords)
 
-        """ Grab atom masses
-        ----------------------------
-        CARTESIAN COORDINATES (A.U.)
-        ----------------------------
-        NO LB      ZA    FRAG     MASS         X           Y           Z
-        0 H     1.0000    0     1.008    0.000000    0.000000    0.000000
-        1 O     8.0000    0    15.999    0.000000    0.000000    1.889726
-        2 H     1.0000    0     1.008    0.000000    1.889726    1.889726
-        """
-        if line[0:28] == "CARTESIAN COORDINATES (A.U.)" and not hasattr(self, 'atommasses'):
+        # Grab atom masses
+        # ----------------------------
+        # CARTESIAN COORDINATES (A.U.)
+        # ----------------------------
+        # NO LB      ZA    FRAG     MASS         X           Y           Z
+        # 0 H     1.0000    0     1.008    0.000000    0.000000    0.000000
+        # 1 O     8.0000    0    15.999    0.000000    0.000000    1.889726
+        # 2 H     1.0000    0     1.008    0.000000    1.889726    1.889726
+        if line[:28] == "CARTESIAN COORDINATES (A.U.)" and not hasattr(self, 'atommasses'):
             next(inputfile)
             next(inputfile)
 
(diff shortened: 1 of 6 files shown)


test/data/testvib.py
```diff
@@ -23,15 +23,14 @@ class GenericIRTest(unittest.TestCase):
...
+                         (num_modes, len(self.data.atomnos), 3))
 
     def testlengths(self):
         """Are the lengths of vibfreqs and vibirs (and if present, vibsyms) correct?"""
-        numvib = 3*len(self.data.atomnos) - 6
-        self.assertEqual(len(self.data.vibfreqs), numvib)
+        numvib = len(self.data.vibfreqs)
         if hasattr(self.data, 'vibirs'):
             self.assertEqual(len(self.data.vibirs), numvib)
         if hasattr(self.data, 'vibsyms'):
```
ATenderholt: I'm not sure I agree with this approach. I'd rather test explicitly for 3N-6 in ```GenericIRTest``` and have a ```GenericLinearIRTest``` that tests for 3N-5.

This could (probably) be accomplished with a simple function that converts ```data.atomnos``` to ```num_vibs```.

src/cclib/parser/orcaparser.py
```diff
@@ -251,6 +364,31 @@ def extract(self, inputfile, line):
...
+        CARTESIAN GRADIENT
+        ------------------
+
+        1   H   :    0.000000004    0.019501450   -0.021537091
+        2   O   :    0.000000054   -0.042431648    0.042431420
+        3   H   :    0.000000004    0.021537179   -0.019501388
+        """
+        if line == 'CARTESIAN GRADIENT\n':
+            next(inputfile)
+            next(inputfile)
```
ATenderholt: Use ```skip_lines(inputfile, ['dashes', 'blank'])``` instead of two ```next(inputfile)```.

src/cclib/parser/orcaparser.py
```diff
@@ -62,6 +62,119 @@ def extract(self, inputfile, line):
...
+
+
+INFO   : the flag for use of LIBINT has been found!
+
+================================================================================
+        """
+        if "WARNINGS" == line.strip():
+            next(inputfile)
+            next(inputfile)
+            next(inputfile)
```
ATenderholt: Use ```skip_lines(inputfile, ['equals', 'blank'])``` instead of the last two ```next(inputfile)```.

src/cclib/parser/orcaparser.py
```diff
@@ -62,6 +62,119 @@ def extract(self, inputfile, line):
...
+                        while line[0] != '*':
+                            # Strip basis specification that can appear after coordinates
+                            line = line.split('newGTO')[0]
+                            coords.append(splitter(line))
+                            line = next(inputfile)[6:].strip()
+
+                line = next(inputfile)
+
+            self.metadata['methods'] = methods
+            self.metadata['coords'] = coords
```
ATenderholt: How is ```metadata['coords']``` different than ```data.atomcoords```? Are you trying to capture input coordinates when they aren't specified as XYZ?

src/cclib/parser/orcaparser.py
```diff
@@ -62,6 +62,119 @@ def extract(self, inputfile, line):
...
+                if line[0] == '!':
+                    methods += line[1:].split()
+                # Impossible to parse without knowing whether a keyword opens a new block
+                elif line[0] == '%':
+                    pass
+                # Geometry block
+                elif line[0] == '*':
+                    vals = line[1:].split()
+                    coord_type = vals[0].lower()
+                    def splitter(line): return line.split()
```
ATenderholt: Do you need this default ```splitter```?

src/cclib/parser/orcaparser.py
jevandezande: The input coords can differ from the first printed coordinates if the molecule is symmetrized with `UseSym`. Also, if the coordinates are input with `gzmat` (gaussian style zmatrix), they don't get printed in that style later.

test/data/testSP.py
```diff
@@ -241,19 +241,10 @@ def testmoments(self):
...
-    @skipForParser('ORCA', 'Does not support metadata yet')
-    @skipForParser('Psi', 'Does not support metadata yet')
-    @skipForParser('QChem', 'Does not support metadata yet')
     def testmetadata(self):
         """Does metadata have expected keys and values?"""
         self.assertTrue(hasattr(self.data, "metadata"))
-        self.assertIn("basis_set", self.data.metadata)
         self.assertIn("methods", self.data.metadata)
         self.assertIn("package", self.data.metadata)
         self.assertIn("package_version", self.data.metadata)
```
ATenderholt: Why was ```testSP.py``` changed for this pull request? Seems like it's mostly unrelated, and should be a separate PR.

jevandezande: It is related to the metadata parsing as a basis set may not be named in the metadata. Some programs will set a default basis set while for others there are many various different ways to read in a basis set. This can be difficult to properly parse (e.g. in PSI4 you can do `set basis cc-pVTZ` or `energy('CCSD(T)/cc-pVTZ')`) and can mean that there is non basis set name (e.g. entering parameters directly into input file). Furthermore, different basis sets can be specified for different atoms, thus there would be multiple basis sets in use.

ATenderholt: Good points, especially about different basis sets for different atoms.

What do @langner and @berquist think? Is ```metadata['basis_set']``` supposed to be flexible enough to handle all of these cases?

test/data/testvib.py
langner: I would suggest putting the LinearIRTest into regressions.

src/cclib/parser/orcaparser.py
langner: Our goal is to parse one set of coordinates for each point where the internal structure is different. Symmetrization and change of format (z-matrix) do not change the structure. If you want to print in a different format, you can easily take the parsed coordinates and transform them.

test/data/testSP.py
langner: I think so. As a string, `metadata['basis_set']` can contain whatever you want, and there are some common ways to specify split basis sets, etc. The intention of `metadata` was to have something to store extra information we don't yet parse or won't ever parse. It is not meant to backwards compatible.

We do parse the basis set parameters in `gbasis` so in a sense `metadata['basis_set']` is redundant. It is only a name.

src/cclib/parser/orcaparser.py
```diff
@@ -62,6 +62,118 @@ def extract(self, inputfile, line):
         if "Program Version" in line:
             self.metadata["package_version"] = line.split()[2]
 
+        """
```
langner: Could you please use the format we typically have for these comments? Namely, start each line with`#` at the indentation level at which you are commenting. `"""` is meant for docstrings.

src/cclib/parser/orcaparser.py
```diff
@@ -62,6 +62,118 @@ def extract(self, inputfile, line):
...
+
+WARNING: your system is open-shell and RHF/RKS was chosen
+  ===> : WILL SWITCH to UHF/UKS
+
+
+INFO   : the flag for use of LIBINT has been found!
+
+================================================================================
+        """
+        if "WARNINGS" == line.strip():
```
langner: It is more Pythonic to do this the other way around, namely `line.strip() == "WARNINGS".

test/data/testSP.py
ATenderholt: So my main concern with this change in ```testSP.py``` is that we're removing a check for ```metadata['basis_set']```. If we don't expect it to always be in ```metadata```, then I'm fine with removing it. Otherwise, I'd prefer to not modify this test.

berquist: I am fine with changing it, because 1. presumably no one is using it right now, and 2. we haven't standardized it (or anything else in `metadata`), in part because it's somewhere between tricky and impossible to parse.

I'm not sure that if we ever choose to properly support a mixed/general basis, we'd expect this string to be there, because it may be misleading.

langner: I don't think we will every standardize `metadata` since the whole point was to parse things in there that are useful but are not necessarily viable attributes. But that doesn't mean we can't test it and expect specific contents for specific cases.

I think we should leave this is in the test. I think we should expect it to be parsed at least for out unit test suite.

langner:
I think regression tests will keep failing after https://github.com/cclib/cclib-data/pull/66 until this gets merged in. This is why I'd like to move the regression test code to the main repo.

langner:
> I think regression tests will keep failing after cclib/cclib-data#66 until this gets merged in. This is why I'd like to move the regression test code to the main repo.

There's no hurry since I reverted the submitted regression file for now.

test/data/testSP.py
jevandezande: In ORCA there is simply no easy way to parse a basis set name without having a list of all possible basis set names, since they are passed in using the multikeyword bang line. I have access to the ORCA source code and thus could find all the possible  basis sets (~400), but this would be rather ugly and subject to change. Instead, we could parse the basis set information block.

```
BASIS SET INFORMATION
---------------------
There are 3 groups of distinct atoms

 Group   1 Type Ir  : 7s6p5d1f contracted to 6s3p2d1f pattern {211111/411/41/1}
 Group   2 Type O   : 7s4p1d contracted to 3s2p1d pattern {511/31/1}
 Group   3 Type H   : 4s1p contracted to 2s1p pattern {31/1}

Atom   0Ir   basis set group =>   1
Atom   1Ir   basis set group =>   1
Atom   2O    basis set group =>   2
```

Would this be useful? I'm not sure what there is of value in this section.
How should I proceed?

ghutchis:
I don't think the basis set information section is particularly useful to get the basis name. You can make some educated guesses (e.g., double-zeta, triple-zeta, etc.) but I'm not sure you can infer the actual basis from that.

I don't think it would be too bad to match basis set names because they have general patterns that could be constructed by regex - I'd have to think carefully about the patterns, but they'd be something like these pseudo-code examples:

* STO-?G
* #-##G?
* #-###G?
* cc-pV?Z
* aug-cc-pV?Z
* pc-?
* def2-*

This approach would need some testing, but I think it's possible to code up some regexp to match the basis set keywords for all the major families without needing a list of 400+ basis sets.

langner:
I'm OK with the regex approach, or any other that would get you the basis set name if you want it parsed for `metadata`.

But, I also think you don't need to solve this problem necessarily in this PR. There's a lot of other good stuff here. Why not pass on this specific thing for now and get everything else merged in?

test/data/testvib.py
```diff
@@ -147,17 +147,18 @@ def testatommasses(self):
 class GenericIRimgTest(unittest.TestCase):
     """Generic imaginary vibrational frequency unittest"""
 
+    def setUp(self):
+        self.numvib = 3*len(self.data.atomnos) - 6
```
berquist: Does this method need to be here if it's the same in the parent class?

test/data/testvib.py
```diff
@@ -222,6 +225,18 @@ class QChemRamanTest(GenericRamanTest):
     max_raman_intensity = 588
 
 
+class GenericLinearIRTest(GenericIRTest):
+    """Test for linear molecules, which have 3N - 5 vibrational modes"""
```
berquist: Can you add docstrings for all the methods?

test/data/testvib.py
```diff
@@ -222,6 +225,18 @@ class QChemRamanTest(GenericRamanTest):
     max_raman_intensity = 588
 
 
+class GenericLinearIRTest(GenericIRTest):
```
berquist: Related to https://github.com/cclib/cclib/issues/426#issuecomment-324831305, do we want this as a unit test or as part of regressions?

test/data/testGeoOpt.py
```diff
@@ -161,6 +161,17 @@ def testmoenergies(self):
...
 
+    @skipForParser("ADF", "Not implemented.")
+    @skipForParser("DALTON", "Not implemented.")
+    @skipForParser("GAMESS", "Not implemented.")
+    @skipForParser("GAMESSUK", "Not implemented.")
+    @skipForParser("Jaguar", "Not implemented.")
+    @skipForParser("Molpro", "Not implemented.")
+    @skipForParser("NWChem", "Not implemented.")
+    @skipForParser("Psi", "Not implemented.")
+    def testgrads(self):
```
berquist: Forgot to transfer the `GaussianGeoOptTest.testgrads` docstring

src/cclib/parser/orcaparser.py
```diff
@@ -62,6 +62,114 @@ def extract(self, inputfile, line):
...
+        # Warning: TCutStore was < 0. Adjusted to Thresh (uncritical)
+        #
+        # WARNING: your system is open-shell and RHF/RKS was chosen
+        #   ===> : WILL SWITCH to UHF/UKS
+        #
+        #
+        # INFO   : the flag for use of LIBINT has been found!
+        #
+        # ================================================================================
+        if line.strip() == "WARNINGS":
```
berquist: This is interesting, I like this a lot. At some point we should think about what gets sent to the logger and what gets added to metadata (or both).

src/cclib/parser/orcaparser.py
```diff
@@ -62,6 +62,114 @@ def extract(self, inputfile, line):
...
+            next(inputfile)
+            name = next(inputfile).split()[-1]
+            methods = []
+            line = next(inputfile)
+            coords = []
+            while line[0] == '|':
+                line = line[6:]
+                # Keywords block
+                if line[0] == '!':
+                    methods += line[1:].split()
```
berquist: I think we currently have `metadata['methods']` as just quantum mechanical methods and not all options, where the length is related to how many calculations have been run (like if you do `$new_job`, or `@@@` in Q-Chem or `--` in Gaussian).

berquist:
For reference, this is what metadata currently looks like with this PR:

```
Attempting to read data/ORCA/basicORCA2.9/dvb_ir.out
metadata
{'package': 'ORCA', 'methods': ['RKS', 'STO-3G', 'B3LYP', 'UseSym', 'NumFreq'], 'package_version': '2.9.1', 'warnings': ['Gradients needed for Numerical Frequencies\nSetting RunTyp to EnGrad\n', 'Gradients are needed\nforcing at least TightSCF\n'], 'info': ['the flag for use of LIBINT has been found!\n'], 'coords': [['C', -1.4152533224, 0.2302217854, 0.0], ['C', 1.4152533224, -0.2302217854, 0.0], ['C', -0.4951331558, 1.3144608674, 0.0], ['C', 0.4951331558, -1.3144608674, 0.0], ['C', 0.8894090436, 1.0909493743, 0.0], ['C', -0.8894090436, -1.0909493743, 0.0], ['H', -0.8795511985, 2.3437343748, 0.0], ['H', 0.8795511985, -2.3437343748, 0.0], ['H', 1.5779041557, 1.9450061275, 0.0], ['H', -1.5779041557, -1.9450061275, 0.0], ['C', 2.8845844962, -0.5210893778, 0.0], ['C', -2.8845844962, 0.5210893778, 0.0], ['H', 3.140335681, -1.5919605685, 0.0], ['H', -3.140335681, 1.5919605685, 0.0], ['C', 3.8800428103, 0.3822535424, 0.0], ['C', -3.8800428103, -0.3822535424, 0.0], ['H', 3.6946765858, 1.462438957, 0.0], ['H', -3.6946765858, -1.462438957, 0.0], ['H', 4.9316453546, 0.0711049543, 0.0], ['H', -4.9316453546, -0.0711049543, 0.0]]}
```

src/cclib/parser/orcaparser.py
langner: Agreed. I am fine with a `metadata['warnings']` thing for now.

src/cclib/parser/orcaparser.py
langner: Yes, let's stick to actual methods, and not properties.

src/cclib/parser/orcaparser.py
jevandezande: Symmetrization in ORCA change the geometry. When you put in something that is close to symmetric, it will change the geometry to make is more symmetric. I guess this is not that common of a use case, so I could remove it.

test/data/testvib.py
jevandezande: `GenericIRimgTest` does not inherit from anything other than `unittest.TestCase`. Should this test class be changed to inherit from `GenericIRTest`?

src/cclib/parser/orcaparser.py
jevandezande: The simple keyword input in ORCA (lines with `!`) freely mixes methods and properties, thus I'm not sure how to separate the two without checking every keyword against a set of valid keywords. This seems to be what the Gaussian metadata parser does. However, when doing DFT in ORCA, the functional name is used directly, without any flag indicating DFT. How should I go about parsing this?

langner:
This needs to be reworked after recent refactoring. @jevandezande do you plan on pushing this forward? If so, you should consider splitting it up into several PRs, so we can merge in things faster. It's just easier to deal with smaller changes.

jevandezande:
This has been rebased to include recent changes and separated into three new pull requests #505, #506, and #507.

Issue #125, opened 2014-08-31
NMR and EPR attributes


ATenderholt:
Just to start the discussion....

Taking a look at the Gaussian NMR page, I see there are options for different gauge origins and methods. Do we parse all of them, similar to how we handle atomic charges? That is, a dictionary with the various origins/methods as keys? Or just the most common? I think NWChem and GAMESS only support the GIAO method

Are there other print options (e.g. eigenvectors) that would be useful?

jevandezande:
I'm looking to start parsing the NMR section of ORCA, and was hoping we could get some sort of consensus on what should be parsed. ORCA 4 only supports GIAO. While there are other methods, such as IGLO (available in ORCA 3), they are generally not recommended, so I don't know if we need a separate method flag. I know very little about NMR and what is important, but my talks with a developer of the NMR module in ORCA suggested that just parsing the isotropic and anisotropic should be sufficient.

I would propose that nmr be a tuple of two numpy arrays `(isotropic, anisotropic)`. However, it could also be a dictionary wherein `isotropic`, `anisotropic`, `shielding_tensor`, etc. are keys. Thoughts?

Here is a sample of all the data ORCA prints out for and NMR computation (He atom).
```
---------------
CHEMICAL SHIFTS
---------------

Note: using conversion factor for au to ppm alpha^2/2 =   26.625677252 
Doing GIAO para- and diamagnetic shielding integrals analytically     ...done
Doing remaining GIAO terms numerically                                ...done
 --------------
 Nucleus   0He:
 --------------

Diamagnetic contribution to the shielding tensor (ppm) : 
            59.949          0.000        -0.000
             0.000         59.949         0.000
            -0.000          0.000        59.949

Paramagnetic contribution to the shielding tensor (ppm): 
            -0.000          0.000         0.000
             0.000         -0.000        -0.000
             0.000         -0.000        -0.000

Total shielding tensor (ppm): 
            59.949          0.000         0.000
             0.000         59.949        -0.000
             0.000         -0.000        59.949


 Diagonalized sT*s matrix:
 
 sDSO            59.949           59.949           59.949  iso=      59.949
 sPSO            -0.000           -0.000           -0.000  iso=      -0.000
        ---------------  ---------------  ---------------
 Total           59.949           59.949           59.949  iso=      59.949



--------------------------
CHEMICAL SHIELDING SUMMARY (ppm)
--------------------------


  Nucleus  Element    Isotropic     Anisotropy
[...]

PR #427, opened 2017-08-23
Parses NMR for ORCA.

jevandezande:
Reference implementation of NMR parsing for issue #125 about how to design the NMR attribute.

TODO
----
- [x] Write tests


langner:
@ATenderholt @berquist I am curious about your opinion here, since I don't have much experience with NMR. Is it reasonable to assume isotropic and anisotropic shifts will always be calculated?

langner:
I'm going to push this one back to v1.5.3 right away.

berquist:
> Is it reasonable to assume isotropic and anisotropic shifts will always be calculated?

Yes, since they are trivial to calculate from the full 3 by 3 tensor. The most important thing is that the full tensor is parsed.

If we're going to use the tuple approach, then the full tensor should be included, but maybe as the last element of the 3; I need to check that all programs actually print the full tensor.

langner:
> > Is it reasonable to assume isotropic and anisotropic shifts will always be calculated?
>
> Yes, since they are trivial to calculate from the full 3 by 3 tensor. The most important thing is that the full tensor is parsed.
> 
> If we're going to use the tuple approach, then the full tensor should be included, but maybe as the last element of the 3; I need to check that all programs actually print the full tensor.

In that case I would prefer to just parse the full tensor and provide a method to extract the other things, if it is trivial.

langner:
@jevandezande Do you think you could modify you code to parse just the full tensor? We can add some additional methods/properties for extracting the parts you're interested in.

langner:
I think it's the full tensor now? Changes need to be reworked after recent refactoring. Also, ideally we'd have two PRs:
1. Adding the attribute itself (no parsing) as a first step, so we get that part down.
2. Adding the implementation for each particular program.

langner:
> I think it's the full tensor now? Changes need to be reworked after recent refactoring. Also, ideally we'd have two PRs:
> 
> 1. Adding the attribute itself (no parsing) as a first step, so we get that part down.
> 2. Adding the implementation for each particular program.

@berquist Do you think this is a good summary of what needs to be done here?

langner:
Talked this over and we should have merged in. @berquist wanted to take a look first.

langner (approved):

berquist (approved):
I think this is a fine start, and any real changes can wait for cclib 2.0.

I've made another branch that rebases this on top of current master and has a fix for some metadata issues with the NMR output.

berquist:
Closing in favor of #1002.

PR #429, opened 2017-08-25
Bump v1.5.1 to v1.5.2 and update changelog

langner:
~~DO NOT SUBMIT~~

~~We will merge this in when we are ready with all the issues assigned to v1.5.2 are either closed or moved over to the next milestone.~~

This should be good to go.

### DIFF
--- a/ANNOUNCE
+++ b/ANNOUNCE
@@ -1,4 +1,4 @@
-On behalf of the cclib development team, we are pleased to announce the release of cclib 1.5.1, which is now available for download from http://cclib.github.io. This is a minor update to version 1.5 that includes some new functionality and attributes, as well as bug fixes and small improvements.
+On behalf of the cclib development team, we are pleased to announce the release of cclib 1.5.2, which is now available for download from http://cclib.github.io. This is a minor update to version 1.5 that includes some new functionality and attributes, as well as bug fixes and small improvements.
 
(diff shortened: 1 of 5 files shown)


ANNOUNCE
```diff
@@ -37,7 +37,7 @@ The cclib development team
 ———
 
 Summary of changes since last version:
-    New attribute polarizabilities for static or dynamic dipole polarizability
-    New attribute pressure for thermochemistry
-    Detecting closed shells in parsed data
-    Support for Python 3.6
+    * Support for writing Molden and WFX files
+    * Support for thermochesmitry attributes in ORCA
+    * Support for chelpg atomic charges in ORCA (Richard Gowers)
```
TDI-tenderholt: Do you want to make this more consistent with the other points? Suggest either removing (Richard Gowers) or adding the other contributors.

CHANGELOG
```diff
@@ -1,3 +1,20 @@
+Changes in cclib-1.5.2:
+
+Features:
+    * Support for writing Molden and WFX files (Gaur Sagar)
```
TDI-tenderholt: Shouldn't it be Sagar Gaur?

ANNOUNCE
langner: Removed.

CHANGELOG
langner: Done.

Issue #289, opened 2016-07-20
Support trajectories (time)


langner:
In #283 Ramon added support for BOMD trajectories in Gaussian, and there is a regression test for that coming with cclib/cclib-data#36. It would be great to also add unit tests for some standard MD method that multiple programs have. I don't know if that would be BOMD?

berquist:
Yes, BOMD as opposed to ELMD, CPMD, ADMP, or anything else similar to Car-Parrinello. BOMD is much simpler theoretically, therefore it should have fewer knobs to turn in each program, and there are less likely to be implementation differences between each program.

We can run BOMD calculations for at least Gaussian, NWChem, ORCA, and Q-Chem.

langner:
Do you think you could run the jobs? I can do most of the parser code.

berquist:
Yes, I'll run the NWChem, ORCA, and Q-Chem jobs as duplicates of the Gaussian calculation that's been added to the regression files.

langner:
Maybe something with DVB might fit better into our current unit tests.

berquist:
Oh, it's DVB with B3LYP/STO-3G, so same as the unit tests.

langner:
Ah cool, didn't even notice. Are the starting coordinates the same?

berquist:
Based on looking at the standard orientation and rotational constants, yes.

langner:
Great!

berquist:
Should these go in `regressions` or will they have unit tests?

langner:
I'd put them in as unit tests, since we don't have other tests that cover `time`. However, is it possible to trim down on the size compared to the regression test Ramon added?

berquist:
Yes, I'll lower the print level.

langner:
@berquist did you ever get around to running these?

berquist:
It looks like I have ORCA, Q-Chem, NWChem, and perhaps GAMESS.

berquist:
It's pretty clear that I can have this in and tested almost immediately, but my tests are failing because the number of steps/time points aren't consistent with the number of SCF energies and geometries. Except for Gaussian, there is always 1 more SCF energy than time point, because first the energy is converged and the gradient is calculated to obtain the initial forces. Well, it looks like Gaussian is doing the same thing, but the 1 fewer may come from the rejected step elimination, I can't quickly tell. The other programs do traditional MD.

Some programs might not even print intermediate geometries (NWChem), but I'll look to see if I can turn the print level up.

[...]

PR #431, opened 2017-08-26
Support and tests for parsing time from BOMD trajectories

berquist:
Adds further support for the `time` attribute, as outlined in https://github.com/cclib/cclib/issues/289.

* [ ] Ready to go


langner:
I think it looks good to go once the unit tests are sorted out. Do we want to call the logfiles `dvb_md` instead of `dvb_bomd`, in case some program doesn't have BOMD?

berquist:
I'm not sure how far you want to get into the weeds here, but some programs will do purely classical MD, some will do QM/MM MD, and some will do CPMD. They shouldn't look very different, but what we choose to have unit tests for in the future might require something more specific than `md`.

langner:
> I'm not sure how far you want to get into the weeds here, but some programs will do purely classical MD, some will do QM/MM MD, and some will do CPMD. They shouldn't look very different, but what we choose to have unit tests for in the future might require something more specific than md.

I'm sure I understand your point. I'm just suggesting we name the file `dvb_md` but use whatever simplest method each program implements (but BOMD would be preferred).

berquist:
Ok, sounds good.

berquist:
Quick note that it is now impossible to read trajectories from ORCA 4.0; you can only print coordinates to trajectory-style XYZ files.

langner:
Given the ongoing changes and merge conflicts with HEAD, I'm pushing this back to v1.5.3.

langner:
> Quick note that it is now impossible to read trajectories from ORCA 4.0; you can only print coordinates to trajectory-style XYZ files.

OK, can we include this XYZ file with the outputs? We do support multifile outputs, so we can support this.

berquist:
Good point. I'll get to that and try GAMESS within the next few days.

berquist:
Reading from the main output (just includes times), the trajectory XYZ file, and another separate file where each SCF iteration is printed works for all important attributes except `atomcoords`. This is because

```
ORCA log file ['dvb_bomd.out', 'dvb_bomd-postrj.xyz', 'dvb_bomd.scf.log']
Identified logfile to be in ORCA format
```

which isn't entirely true. Is it possible to modify ccio to have fallback work for one file and be added to the combined `ccdata` correctly?

```python
from cclib.io import ccopen, ccread

filenames = [
    'dvb_bomd.out',
    'dvb_bomd-postrj.xyz',
    'dvb_bomd.scf.log',
]

# job = ccopen(filenames)
# print(job)
# data = job.parse()
data = ccread(filenames, verbose=True)

print('\n'.join(dir(data)))

import openbabel as ob
import pybel as pb
assert hasattr(data, 'atomcoords')
```

berquist:
My current approach will be:

1. at the end of `ccopen`, if `is_listofstrings`, see if any can be opened by a fallback mechanism and parse them, returning a list of `ccdata`s
2. perform the "main" parse with `filetype(source, ...)` on the combined source
3. add all the `ccdata`s together by combining their attributes, starting with the fallback results so ours take priority, because the addition is not commutative

This could result in duplicate parsing, since there is overlap between the list of extensions provided by `pybel.informats.keys()` and our own classes.

langner:
It's hard to grok this PR... there are a lot of changes and now conflicts with master. Could you separate this into a sequence of changes?

Ideally a PR should contain a single self-contained change.

berquist:
This is now confusing. Cherry-picking the simplest things into a new PR.

PR #433, opened 2017-09-06
Move regression test code back to main repo

langner:
This is implementing cclib/cclib-data#72.

PR #434, opened 2017-09-06
Q-Chem: updated printing in RPA/CIS output

berquist:
This needs to go in before https://github.com/cclib/cclib-data/pull/74.


langner:
If it wouldn't be too much trouble, let's merge in #433 first and include the regression test alongside the fix here.

langner:
Great. You just need to remove the file from cclib/cclib-data#74 now.

berquist:
Done.

PR #439, opened 2017-10-05
Setupscript + gradient parsing for Molpro + Pep8 changes

mcocdawc:
1. Changes in setupscript:

Previously it was not possible to install the module in development mode or use pip for installing. Now you can do ``pip install -e .``
I hope there were no special reasons for sticking to distutils.

2. Gradient parsing for Molpro implemented.

Unit is eV / Angstrom.

3. PEP 8

I made some changes to make the molproparser a bit more PEP8 compatible.
For readability reasons I changed:

```
list(map(f, A))
to
[f(x) for x in A]
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,10 @@ chemistry log files. It also provides a platform to implement
 algorithms in a package-independent manner.
 """
 
-import sys
+from __future__ import with_statement
+from __future__ import absolute_import
+from setuptools import setup, find_packages
+from io import open  # pylint:disable=redefined-builtin
 
 
 # Chosen from http://www.python.org/pypi?:action=list_classifiers
@@ -28,46 +31,34 @@ Topic :: Scientific/Engineering :: Chemistry
 Topic :: Software Development :: Libraries :: Python Modules"""
 
 
+def readme():
+    '''Return the contents of the README.md file.'''
+    with open('README.md') as freadme:
+        return freadme.read()
+
+
 def setup_cclib():
-
-    # Import from setuptools only if requested.
-    if 'egg' in sys.argv:
-        sys.argv.pop(sys.argv.index('egg'))
-        from setuptools import setup
-
-    from distutils.core import setup
-
-    # The list of packages to be installed.
-    cclib_packages = [
-        'cclib',
-        'cclib.bridge',
-        'cclib.io',
-        'cclib.method',
-        'cclib.parser',
-        'cclib.progress',
-    ]
 
     doclines = __doc__.split("\n")
 
     setup(
-        name = "cclib",
-        version = "1.5.2",
-        url = "http://cclib.github.io/",
-        author = "cclib development team",
-        author_email = "cclib-users@lists.sourceforge.net",
-        maintainer = "cclib development team",
-        maintainer_email = "cclib-users@lists.sourceforge.net",
-        license = "BSD 3-Clause License",
-        description = doclines[0],
-        long_description = "\n".join(doclines[2:]),
-        classifiers = classifiers.split("\n"),
-        platforms = ["Any."],
-        packages = cclib_packages,
-        package_dir = { 'cclib':'src/cclib' },
-        scripts = ["src/scripts/ccget", "src/scripts/ccwrite", "src/scripts/cda"],
+        name="cclib",
+        version="1.5.2",
+        url="http://cclib.github.io/",
+        author="cclib development team",
+        author_email="cclib-users@lists.sourceforge.net",
+        maintainer="cclib development team",
+        maintainer_email="cclib-users@lists.sourceforge.net",
+        license="BSD 3-Clause License",
+        description=doclines[0],
+        long_description="\n".join(doclines[2:]),
+        classifiers=classifiers.split("\n"),
+        platforms=["Any."],
+        packages=find_packages('src'),
+        package_dir={'': 'src'},
+        scripts=["src/scripts/ccget", "src/scripts/ccwrite", "src/scripts/cda"]
     )
 
 
 if __name__ == '__main__':
-
     setup_cclib()
--- a/src/cclib/parser/molproparser.py
+++ b/src/cclib/parser/molproparser.py
@@ -10,6 +10,7 @@
 
 import csv
 import itertools
+from itertools import combinations_with_replacement
 
 import numpy
 
@@ -40,9 +41,8 @@ def create_atomic_orbital_names(orbitals):
     for i, orb in enumerate(orbitals):
 
         # Cartesian can be generated directly by combinations.
-        cartesian = [
-            ''.join(x) for x in
-            itertools.combinations_with_replacement(['x', 'y', 'z'], i + 2)]
+        cartesian = [''.join(x) for x in
+                     combinations_with_replacement(['x', 'y', 'z'], i + 2)]
 
         # For spherical functions, we need to construct the names.
         pre = str(i + 3) + orb.lower()
@@ -196,8 +196,7 @@ class Molpro(logfileparser.Logfile):
                             aonum += 1
                         else:
                             functype = s
-                            element = self.table.element[
-                                self.atomnos[atomno - 1]]
+                            element = self.table.element[self.atomnos[atomno - 1]]
                             aoname = "%s%i_%s" % (element, atomno, functype)
                             aonames.append(aoname)
                     line = next(inputfile)
@@ -249,7 +248,7 @@ class Molpro(logfileparser.Logfile):
             mocoeffs.append(coeffs)
 
             # The loop should keep going until there is a double blank line,
-            # and there is # a single line between each coefficient block.
+            # and there is a single line between each coefficient block.
             line = next(inputfile)
             if not line.strip():
                 line = next(inputfile)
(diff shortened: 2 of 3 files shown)


src/cclib/parser/molproparser.py
```diff
@@ -222,18 +248,20 @@ def _parse_orbitals(self, inputfile, line):
                 line = next(inputfile)
             mocoeffs.append(coeffs)
 
-            # The loop should keep going until there is a double blank line, and there is
-            # a single line between each coefficient block.
+            # The loop should keep going until there is a double blank line,
+            # and there is # a single line between each coefficient block.
```
berquist: duplicate #

src/cclib/parser/molproparser.py
```diff
@@ -282,7 +313,9 @@ def extract(self, inputfile, line):
             line = next(inputfile)
             while line.strip():
                 temp = line.strip().split()
-                atomcoords.append([utils.convertor(float(x), "bohr", "Angstrom") for x in temp[3:6]])  # bohrs to angs
+                atomcoords.append(
+                    [utils.convertor(float(x), "bohr", "Angstrom")
+                     for x in temp[3:6]])
```
berquist: I think it would have been sufficient to not break on the first append, and just break inside the list comprehension. Dangling open parens for method calls look funny.

src/cclib/parser/molproparser.py
```diff
@@ -222,18 +248,20 @@ def _parse_orbitals(self, inputfile, line):
...
-        # If symmetry was used (offset was needed) then we will need to pad all MO vectors
+        # If symmetry was used (offset was needed)
+        # then we will need to pad all MO vectors
         # up to nbasis for all irreps before the last one.
         if offset > 0:
             for im, m in enumerate(mocoeffs):
                 if len(m) < self.nbasis:
-                    mocoeffs[im] = m + [0.0 for i in range(self.nbasis - len(m))]
+                    mocoeffs[im] = (
+                        m + [0.0 for i in range(self.nbasis - len(m))])
```
berquist: I'm not sure this is correct. Doesn't this turn it into a list inside of a tuple? You should also never leave just one dangling delimiter. Either leave it as-is, or insert a newline after the last one as well. The original was easier to read.

src/cclib/parser/molproparser.py
```diff
@@ -154,28 +170,34 @@ def _parse_orbitals(self, inputfile, line):
...
                         if s.isdigit():
                             atomno = int(s)
-                            atombasis[atomno-1].append(aonum)
+                            atombasis[atomno - 1].append(aonum)
                             aonum += 1
                         else:
                             functype = s
-                            element = self.table.element[self.atomnos[atomno-1]]
+                            element = self.table.element[
+                                self.atomnos[atomno - 1]]
```
berquist: See later dangling delimiter comment; this was fine as it was.

src/cclib/parser/molproparser.py
```diff
@@ -28,20 +29,25 @@ def create_atomic_orbital_names(orbitals):
...
+    # and the # spherical ones are called 3d0, 3d1-, 3d1+, 3d2- and 3d2+.
+    # For F orbitals, the Cartesians are xxx, xxy, xxz, xyy, ...
+    # and the sphericals are 4f0, 4f1-, 4f+ and so on.
     for i, orb in enumerate(orbitals):
 
         # Cartesian can be generated directly by combinations.
-        cartesian = list(map(''.join, list(itertools.combinations_with_replacement(['x', 'y', 'z'], i+2))))
+        cartesian = [
+            ''.join(x) for x in
+            itertools.combinations_with_replacement(['x', 'y', 'z'], i + 2)]
```
berquist: I'm ok with removing `map`, but see the later dangling delimiter comment.

One additional way it could be shortened is to stick a `from itertools import combinations_with_replacement` at the top of the file.

setup.py
```diff
@@ -12,7 +12,10 @@
 algorithms in a package-independent manner.
 """
 
-import sys
+from __future__ import with_statement
+from __future__ import absolute_import
+from setuptools import setup, find_packages
+from io import open  # pylint:disable=redefined-builtin
```
berquist: I don't think this is correct. Wouldn't this import `cclib.io`? Nothing should need to be imported to read in a plain text file.

src/cclib/parser/molproparser.py
```diff
@@ -916,6 +916,29 @@ def extract(self, inputfile, line):
...
+            lines_read = 0
+            while lines_read < self.natom:
+                line = next(reader)
+                if line:
+                    grad.append([convert(float(x)) for x in line[1:]])
+                    lines_read += 1
+            try:
+                self.grads.append(grad)
+            except AttributeError:
+                self.grads = [grad]
```
berquist: We usually try to avoid try/except when setting attributes, and prefer

```python
if not has_attribute(self, 'grads'):
    self.grads = []
self.grads.append(grad)
```

src/cclib/parser/molproparser.py
```diff
@@ -916,6 +916,29 @@ def extract(self, inputfile, line):
                 self.atomcharges = {}
             self.atomcharges['mulliken'] = charges
 
+        if 'GRADIENT FOR STATE' in line:
+            reader = csv.reader(inputfile, delimiter=' ',
```
berquist: The CSV reader can be avoided entirely. `line = next(inputfile)` or `next(inputfile)` is fine. For tokenization, see below.

src/cclib/parser/molproparser.py
```diff
@@ -916,6 +916,29 @@ def extract(self, inputfile, line):
...
+            def convert(x):
+                return (utils.convertor(x, 'hartree', 'eV')
+                        / utils.convertor(1, 'bohr', 'Angstrom'))
+
+            grad = []
+            lines_read = 0
+            while lines_read < self.natom:
+                line = next(reader)
+                if line:
+                    grad.append([convert(float(x)) for x in line[1:]])
```
berquist: This would become

```python
grad.append([convert(float(x)) for x in line.split()[1:]])
```

src/cclib/parser/molproparser.py
mcocdawc: Yes you are right, that it can be avoided entirely in this case. Since most outputfiles are a mix of structured and unstructured data, I like to explicitly use a function that assumes structure, when parsing a table. Just for the human.

src/cclib/parser/molproparser.py
mcocdawc: Changed accordingly

src/cclib/parser/molproparser.py
mcocdawc: Changed accordingly

setup.py
mcocdawc: No, this is to get a consistent open function in python 2.7 and 3.x.
AFAIK the import manager does not recurse through directories to find a module matching the name to import and your io module is several layers deeper than the setup.py script.

http://python-future.org/compatible_idioms.html
```
# Python 2 and 3: alternative 1
from io import open
f = open('myfile.txt', 'rb')
data = f.read()              # as bytes
text = data.decode('utf-8')  # unicode, not bytes
```

src/cclib/parser/molproparser.py
mcocdawc: It then exceeds 80 characters

src/cclib/parser/molproparser.py
mcocdawc: Commas not parentheses make a tuple.

mcocdawc:
1. PEP8:

I wanted to open two files in splitscreen which did not work really well because currently there are too many lines exceeding 80 characters.

So yes I think there is an inherent advantage in strictly adhering to the 80 character limit:

And for example this line here:
```
-                            element = self.table.element[self.atomnos[atomno-1]]
# vs
+                            element = self.table.element[
+                                self.atomnos[atomno - 1]]
```
If it is not broken into two lines, it does use more than 80 characters.
So are you sure, that you want this?

Regarding the dangling delimiters: I will change that.

2. Editor

I use atom, with the python-ide, python-indent and automatic linter packages
Now comes the perhaps embarassing part for me:
I inserted newlines manually.


3. Units:

In the scfenergies part of your documentation I found:
[...] An array containing the converged SCF energies of the calculation, in eV. [...].

After longer searching I could find no definitive unit for the forces in the documentation. So I assumed, that the energy unit would be the same as for the SCF energies. Getting gradients in hartree / bohr is a bit inconsistent to the other energies.
So do you still want hartree / bohr? In any way the documentation should be more explicit.

src/cclib/parser/molproparser.py
mcocdawc: Again it exceeds then 80 characters

src/cclib/parser/molproparser.py
mcocdawc: Sorry :+1: :(

langner:
Thanks for all the improvements! It would be really great if you could break this up into three separate PRs. Would that be possible?

mcocdawc:
As far as I understand GIT, I don't think that this is possible. (At least not while keeping a clean logical snapshot history). :(

ATenderholt:
I think what Karol is suggestion is that you create three branches off of your `master`. Then in each branch (NOT `master`), you can do something like `git rebase -i HEAD~7` and select the appropriate commits.

(Another option is create branches off a specific commit such as the parent of `f6ad8109a3714cad8f30076707789a7dba83b1d5`, and then use `git cherry-pick`).

mcocdawc:
I am more [this kind of user](https://xkcd.com/1597/)  ;) but I will try my best

mcocdawc:
Due to breaking the PR into separate parts, I close this one here

langner:
Thanks Oskar, I really appreciate you taking the time to do this in smaller increments.

PR #441, opened 2017-10-14
Changed setup script

mcocdawc:
Previously it was not possible to install the module in development mode or use pip for installing. Now you can do e.g. ``pip install -e .``
I hope there were no special reasons for sticking to distutils.

Breaks pull request #439 into three parts.


setup.py
```diff
@@ -28,46 +31,34 @@
 Topic :: Software Development :: Libraries :: Python Modules"""
 
 
-def setup_cclib():
-
-    # Import from setuptools only if requested.
-    if 'egg' in sys.argv:
-        sys.argv.pop(sys.argv.index('egg'))
-        from setuptools import setup
+def readme():
```
berquist: I don't see this function being called anywhere in the file or the rest of the codebase. Is it for PyPI? If so, can you add a comment? Otherwise, it seems unnecessary.

setup.py
```diff
@@ -12,7 +12,10 @@
 algorithms in a package-independent manner.
 """
 
-import sys
+from __future__ import with_statement
+from __future__ import absolute_import
+from setuptools import setup, find_packages
```
langner: Please just import setuptools and use it a s a module.

setup.py
mcocdawc: Oh yes, it is unnecessary. Sorry.
In the projects I saw, it was always used like this:
```
long_description=readme()
```
you have the approach with the ``__doc__`` variable, which looks cleaner.

langner:
Thanks for this!

PR #442, opened 2017-10-14
Molpro grads

mcocdawc:
Gradient parsing for Molpro implemented.

Unit is currently Hartree / Bohr.

In the scfenergies part of your documentation I found:
[...] An array containing the converged SCF energies of the calculation, in eV. [...].

After longer searching I could find no definitive unit for the forces in the documentation. Getting gradients in hartree / bohr is a bit inconsistent to the other energies in eV.
Using atomic units makes IMHO only sense, if the complete fraction consists only of atomic units.
This is why I would use eV / Ångström.

Anyway the documentation should be more explicit about the units in the gradients.

Breaks #439 into three parts.


berquist:
I agree that it's confusing, partially because it isn't implemented for most programs, and our documentation is currently very out of date. I would also argue that having energies in eV is the real mistake, but it's technical debt we have to live with until at least version 2.0.

There are also failures during the tests, but 0 is getting returned by Travis. See https://github.com/cclib/cclib/issues/444 for the exact problem.

mcocdawc:
Sorry for not changing and testing yet. I finish my master thesis in less than two weeks. Than I am back again. ;)

langner:
Good luck!

mcocdawc:
OK, I have looked into the test details, but cannot find the reason, why it fails.

It says, that ``self.natom`` is not defined, although other branches of the conditional in the same method assume the existence of ``self.natom``. Besides I could not reproduce the bug at home with random molpro output files. 
Is it possible to see the exact molpro output file where the test fails?

berquist:
I cannot reproduce it either. The test is `test/io/testccio.py/ccopenTest/test_list_of_filenames`. It may have something to do with passing a list of files where `natom` isn't in one of them. Though, files are parsed in the order given, so if it's in the `log` file but not the `out` file, it shouldn't matter.

mcocdawc:
Sorry to ask, but how should I proceed then?

langner:
It seems to me that something is not working correctly in ccopen for lists of strings. Let me take a look...

langner:
Note that it seems `natom` is in the out file and not in the log file (which has `GRADIENT FOR STATE`), but the order in the `test_list_of_filenames` is `ilenames = ["dvb_gopt.log", "dvb_gopt.out"]`.

langner:
I can reproduce it now. It seems that after #441 you might need to remove the installed library manually, because previous installations override the egg.

langner:
Switching the order of logfiles fixes the issue. I filed a PR against your branch @mcocdawc: https://github.com/mcocdawc/cclib/pull/1

mcocdawc:
Thanks for the PR.

Regarding the manual reinstallation: You can just install it via ``pip install -e`` to obtain development mode.
(If I understand your comment correctly)

mcocdawc:
Just to avoid a misunderstanding. I don't know if I should do any more work for this pull request.

berquist:
Thanks for the poke. I want to make sure the units are in hartree/bohr first, so I'll need to run some calculations and get back to you soon.

mcocdawc:
aye aye

berquist:
Ok, it's definitely in hartree/bohr. Can you apply the patch I've just sent you?

berquist:
Great! Thanks for being patient.

mcocdawc:
I think I let you wait even longer in the beginning. ;)

PR #443, opened 2017-10-14
PEP8 changes

mcocdawc:
Making molproparser more PEP8 compatible.
Breaking #439  into three parts.
Due to length I won't repeat the discussion from #439 here.
The two main questions were 
1. Dangling delimiters:
2. 80 character line wrapping:

Before putting more work into it, I need a decision about the style to follow.
The possibilities are:
```python
                    # Note that this exceeds 80 characters
                    mocoeffs[im] = m + [0.0 for i in range(self.nbasis - len(m))]

                    #  vs

                    mocoeffs[im] = (
                        m + [0.0 for i in range(self.nbasis - len(m))])

                    #  vs

                    mocoeffs[im] = (
                        m + [0.0 for i in range(self.nbasis - len(m))]
                    )
```
I am strongly in favor of the second version. 
The first possibility uses more than 80 characters and is pretty annoying when using vimdiff etc.
So I definetely wouldn't use it.

Regarding the last version: The examples in [PEP8](https://www.python.org/dev/peps/pep-0008)
don't have the closing bracket in a new line and they put for sure a lot of thinking into it.
The last version is familiar for people coming from K&R C-style, but uses a lot of vertical space.
Possible errors that should be prevented by this style cannot happen in python by construction.
(Difference between indentation and actual control flow. Example below)
With modern IDEs, highlighting matching parentheses, and the immediate SyntaxError the probability to miss a closing bracket is extremely low.
A last argument comes from implicit parentheses, as can be seen [here](https://en.wikipedia.org/wiki/Indentation_style#Lisp_style)

Anyway: Your project, your rules. 

```C
for (int i = 0; i < 10; i++)
    whoops(bar);   /* repeated 10 times, with i from 0 to 9 */
{
    only_once();   /* Programmer intended this to be done 10 times */
} //for (i) <-- This comment is no longer valid, and is very misleading!
```


langner:
> Before putting more work into it, I need a decision about the style to follow.
> The possibilities are:
> 
>                     # Note that this exceeds 80 characters
>                     mocoeffs[im] = m + [0.0 for i in range(self.nbasis - len(m))]
> 
>                     #  vs
> 
>                     mocoeffs[im] = (
>                         m + [0.0 for i in range(self.nbasis - len(m))])
> 
>                     #  vs
> 
>                     mocoeffs[im] = (
>                         m + [0.0 for i in range(self.nbasis - len(m))]
>                     )
> I am strongly in favor of the second version.
> The first possibility uses more than 80 characters and is pretty annoying when using vimdiff etc.
> So I definetely wouldn't use it.

We haven't followed the 80 character rule diligently, and I don't see why should start to. Of course, it's nice to keep lines at a reasonable length. In this case, another option is to break up the logic into two lines, for example:

```python
n = self.nbasis = len(m)
mocoeffs[im] = m + [0.0 for i in range(n)]
```

berquist:
I have used both the former (inside function calls) and the latter (for lists of strings). For all the examples in this PR, I would leave them as-is. I think some parts of the codebase are unreadable, but not because they don't follow this particular PEP 8 convention. Additionally, PEP 8 is a suggestion, not a rule, and in many places this would hurt readability.

My vote is no, because we should focus more on other aspects of readability, like cleaning up logic and better naming.

langner:
So, it's agreed we're not enthusiastic about making cosmetic changes to existing code. Another reason to not do this is that it will obscure the history of the code more.

Like Eric said, the code is not readable due to PEP8 violations. The parser would need serious refactoring to improve readability.

mcocdawc:
aye aye. Sorry for being away for so long.

Issue #438, opened 2017-09-20
QChem: charge being set to number of ECP-replaced electrons


theochemtheo:
Hi,

Qchem provides a built in version of [libwfa](https://github.com/libwfa/libwfa), a tool for analyzing calculations, primarily excited states. When libwfa is activated (`state_analysis true` in the `$rem` block), an addition segment can be found in the output file containing the analysis. I think this extra section is causing parsing issues, since it contains some lines that are similar to those used to find excited state energies.

I have attached two example files, one without libwfa activated and one with libwfa.

[water_qchem500.txt](https://github.com/cclib/cclib/files/1317220/water_qchem500.txt)
[water_qchem500+libwfa.txt](https://github.com/cclib/cclib/files/1317221/water_qchem500.libwfa.txt)

The first one parses fine and the relevant excited state properties can be obtained. The second one does not, although ground state properties are unaffected.

Thanks,

Theo

berquist:
Hi Theo, thanks for the bug report and especially the examples. I'm not having any trouble parsing these, and I bet it has to do with the version of cclib you're using. Can you check it? If you're using a version from a package manager like apt-get and not PyPI (`pip`) or the Git repo, it can be really out-of-date.

```
[eric@copper]$ python
Python 3.6.2 (default, Jul 20 2017, 03:52:27) 
[GCC 7.1.1 20170630] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cclib
>>> cclib.__version__
'1.5.2'
```

---

Here is the output I get from the command line script:

```
[eric@copper]$ ccget --verbose --list ~/Dropbox/research/cclib/438/*out
Attempting to read /home/eric/Dropbox/research/cclib/438/water_qchem500.libwfa.out
Identified logfile to be in QChem format
cclib can parse the following attributes from /home/eric/Dropbox/research/cclib/438/water_qchem500.libwfa.out:
  atomcoords
  atomnos
  charge
  coreelectrons
  etenergies
  etoscs
  etsecs
  etsyms
  homos
  metadata
  moenergies
  moments
  mult
  natom
  nbasis
  nmo
  scfenergies
  scftargets
  scfvalues
Attempting to read /home/eric/Dropbox/research/cclib/438/water_qchem500.out
Identified logfile to be in QChem format
cclib can parse the following attributes from /home/eric/Dropbox/research/cclib/438/water_qchem500.out:
  atomcharges
  atomcoords
  atomnos
  charge
  coreelectrons
[...]

PR #446, opened 2017-10-30
QChem: `coreelectrons` and charges when using ECPs

berquist:
Fixes https://github.com/cclib/cclib/issues/438.

The regression files are located in https://github.com/cclib/cclib-data/pull/75.


src/cclib/parser/qchemparser.py
```diff
@@ -159,6 +159,82 @@ def after_parsing(self):
...
+"iprint >= 100", otherwise coreelectrons may only be partially correct."""
+                    self.logger.warning(msg)
+                    for entry in self.user_input['ecp']:
+                        element, index, ncore = entry
+                        if ncore == 0:
+                            if index == -1:
+                                for i in range(self.natom):
+                                    if elements[i] == element:
+                                        msg = "Can't calculate coreelectrons for {} {}".format(element, i + 1)
+                                        self.logger.warning
```
ATenderholt: Missing warning message?

src/cclib/parser/qchemparser.py
```diff
@@ -309,45 +385,142 @@ def extract(self, inputfile, line):
...
+                        else:
+                            charge, mult = [int(x) for x in line.split()]
+                            self.user_input['molecule']['charge'] = charge
+                            self.user_input['molecule']['mult'] = mult
+
+                    line = next(inputfile).lower()
 
             # Parse the basis set name
             if 'Requested basis set' in line:
+                # if 'is non-standard' not in line
```
ATenderholt: Commented-out code that should be removed?

test/data/testCore.py
```diff
@@ -32,6 +34,12 @@ def testcorrect(self):
         ans = numpy.array(ans, "i")
         numpy.testing.assert_array_equal(self.data.coreelectrons, ans)
 
+    @skipForParser('ADF', '?')
```
ATenderholt: Is there a more descriptive reason to skip for ADF?

src/cclib/parser/qchemparser.py
```diff
@@ -159,6 +159,82 @@ def after_parsing(self):
...
+                    msg = """ECPs are present, but the number of core \
+electrons isn't printed at all. Rerun with "iprint >= 100", otherwise \
+coreelectrons is incorrect (all zeros)."""
+                    self.logger.warning(msg)
+                elif ecp_is_gen and not has_iprint:
+                    msg = """ECPs are present, but the number of core \
+electrons are only printed for user-provided definitions. Rerun with \
+"iprint >= 100", otherwise coreelectrons may only be partially correct."""
+                    self.logger.warning(msg)
+                    for entry in self.user_input['ecp']:
```
langner: How bad is the "partial correctness" here? Do we really want to parse if it's going to be incorrect?

src/cclib/parser/qchemparser.py
berquist: A fix for another day...

src/cclib/parser/qchemparser.py
berquist: Here is an example.

The first case. There is no ECP info in the output, but the number of core electrons can be read from the user input since the only ECP is user-defined (the 10 in `Stevens_ECP   2   10`).

```
$rem
method = hf
basis = gen
ecp = gen
n_frozen_core = fc
ecp_disable_reconstruction = true
$end

$molecule
 0 1
 Al
 H1  Al  r
 H2  Al  r  H1  120.
 F3  Al  r  H1  120.  H2  180.

 r = 1.6
$end

$ecp
 AL
   Stevens_ECP   2   10
   d potential
      1
      1       1.95559     -3.03055
   s-d potential
      2
      0       7.78858      6.04650
      2       1.99025     18.87509
   p-d potential
      2
      0       2.83146      3.29465
      2       1.38479      6.87029
   ****
$end

$basis
 AL
   SP  3  1.00
      0.9011     -0.30377    -0.07929
      0.4495      0.13382     0.16540
      0.1405      0.76037     0.53015
   SP  1  1.00
      0.04874     0.32232     0.47724
   ****
 H
   3-21G
   ****
 F
   3-21G
   ****
$end

```

The second case. Replacing the basis set on fluorine with a built-in ECP means there's no need to enter the number of core electrons, so it can't be read.

```
$rem
method = hf
basis = gen
ecp = gen
n_frozen_core = fc
ecp_disable_reconstruction = true
$end

$molecule
 0 1
 Al
 H1  Al  r
 H2  Al  r  H1  120.
 F3  Al  r  H1  120.  H2  180.

 r = 1.6
$end

$ecp
 AL
   Stevens_ECP   2   10
   d potential
      1
      1       1.95559     -3.03055
   s-d potential
      2
      0       7.78858      6.04650
      2       1.99025     18.87509
   p-d potential
      2
      0       2.83146      3.29465
      2       1.38479      6.87029
   ****
F
sbkjc
   ****
$end

$basis
 AL
   SP  3  1.00
      0.9011     -0.30377    -0.07929
      0.4495      0.13382     0.16540
      0.1405      0.76037     0.53015
   SP  1  1.00
      0.04874     0.32232     0.47724
   ****
 H
   3-21G
   ****
 F
   sbkjc
   ****
$end
```

cclib output:

```
[eric@copper]$ ccget --verbose coreelectrons AlH2F_test.out AlH2F_sbkjc.out
Attempting to read AlH2F_test.out
Identified logfile to be in QChem format
[QChem AlH2F_test.out WARNING] Attribute charge changed value (10 -> 0)
[QChem AlH2F_test.out WARNING] ECPs are present, but the number of core electrons are only printed for user-provided definitions. Rerun with "iprint >= 100", otherwise coreelectrons may only be partially correct.
coreelectrons
[10  0  0  0]
Attempting to read AlH2F_sbkjc.out
Identified logfile to be in QChem format
[QChem AlH2F_sbkjc.out WARNING] Attribute charge changed value (12 -> 0)
[QChem AlH2F_sbkjc.out WARNING] ECPs are present, but the number of core electrons are only printed for user-provided definitions. Rerun with "iprint >= 100", otherwise coreelectrons may only be partially correct.
[QChem AlH2F_sbkjc.out WARNING] Can't calculate coreelectrons for F 4
coreelectrons
[10  0  0  0]
```

The warning for the first output can be safely ignored. I was thinking about putting an additional message that this could be ignored for such cases. For the second output, `coreelectrons[3]` is wrong; from the charge we can tell it should be 2. In the general case, this number can't be guessed.

Because of the charge discrepancy, you could always get the _total_ number of core electrons right, it's just the assignment to individual atoms that might be wrong. But we'd need to dump additional information that is only handled internally right now.

> How bad is the "partial correctness" here?

It's not great, because I wonder if people read the warnings and are careful about this sort of thing. The partial incorrectness may also be an edge case, since I don't know how common it is to mix built-in and user-defined ECPs. I'd think one or the other is more common, but I'm not an ECP user. Maybe @theochemtheo could comment on his typical ECP usage.

> Do we really want to parse if it's going to be incorrect?

Probably not. In that case, I would stop `logfileparser` from automatically creating `coreelectrons`.

theochemtheo: @berquist I tend to work with organometallics that only contain one metal, so I only use one ECP at a time, but I'm not sure that would hold for all people.

I guess you could get around this if you had a list of Q-chem's built in ECPs for each atom ([See section 8.7.2 of the Q-chem 5.0 manual](http://www.q-chem.com/qchem-website/manual/qchem50_manual/sect-ECP_derivs.html)) but this is kind of a cumbersome way of doing it.

berquist: It won't happen for now, but I don't think we're going to completely rule out that option. There is interest in being able to parse basis set names from ORCA outputs, which will at minimum require a regex approach. An alternative that may simplify things across parsers would be to keep track of what basis sets/ECPs each package comes with. We're fortunate that this list doesn't change very often.

berquist:
So, is the conclusion that `coreelectrons` should only be set at all if we can guarantee its correctness, otherwise don't set it and print a warning?

langner:
> So, is the conclusion that coreelectrons should only be set at all if we can guarantee its correctness, otherwise don't set it and print a warning?

My preference would be to provide only data we know is correct. That would include the case where `coreelectons` is missing only for one atom and we can infer the value somehow. Otherwise just display message to add `iprint` to the input and move on.

berquist:
Ok, it should be taken care of.

langner:
Looks good to me. I'm going to submit the regression files in cclib/cclib-data#75 to see how that works with this PR. When the files are missing in master, it prints out the function with no corresponding files, I hope it work the other way around as well.

@ATenderholt: any other concerns?

ATenderholt:
The changes are a bit hard to follow, especially with all of the nested if-statements. Can those be refactored out into some descriptively-named methods?

langner:
> The changes are a bit hard to follow, especially with all of the nested if-statements. Can those be refactored out into some descriptively-named methods?

I agree, although for me it's not a blocker.

berquist:
I'll try and do something about it tomorrow. I agree it's pretty bad, I had a hard time writing it.

berquist:
https://github.com/cclib/cclib/pull/446/commits/a8c2d4ab2775cfb0918fe4b009557bbfb35f4a16 is the commit breaking 3.2, but I don't have time ti fix it right now.

berquist:
I've confirmed it fails with Python 3.3/NumPy 1.10.0 and Python 3.4/NumPy 1.10.0, but not Python 3.4/NumPy 1.13.3, so it must be a NumPy problem.

ATenderholt:
I think dropping support for 3.2 is fine. Note that the release notes for NumPpy 1.13.0 says it only supports 2.7 and 3.4 - 3.6.

If you wanted to maintain 3.2 support could you do something like the following? Not as elegant, but I think it should work.

`new_mask = [ x[0] for x in enumerate(mask) if x[1] ]`

berquist:
That's a bit too much of a version jump just for one or two short lines of code. Forgot about the enumerate trick, thanks.

langner:
Squashing and merging this one.

Issue #195, opened 2015-03-25
Parsing multiconfigurational/multireference outputs


berquist:
There is currently no ability to handle properties specific to these kinds of jobs, such as state energies, transition energies, or orbital coefficients.

Parsers that would support some level of multiconfigurational SCF or multireference correlated calculations:
* DALTON
* GAMESS
* GAMESS-UK
* Gaussian
* (Molcas)
* Molpro
* NWChem
* ORCA
* Psi

Parsers that wouldn't:
* ADF
* Jaguar
* Q-Chem

Although the possible programs allow for varying levels of sophistication, the list is sizable.

There is also the matter of choosing appropriate test jobs. I've been working on some good examples for about half of the above list (DALTON, GAMESS, Molcas, ORCA), but in general these are not black-box methods and often results between programs are not comparable, especially for MRCI-type calculations. "Plain" CASSCF should fare a bit better in that regard.

Due to the length of the list and complexity of the calculations, I propose this as a post-v2.0 project.

langner:
If there's a specific thing you'd like to see fast, like state energies or coefficients, I say go for it. I would, however, seprate MCSF and MRCI.

berquist:
> If there's a specific thing you'd like to see fast, like state energies or coefficients, I say go for it.

Maybe, but only once the other outstanding issues for 1.4/2.0 are squared away.

> I would, however, seprate MCSF and MRCI.

This is a good idea and I'll definitely do this. Just wanted to make a "mega-issue" to start with an collect opinions/ideas.

berquist:
If done properly, this should be somewhat dependent on https://github.com/cclib/cclib/issues/258.

jevandezande:
I recently got a request from someone for CASSCF support. I'm willing to write up the ORCA section, but I think we may need to figure out how we want to organize this. To me it would make sense for the CASSCF to be a dictionary including various properties like occupations, CSFs, window size (#electrons, #orbitals), etc. One problem I can forsee is that while ORCA pretends to support symmetry, it doesn't really work that well, while other codes rely on it heavily.

Thoughts?

berquist:
The only think I found odd was that it doesn't even try to tell you what spin an electron in a singly-occupied orbital is like Molcas does, but I'm familiar with the ORCA workflow so it doesn't bother me as much. Basically everything else you can get it to print. At the bare minimum, we need everything from this part of the output:

```
---------------------------------------------
[...]

PR #447, opened 2017-11-08
Orca CASSCF

jevandezande:
There is no standard yet implemented for CASSCF (see #195), so I went ahead and parsed as much as I could of the output. Currently it fails if the computations contains multiple roots due to the fact that CASSCF prints the absorption spectrum differently than other methods.

I imagine the bare minimum of data to save would be:
- Coefficients and orbital occupations
- CASSCF energy
- Energies of each root (absolute energy or relative energy?)
- Orbital energies (and symmetries if available)

I am sure there is a lot more information that others may consider important. Suggestions are encouraged as I don't perform a lot of CASSCF computations.

TODO
----
- [x] Fix bugs in spectra parsing

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -995,13 +995,14 @@ States  Energy Wavelength    D2        m2        Q2         D2+m2+Q2       D2/TO
             #    MO =    0  IRREP= 0 (A)
             #    MO =    1  IRREP= 1 (B)
             vals = next(inputfile).split()
+            # Symmetry section is only printed if symmetry is used
             if vals[0] == 'Symmetry':
-                # Symmetry section is only printed if symmetry is used
                 assert vals[-1] == 'ON'
                 point_group = next(inputfile).split()[-1]
                 used_point_group = next(inputfile).split()[-1]
                 num_irreps = int(next(inputfile).split()[-1])
                 num_active = 0
+                # Parse the irreps
                 for i, line in zip(range(num_irreps), inputfile):
                     reg = r'Irrep\s+(\w+) has\s+(\d+) SALCs \(ofs=\s*(\d+)\) #\(closed\)=\s*(\d+) #\(active\)=\s*(\d+)'
                     groups = re.search(reg, line).groups()
@@ -1009,11 +1010,14 @@ States  Energy Wavelength    D2        m2        Q2         D2+m2+Q2       D2/TO
                     salcs, ofs, closed, active = map(int, groups[1:])
                     num_active += active
                 self.skip_line(inputfile, 'Symmetries')
+                # Parse the symmetries of the active orbitals
                 for i, line in zip(range(num_active), inputfile):
                     reg = r'(\d+)  IRREP= (\d+) \((\w+)\)'
                     groups = re.search(reg, line).groups()
                     mo, irrep_idx, irrep = groups
 
+            # Skip until the system specific settings
+            # This will align the cases of symmetry on and off
             line = next(inputfile)
             while line[:25] != 'SYSTEM-SPECIFIC SETTINGS:':
                 line = next(inputfile)
@@ -1023,19 +1027,18 @@ States  Energy Wavelength    D2        m2        Q2         D2+m2+Q2       D2/TO
             # Number of active orbitals           ...    4
             # Total number of electrons           ...    4
             # Total number of orbitals            ...   20
-
             num_el = int(next(inputfile).split()[-1])
             num_orbs = int(next(inputfile).split()[-1])
             total_el = int(next(inputfile).split()[-1])
             total_orbs = int(next(inputfile).split()[-1])
 
-            self.skip_lines(inputfile, ['b', 'Determined'])
-
             # Determined orbital ranges:
             #    Internal       0 -   -1 (   0 orbitals)
             #    Active         0 -    3 (   4 orbitals)
             #    External       4 -   19 (  16 orbitals)
+            self.skip_lines(inputfile, ['b', 'Determined'])
             orbital_ranges = []
+            # Parse the orbital ranges for: Internal, Active, and External orbitals
             for i in range(3):
                 vals = next(inputfile).split()
                 start, end, num = int(vals[1]), int(vals[3]), int(vals[5])
@@ -1047,9 +1050,6 @@ States  Energy Wavelength    D2        m2        Q2         D2+m2+Q2       D2/TO
             line = next(inputfile)
             while line[:8] != 'CI-STEP:':
                 line = next(inputfile)
-            self.skip_line(inputfile, 'CI strategy')
-
-            num_blocks = int(next(inputfile).split()[-1])
 
             # CI-STEP:
             # CI strategy                         ... General CI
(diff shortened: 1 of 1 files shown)


src/cclib/parser/orcaparser.py
```diff
@@ -934,6 +935,229 @@ def energy_intensity(line):
...
+            vals = next(inputfile).split()
+            if vals[0] == 'Symmetry':
+                # Symmetry section is only printed if symmetry is used
+                assert vals[-1] == 'ON'
+                point_group = next(inputfile).split()[-1]
+                used_point_group = next(inputfile).split()[-1]
+                num_irreps = int(next(inputfile).split()[-1])
+                num_active = 0
+                for i, line in zip(range(num_irreps), inputfile):
+                    reg = r'Irrep\s+(\w+) has\s+(\d+) SALCs \(ofs=\s*(\d+)\) #\(closed\)=\s*(\d+) #\(active\)=\s*(\d+)'
```
ATenderholt: How about using `re.compile` outside of the for-loop so it only has to be created once?

src/cclib/parser/orcaparser.py
```diff
@@ -934,6 +935,229 @@ def energy_intensity(line):
...
+                    groups = re.search(reg, line).groups()
+                    irrep = groups[0]
+                    salcs, ofs, closed, active = map(int, groups[1:])
+                    num_active += active
+                self.skip_line(inputfile, 'Symmetries')
+                for i, line in zip(range(num_active), inputfile):
+                    reg = r'(\d+)  IRREP= (\d+) \((\w+)\)'
```
ATenderholt: Same here. Use `re.compile` outside of for loop.

src/cclib/parser/orcaparser.py
```diff
@@ -934,6 +935,229 @@ def energy_intensity(line):
...
+                    groups = re.search(reg, line).groups()
+                    mo, irrep_idx, irrep = groups
+
+            line = next(inputfile)
+            while line != 'SYSTEM-SPECIFIC SETTINGS:\n':
```
ATenderholt: Is this robust on Windows?

src/cclib/parser/orcaparser.py
```diff
@@ -934,6 +935,229 @@ def energy_intensity(line):
...
+            for i in range(3):
+                vals = next(inputfile).split()
+                start, end, num = int(vals[1]), int(vals[3]), int(vals[5])
+                # Change from inclusive to exclusive in order to match python
+                end = end + 1
+                assert end - start == num
+                orbital_ranges.append((start, end, num))
+
+            line = next(inputfile)
+            while line != 'CI-STEP:\n':
```
ATenderholt: Robust in windows? Maybe you can do something like `while line[:8] != 'CI-STEP:'`.

src/cclib/parser/orcaparser.py
```diff
@@ -934,6 +935,229 @@ def energy_intensity(line):
...
+                # If using symmetry
+                if vals[0] == 'Irrep':
+                    irrep_idx = int(vals[-2])
+                    irrep = vals[-1].strip('()')
+                    vals = next(inputfile).split()
+                num_confs = int(vals[-1])
+                num_csfs = int(next(inputfile).split()[-1])
+                num_roots = int(next(inputfile).split()[-1])
+                for r, line in zip(range(num_roots), inputfile):
+                    reg = r'=(\d+) WEIGHT=\s*(\d\.\d+)'
```
ATenderholt: Compile regex outside of loop.

src/cclib/parser/orcaparser.py
jevandezande: Good question, probably not. Can anyone who uses Windows chime in?

src/cclib/parser/orcaparser.py
jevandezande: Compiling doesn't always result in a significant speedup due to caching done by the interpreter ([compiling python regex](https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile)). I imagine the overhead of keeping track of the regexes may make this code more difficult to maintain. Furthermore, this parses 10 lines at most, so looking for a potential speedup in this section of the code is like lipstick on a pig.
Of course, hard results are much more useful. Thus I tested what I expected to be the section of the code that can most benefit from compiled regexes. On 50 roots of chemists water, the [non-compiled ran faster](https://pastebin.com/h0xXzzVw)!? Python does a lot of magic behind the scenes with caching. I tried with a variety of different input, and the compiled version was consistently slower (though only marginally).
Barring some error in my testing technique (very possible), I think that compiling regular expressions only makes the coding more difficult. Any speedup that could be had is probably inconsequential compared to the rest of the code.

jevandezande: Also, from the official python documentation on [re.compile](https://docs.python.org/3.5/library/re.html#re.compile).
"Note: The compiled versions of the most recent patterns passed to re.compile() and the module-level matching functions are cached, so programs that use only a few regular expressions at a time needn’t worry about compiling regular expressions."

jevandezande:
An Oca 2.6 regression fails (dvb_td) for my bug fix of the absorption since it has the same name. The only way to deal with this would require some ugly reworking of the code as two different methods of printing have the same header. Orca 2.6 is over 10 years old and has been superseded by two major version numbers, would it be possible to delete the regression?
Is there a better way to work around this? Perhaps parsing the version number as utilizing that?

src/cclib/parser/orcaparser.py
ATenderholt: That's really good to know. I didn't realize that patterns were cached by module-level functions.

For your test, perhaps the reason the non-compiled versions ran faster was because you used `re.search` for both string regex and compiled regex instead of using the `search()` function of the compiled regex. For example:

```
prog = re.compile(pattern)
result = prog.search(string)
```
 instead of 

```
prog = re.compile(pattern)
result = re.search(prog, string)
```

Of course, those may be equivalent behind-the-scenes.

Regardless, my suggestion for compiling the regex and place outside of the loop was also about code style (i.e only variable objects in the for-loop), but I don't feel strongly about that since in this case, putting the regex string in the loop is to make the code easier to read.

berquist:
> Is there a better way to work around this? Perhaps parsing the version number as utilizing that?

This is done in the Psi parser in a pretty rough way, where `self.version` is set to 3 or 4:

```python
def before_parsing(self):

    # There are some major differences between the output of Psi3 and Psi4,
    # so it will be useful to register which one we are dealing with.
    self.version = None

    # Early beta versions of Psi4 normalize basis function
    # coefficients when printing.
    self.version_4_beta = False

    # This is just used to track which part of the output we are in for Psi4,
    # with changes triggered by ==> things like this <== (Psi3 does not have this)
    self.section = None
```

In the future we probably want to move to a solution based on `metadata`. It's already parsed for [ORCA](https://github.com/cclib/cclib/blob/a4702020165012d807b05e3aafe2b8a1f2c595d2/src/cclib/parser/orcaparser.py#L63), so you can use it:

``` python

        #  extract the version number first
        if "Program Version" in line:
```

src/cclib/parser/orcaparser.py
langner: Or l.strip() != "..."

langner:
@jevandezande do you plan to continue this? You'll need to make some changes after we moved around files in #498.

jevandezande:
I've rebased the pull request. Now I just need input into what people think should be saved. I currently parses pretty much everything that ORCA outputs, which is way more data than needed.
I've hacked out some tests (not pushed to PR), but they still need some more polishing.

langner:
I would recommend you add very verbose comments now to the code you added here, and that we merge that in. Then work separately on pulling the parsed data out into attributes, one attribute at a time, so we can discuss each in a broader context.

Also, for the future, please separate fixes (there's one here IIUC) and unrelated changes.

jevandezande:
I've added more comments to the parsing. Hopefully how everything is being parsed is very understandable.

berquist:
Poke. If you rebase this, I'll merge it.

jevandezande:
Rebased.

Issue #448, opened 2017-11-14
Specifying the geometry to write out to xyz file


ChayaSt:
I need to be able to write out specific coordinates to an XYZ file. It will be great if `writexyz` can have that option.

berquist:
Let me know if the contents of that PR are sufficient. I'm assuming it's more important to be flexible from Python than the command line?

PR #449, opened 2017-11-15
Allow generating XYZ files with arbitrary zero-based indices.

berquist:
Implements feature request from https://github.com/cclib/cclib/issues/448.

It may be nice to have increased flexibility for `--indices`.


ChayaSt:
When I try `data.writexyz(indices=index)` I get the following error:

`TypeError: writexyz() got an unexpected keyword argument 'indices'`

This might be because `writexyz` in [`data.py`](https://github.com/berquist/cclib/blob/xyzwriter/src/cclib/parser/data.py#L310-L312) doesn't pass `**kwargs`. 
Actually, I think that's it. When I add `**kwargs` to my clone it recognized the argument.

berquist:
I glossed over that you were using the writexyz method. I'm going to rewrite this to not use kwargs.

ChayaSt:
It's working for me now. 
Thanks!

PR #450, opened 2017-11-15
Append DF-MP2 to metadata methods

ChayaSt:
When parsing a psi4 output file with the newer `DF-MP2` method, `data.metadata` has an empty list for `methods`. I added [this line](https://github.com/ChayaSt/cclib/blob/f959579c87e02f26e869f1b34e1c02b27f5149b8/src/cclib/parser/psiparser.py#L749) to fix this.


berquist:
This is fine, but I would not be surprised if this changes at some point to be simply `MP2` with information about density fitting captured elsewhere; we are going to be working with MolSSI and implementing their metadata spec, which will be more comprehensive.

PR #454, opened 2017-12-11
Add file reader interface and XYZ reader implementation

berquist:
Because the ORCA 4.0 BOMD implementation requires reading coordinates from an XYZ trajectory, and we will need to read QCJSON files, this PR includes a general interface for reading files that mimics the writer.

The PR also includes a small cleanup of the CJSON code, especially converting all the Python from camelCase to snake_case. The JSON keys are untouched.

Right now, the parser and reader infrastructure is almost completely separate. In the near future they should be combined, so all our effort of reading URLs/streams works for all files.


langner:
I don't quite get why we need to combine the reader/writer and parser classes, or I am misunderstanding your intentions. This topic is orthogonal to this PR anyway, so we don't need to discuss it here.

src/cclib/io/ccio.py
```diff
@@ -131,49 +138,17 @@ def guess_filetype(inputfile):
...
         source - a single logfile, a list of logfiles (for a single job),
                  an input stream, or an URL pointing to a log file.
-        *args, **kargs - arguments and keyword arguments passed to filetype
+        *args, **kwargs - arguments and keyword arguments passed to filetype
 
-    Returns:
-      one of ADF, DALTON, GAMESS, GAMESS UK, Gaussian, Jaguar, Molpro, MOPAC,
-      NWChem, ORCA, Psi, QChem, CJSON or None (if it cannot figure it out or
-      the file does not exist).
+    Returns: one of ADF, DALTON, GAMESS, GAMESS UK, Gaussian, Jaguar, Molpro,
```
langner: In this format "returns" is usually on its own line. I think it's for easier parsing with sphinx.

src/cclib/io/ccio.py
```diff
@@ -276,26 +256,53 @@ def ccopen(source, *args, **kargs):
...
-    be extended to other formats and libraries, too.
+    Currently this will read files with Open Babel, but this can
+    easily be extended to other formats and libraries, too.
     """
 
     if isinstance(source, str):
         ext = os.path.splitext(source)[1][1:].lower()
         if _has_cclib2openbabel:
-            if ext in ('xyz', ):
+            import pybel as pb
```
langner: Is pybel always available if python-openbabel is installed?

src/cclib/io/filereader.py
```diff
@@ -0,0 +1,44 @@
...
+# Copyright (c) 2017, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
+
+"""Generic file reader and related tools"""
+
+
+class Reader(object):
+    """Abstract class for reader objects."""
```
langner: If this is really an abstract class, then use abc?

berquist: abc? https://docs.python.org/3.6/library/abc.html?

src/cclib/io/ccio.py
berquist: Yes.

src/cclib/io/__init__.py
```diff
@@ -9,10 +9,12 @@
 
 from .cjsonwriter import CJSON as CJSONWriter
```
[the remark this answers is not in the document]
berquist: Do you not want the reader imports separated from the writers? It looks like:

``` python
from .cjsonwriter import CJSON as CJSONWriter
from .cmlwriter import CML
from .xyzwriter import XYZ as XYZWriter
from .moldenwriter import MOLDEN
from .wfxwriter import WFXWriter

from .cjsonreader import CJSON as CJSONReader
from .xyzreader import XYZ as XYZReader
```

berquist:
This will need to go in before https://github.com/cclib/cclib/pull/458 so I can fix up the imports from this PR in that PR.

langner:
> This will need to go in before #458 so I can fix up the imports from this PR in that PR.

Oops, already merged that one in :)

src/cclib/io/filereader.py
langner: Yeah, if if it's really meant to be an abstract base class.

src/cclib/io/__init__.py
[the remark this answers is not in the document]
langner: I would prefer to just put them all together. I'd like to see readers and writers in the same module in the future, maybe merged into one class?

src/cclib/io/filereader.py
berquist: I guess I'm not really sure. I'm looking for C++-style behavior.

langner: Could you be more specific what "C++-style behavior"?

langner:
Look pretty much ready for merging (sans conflicts), but moving to v1.5.4 to speed up the v1.5.3 release.

langner:
Looks like you need to make some changes after we moved around files in #498.

langner:
> Merge branch 'master' into cjson  …

In the future, you might find it better to rebase instead of merging. Anyway, it's still reporting there are some conflicts.

Issue #455, opened 2017-12-11
Move from unittest to pytest and enable coverage reporting


berquist:
This should be a drop-in replacement, since we don't use any special feature of `unittest`, but all imports need to be converted to relative module imports.

```
/home/eric/development/cclib_berquist (master|…) (env: cclib_berquist 3.6.3)
[eric@copper]$ pytest -v test
============================================================================================================================ test session starts =============================================================================================================================
platform linux -- Python 3.6.3, pytest-3.3.1, py-1.5.2, pluggy-0.6.0 -- /media/Backup/virtualenvs/3/cclib_berquist/bin/python3.6
cachedir: .cache
rootdir: /home/eric/development/cclib_berquist, inifile:
plugins: cov-2.5.1
collected 2 items / 5 errors                                                                                                                                                                                                                                                 

=================================================================================================================================== ERRORS ===================================================================================================================================
____________________________________________________________________________________________________________________ ERROR collecting test/test_bridge.py ____________________________________________________________________________________________________________________
ImportError while importing test module '/home/eric/development/cclib_berquist/test/test_bridge.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test/test_bridge.py:14: in <module>
    from testopenbabel import *
E   ModuleNotFoundError: No module named 'testopenbabel'
_____________________________________________________________________________________________________________________ ERROR collecting test/test_data.py _____________________________________________________________________________________________________________________
ImportError while importing test module '/home/eric/development/cclib_berquist/test/test_data.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test/test_data.py:40: in <module>
    all_modules = {tn: importlib.import_module('data.test' + tn) for tn in module_names}
test/test_data.py:40: in <dictcomp>
    all_modules = {tn: importlib.import_module('data.test' + tn) for tn in module_names}
/usr/lib/python3.6/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
E   ModuleNotFoundError: No module named 'data.testSP'
______________________________________________________________________________________________________________________ ERROR collecting test/test_io.py ______________________________________________________________________________________________________________________
ImportError while importing test module '/home/eric/development/cclib_berquist/test/test_io.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test/test_io.py:14: in <module>
    from testccio import *
E   ModuleNotFoundError: No module named 'testccio'
[...]

Issue #367, opened 2017-05-02
cclib imports should be relative


baoilleach:
cclib should use relative imports, and not require its presence on the PYTHONPATH.

For example, filewriter.py contains:
   from cclib.parser.utils import PeriodicTable

I bundle cclib with GaussSum, and so this fails.

baoilleach:
The problem with GaussSum was addressed by moving the cclib codebase up a directory so that "from cclib" now works. However, I do have a feeling that within the library all imports should be relative. It would be good to look up best practices on this as it must be a problem that many Python libraries have faced.

berquist:
> The problem with GaussSum was addressed by moving the cclib codebase up a directory so that "from cclib" now works.

Right, taking advantage of the quick-and-dirty import from the file or folder that's in my current directory, and nothing breaks because there are no relative imports in the library but there are `__init__.py` files in the correct locations.

One case where this is a potential problem is if there are multiple copies of cclib in your PYTHONPATH, it's easier for subtle problems to appear. I might have had problems with this in the past where something is wrong with my virtualenv I develop in, and I pick up the system copy instead since they live in the same module namespace, which shouldn't be possible with the relative imports.

Here is a comment on it from [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports):

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path ):

>     import mypkg.sibling
>     from mypkg import sibling
>     from mypkg.sibling import example

> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:

>     from . import sibling
>     from .sibling import example

> Standard library code should avoid complex package layouts and always use absolute imports.

> Implicit relative imports should never be used and have been removed in Python 3.

where an implicit relative import would be just `import sibling`.

I am not convinced that our absolute imports are too verbose and need to be shortened, but I agree that we need more examples. I decided to go with this package structure for a project of mine, and it can be irritating working inside of it; a loose package script that was previously run as `python test_ecd.py` changes to `python -m pyresponse.test_ecd`, so we can already see how our testing scripts might change.

langner:
Let's deal with this for the next release (1.5.2 or 1.6, whatever comes first).

langner:
Will review for v1.5.3.

berquist:
[...]

PR #456, opened 2017-12-12
Enable use of pytest and code coverage analysis

berquist:
Issue: https://github.com/cclib/cclib/issues/455

For this to work, it also solves https://github.com/cclib/cclib/issues/367. Everything should be a module now.

Currently Travis is not running this, because the output style isn't great. pytest as a drop-in replacement works, but our tests (in particular regressions) are not in the "right" format. Maybe pytest is only useful for us to run while developing?

- [x] Can I remove all the `if __name__ == '__main__'` calls in the source? Maintaining that is difficult and I don't think they're ever run.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,6 @@
+collect_ignore = [
+    # TODO doctest won't ignore specific files?
+    # https://stackoverflow.com/q/41358778/3249688
+    # 'src/cclib/progress/qt4progess.py',
+    'src/cclib/progress',
+]
--- a/conftest.py
+++ b/conftest.py
@@ -1,6 +1,28 @@
-collect_ignore = [
-    # TODO doctest won't ignore specific files?
-    # https://stackoverflow.com/q/41358778/3249688
-    # 'src/cclib/progress/qt4progess.py',
-    'src/cclib/progress',
+import sys
+
+
+version_major = sys.version_info.major
+version_minor = sys.version_info.minor
+
+
+paths_allver = [
+    'src/cclib/progress/qt4progress.py',
 ]
+
+paths_2_7_only = [
+    'src/cclib/bridge/cclib2pyquante.py',
+]
+
+
+def match_path(path, partial_paths):
+    return any(partial_path in str(path)
+               for partial_path in partial_paths)
+
+
+def pytest_ignore_collect(path, config):
+    if match_path(path, paths_allver):
+        return True
+    if version_major != 2:
+        if match_path(path, paths_2_7_only):
+            return True
+    return False
--- a/conftest.py
+++ b/conftest.py
@@ -4,7 +4,7 @@ import sys
 version_major = sys.version_info.major
 version_minor = sys.version_info.minor
 
-# Paths than should be ignored for all Python versions.
+# Paths that should be ignored for all Python versions.
 paths_allver = [
     'src/cclib/progress/qt4progress.py',
 ]
--- a/conftest.py
+++ b/conftest.py
@@ -4,25 +4,36 @@ import sys
 version_major = sys.version_info.major
 version_minor = sys.version_info.minor
 
-
+# Paths than should be ignored for all Python versions.
 paths_allver = [
     'src/cclib/progress/qt4progress.py',
 ]
 
-paths_2_7_only = [
+# Paths that should run only for Python 2.7.
+paths_only_2_7 = [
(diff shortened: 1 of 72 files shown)


src/cclib/bridge/cclib2biopython.py
```diff
@@ -12,7 +12,7 @@
     # Fail silently for now.
     pass
 
-from cclib.parser.utils import PeriodicTable
+from ..parser.utils import PeriodicTable
```
langner: I think with this we should follow PEP8. Absolute imports are recommended, as they are usually more readable and tend to be better behaved: https://www.python.org/dev/peps/pep-0008/#imports

langner:
Whatever we do, I think it would be better to stick to one test framework.

src/cclib/bridge/cclib2biopython.py
berquist: Ok, I made all the changes for `src`, but I'm not sure what to do about imports within the test directory for relative calls.

berquist:
> Whatever we do, I think it would be better to stick to one test framework.

In that case I'll set Travis to use it for now.

langner:
@ATenderholt do you have any problems with using pytest throughout?

berquist:
1. Test suite takes 7 minutes for 2.7 and 11 minutes for 3.6 to run rather than 4 minutes. I'm not sure if this is because coverage is enabled, or because of pytest.

2. Pytest is 2.7 and 3.4+, so it won't work for 3.2.

berquist:
On my desktop, first time is for unit tests, second is for regressions.

* test coverage on: 35.77 s, 147.96 s, 3 min total
* test coverage off: 12.31 s, 48.63 s, 1 min total

So that's where the time difference comes from.

ATenderholt:
I think switching to pytest is a good idea. Seems easier to write and maintain tests, and the code coverage is a nice benefit.

langner:
All right, then it's decided. Eric, I think Travis has an installation issue with pytest for Python 3.2?

berquist:
Yes, 3.4 is the minimum supported 3.x version. Can I bump the Travis version? I have some computers with access to Ubuntu 14.04, which is no longer supported, and they have 3.4.3.

langner:
I see. I'm somewhat nervous about dropping support for a Python version we already supported.

berquist:
I fixed Python 3.2 by installing ancient versions of pytest + dependencies. It may break if we start using pytest-only features on our tests.

1. Something is wrong with coverage checking. I'm not sure if checking the two trees (see Travis log) is messing up the statistics, because I thought testing regressions would make it go up.

2. Right now output from `test_all` is being silenced. I'm not sure what it will look like when a test inside of it fails.

test/method/testnuclear.py
```diff
@@ -17,12 +17,14 @@
 
 import numpy
 
-sys.path.append("..")
-from test_data import getdatafile
 from cclib.method import Nuclear
 from cclib.parser import QChem
 from cclib.parser import utils
 
+sys.path.append("..")
```
dgasmith: To be safe, shouldn't this be `sys.path.insert(1, "..")`.

dgasmith:
Big fan of `pytest` overall, after getting used to it I use it for all of my projects.

conftest.py
```diff
@@ -0,0 +1,39 @@
+import sys
+
+
+version_major = sys.version_info.major
+version_minor = sys.version_info.minor
+
+# Paths than should be ignored for all Python versions.
```
langner: than->that

berquist:
Ok, I’ll turn doctests into unit tests and remove all doctests from the
codebase.

As far as I can tell, pytest is dependent on solidifying the import
situation. I am also going to try and do the path prepend  Daniel suggested.

I’ll break it apart right now, just need some coffee...

berquist:
For those who see this in the future and wonder why it needs to be `sys.path.insert(1, '..')` and not `sys.path.insert(0, '..')`: https://stackoverflow.com/q/10095037

Short answer: it will mess with using virtualenvs.

langner:
> For those who see this in the future and wonder why it needs to be sys.path.insert(1, '..') and not sys.path.insert(0, '..'): https://stackoverflow.com/q/10095037
> 
> Short answer: it will mess with using virtualenvs.

Could you add a comment to this effect near the code?

berquist:
The new PR is https://github.com/cclib/cclib/pull/491.

PR #458, opened 2018-02-05
Switch completely to using absolute imports

berquist:
These are all the commits from https://github.com/cclib/cclib/pull/456 that make imports consistently absolute rather than relative. This makes it possible to run properly as a module.

It also switches path manipulation to use `sys.path.insert(1, 'path')` rather than `sys.path.append('path')` per https://stackoverflow.com/q/10095037.


langner:
Excellent!

PR #463, opened 2018-02-23
fix accidental np instead of numpy in regression.py

berquist:
https://github.com/cclib/cclib/pull/462 accidentally broke master.


langner:
Why didn't it break Travis in that PR?

berquist:
The log shows that since this was merged before the PR that added the data file, the test never ran. So, the issue with the order in which PRs are merged remains.

langner:
> The log shows that since this was merged before the PR that added the data file, the test never ran. So, the issue with the order in which PRs are merged remains.

Ah yes, that makes sense. I don't think there's any simple way to get rid of that race condition... we'll just have to make sure we use the right order by convention.

PR #466, opened 2018-03-08
fixed some typos in ccwrite

pasenor:
There are two unfortunate typos causing the following errors:

	$ python ../src/scripts/ccwrite cjson ./DALTON/basicDALTON-2013/C_bigbasis.out
    
      File "../src/scripts/ccwrite", line 45
        type=int,
           ^
    SyntaxError: invalid syntax

and then

      File "/opt/pycharm-community-2017.3.4/helpers/pydev/pydevd.py", line 1668, in <module>
        main()
      File "/opt/pycharm-community-2017.3.4/helpers/pydev/pydevd.py", line 1662, in main
        globals = debugger.run(setup['file'], None, None, is_module)
      File "/opt/pycharm-community-2017.3.4/helpers/pydev/pydevd.py", line 1072, in run
        pydev_imports.execfile(file, globals, locals)  # execute the script
      File "/opt/pycharm-community-2017.3.4/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
        exec(compile(contents+"\n", file, 'exec'), glob, loc)
      File "/home/gosha/code/cclib/src/scripts/ccwrite", line 101, in <module>
        main()
      File "/home/gosha/code/cclib/src/scripts/ccwrite", line 97, in main
        **ccwrite_kwargs)
    TypeError: ccwrite() got multiple values for argument 'indices'

which is due to the order of unnamed arguments in the `ccwrite` invocation

### DIFF
--- a/src/cclib/io/ccio.py
+++ b/src/cclib/io/ccio.py
@@ -297,7 +297,7 @@ def fallback(source):
 
 
 def ccwrite(ccobj, outputtype=None, outputdest=None,
-            indices=None, terse=False , returnstr=False,
+            indices=None, terse=False, returnstr=False,
             *args, **kwargs):
     """Write the parsed data from an outputfile to a standard chemical
     representation.
--- a/src/scripts/ccwrite
+++ b/src/scripts/ccwrite
@@ -41,7 +41,7 @@ def main():
                         action='store_true',
                         help='use experimental features (currently optdone_as_list)')
 
-    parser.add_argument('-i', '--index'
+    parser.add_argument('-i', '--index',
                         type=int,
                         default=None,
                         help='optional zero-based index for which structure to extract')
@@ -93,7 +93,7 @@ def main():
 
         # The argument terse presently is only applicable to
         # CJSON/JSON formats
-        ccwrite(data, outputtype, outputdest, terse, indices=index,
+        ccwrite(data, outputtype, outputdest, terse=terse, indices=index,
                 **ccwrite_kwargs)
 
 


pasenor:
This is related to #377

langner:
Thanks for finding this! Could you add a test that covers this?

pasenor:
Yes, but I could use some pointers on how to run the regression suite. It now complains there is no module `__main__.test_data`

langner:
> and it complains there is no module __main__.test_data

What's the full output?

pasenor:
$ cd cclib/test
    $ python regression.py 
    
      File "regression.py", line 69, in <module>
        from .test_data import all_modules
    ModuleNotFoundError: No module named '__main__.test_data'; '__main__' is not a package

Looks like regression.py should not be run directly as a script? I temporarily fixed it by making the imports absolute

langner:
> Looks like regression.py should not be run directly as a script? I temporarily fixed it by making the imports absolute

I should be runable directly as a script. I am also having issues here, but slightly different. We need to figure this out.

pasenor:
Well, in travis it's called as a module, and probably we cannot have it both ways.
Anyway, I now don't think this is a good case for a regression test, but rather a unit test with a simple check that cclib.io.ccio.ccwrite is called correctly.

Please take a look at the last commit, if you have no objections to the approach, then we probably should expand it and try to catch some more little bugs, for example it seems the parameter `--terse` is broken

    $ cd cclib/data
    $ python ../src/scripts/ccwrite cjson ./ADF/basicADF2007.01/dvb_gopt.adfout --terse=false
    
    usage: ccwrite [-h] [-v] [-t] [-u] [-i INDEX]
                   {json,cjson,cml,xyz,molden,wfx} compchemlogfile
                   [compchemlogfile ...]
    ccwrite: error: argument -t/--terse: ignored explicit argument 'false'
    
    $ python ../src/scripts/ccwrite cjson ./ADF/basicADF2007.01/dvb_gopt.adfout --terse=true
    usage: ccwrite [-h] [-v] [-t] [-u] [-i INDEX]
                   {json,cjson,cml,xyz,molden,wfx} compchemlogfile
                   [compchemlogfile ...]
    ccwrite: error: argument -t/--terse: ignored explicit argument 'true'

requirements.txt
```diff
@@ -3,3 +3,6 @@ biopython
 openbabel
 # For some reason PyQuante can't be installed from PyPI.
 # PyQuante
+
+# for unit tests
+mock==2.0.0 ; python_version < '3.3'
```
berquist: I don't understand the syntax. Does this mean only install mock for Python versions < 3.3, or something else?

berquist:
> It should be runable directly as a script. I am also having issues here, but slightly different. We need to figure this out.

I made it only be usable as a module, not a script, so that pytest would work, but we aren't using pytest yet.

requirements.txt
pasenor: [Yes, apparently it's called an "environment marker"](https://www.python.org/dev/peps/pep-0496/)

In python 3.4 and later `mock` is in the standard library as `unittest.mock`

pasenor: I just looked at travis logs, and it never does `pip install -r requirements.txt` and all dependencies are listed individually in `.travis.yml`. Is this because of the problems with pyquante?

berquist: I was probably using it for a while, discovered the versioning issue, and just left the file behind.

I may also have tried `python_version` in the past and I just didn't remember, and older versions of pip had trouble with it?

One comment is that PyQuante requires 2.7.

If you can get this to work on all our testing versions, that would simplify Travis immensely.

pasenor: PR #467

langner:
> It should be runable directly as a script. I am also having issues here, but slightly different. We need to figure this out.
> 
> I made it only be usable as a module, not a script, so that pytest would work, but we aren't using pytest yet.

Oh, that's not great. I really liked running it locally as a script after changing code.

manifest.py
```diff
@@ -18,9 +18,6 @@
 
 source = os.path.join('src', 'cclib')
 files.append(os.path.join(source, "__init__.py"))
-files.append(os.path.join("src", "scripts", "ccget"))
```
langner: Any chance you could put these other fixes in a separate PR? Just keep the ones to fix ccwrite here.

requirements.txt
```diff
@@ -4,3 +4,6 @@ openbabel
 
 # For some reason PyQuante can't be installed from PyPI.
 https://sourceforge.net/projects/pyquante/files/PyQuante-1.6/PyQuante-1.6.5/PyQuante-1.6.5.tar.gz  ; python_version <= '2.7'
+
+# for unit tests
+mock==2.0.0 ; python_version < '3.3'
```
langner: Same for all these other changes not directly related to ccwrite.

manifest.py
pasenor: It's related to testing of ccwrite (is there a simpler way I'm missing?). I can move it to a different PR, but then this one will be left without a test.

pasenor:
>It should be runable directly as a script. I am also having issues here, but slightly different. We need to figure this out.

>I made it only be usable as a module, not a script, so that pytest would work, but we aren't using pytest yet.

>Oh, that's not great. I really liked running it locally as a script after changing code.

You can still run it locally as 

        python -m test.regression

pasenor:
Moved the tests to a separate PR #473 since they involve significant changes to the scripts

langner:
> You can still run it locally as
> 
>     python -m test.regression

True. We can look into that later. I think the ccwrite stuff here is good to go, yes?

PR #467, opened 2018-03-11
using requirements.txt in Travis CI

pasenor:
The new version of pip allows to do all version management directly in requirements.txt, so there is no more need to repeat the dependencies separately in yaml.


langner:
This is excellent, thank you!

Issue #452, opened 2017-11-20
closed_shell() results in error for Gaussian files


jaimergp:
```
import cclib as cc
p = cc.parser.Gaussian("Au1_ts3.ok.qfi").parse()
p.closed_shell
```
First, line 385 in cclib.parser.data has a typo that results in AttributeError. It should be `orbitals.Orbitals`, but now it is

```
return orbitals.orbitals(self).closed_shell()
```

After fixing that, it fails because data contains no attribute named `mocoeffs`. I guess because my file does not contain information for that. Anyway, it should be fail gracefully with a nice error message, right?

```
$CONDA_ENV/lib/python3.6/site-packages/cclib/method/orbitals.py in closed_shell(self)
     43         # If there are beta orbitals, we can assume the system is closed
     44         # shell if the orbital energies are identical within numerical accuracy.
---> 45         if len(self.data.mocoeffs) == 2:
     46             precision = 10e-6
     47             return numpy.allclose(*self.data.moenergies, atol=precision)
```

berquist:
Hi Jaime,

Sorry for the late response. Could you post an output file that triggers this behavior?

jaimergp:
Hi! Yes, of course. This is the (zipped) file I used.


[Au1_ts3.zip](https://github.com/cclib/cclib/files/1552353/Au1_ts3.zip)

kunalsharma05:
I would like to work on this issue. A simple try except would be able to check if data has attribute "mocoeffs" or not. 
@langner What should be the message that should be shown if there is an Attribute error?

ATenderholt:
I don't think it should be handled via try-except. Better to validate required attributes in the `Orbitals` constructor. As far as messages, perhaps look at some other examples (e.g. WFX writer).

langner:
@kunalsharma05 I agree with @ATenderholt, it would be much better to handle this sort of thing generically. So in this case, that would mean in the initializer of the base class (`Method.__init__`) - it's defined in `calculationmethod.py`.

The WFX writer logic Adam was referring to: https://github.com/cclib/cclib/blob/master/src/cclib/io/wfxwriter.py#L93

kunalsharma05:
@langner @ATenderholt So basically we want to have a list of required attributes for each sub class (e.g `Orbitals(Method)`) and place a check in the initializer of the base class (`Method.__init__`) that whether the parsed data has the required attributes or not. 
If the required attributes are not satisfied then we do not create an object for that class.

[...]

PR #469, opened 2018-03-14
Fix for Issue #452

kunalsharma05:
Hello,
Can you please review my commit for Issue #452. I have written the code for only orbitals class and will extend it to other method classes.
The usage is the same:
`import cclib as cc`
`p = cc.parser.Gaussian("Au1_ts3.ok.qfi").parse()`
`p.closed_shell`


src/cclib/method/calculationmethod.py
```diff
@@ -31,24 +31,45 @@ class Method(object):
     >>> import cda, cspa, density, fragments, lpa, mbo, mpa, nuclear, opa, population, volume
     """
 
-    def __init__(self, data, progress=None, loglevel=logging.INFO, logname="Log"):
+    def __init__(self, data, requiredAttr, progress=None, loglevel=logging.INFO, logname="Log"):
```
ATenderholt: I don't think this should be passed in as a parameter. It should default to an empty list, and subclasses of `Method` can override as necessary. See https://github.com/cclib/cclib/blob/master/src/cclib/io/filewriter.py#L31 and https://github.com/cclib/cclib/blob/master/src/cclib/io/wfxwriter.py#L93.

src/cclib/method/calculationmethod.py
```diff
@@ -31,24 +31,45 @@ class Method(object):
...
+            else:
+                miss.append(attr)
+
+        if len(miss) == 0:
+            pass
+            return ''
+        else:
+            self.initFlag =1
+            missingAttr = ','.join(i for i in miss)
+            return missingAttr
```
ATenderholt: This function seems unnecessarily complex. `get` is never used, `initFlag` is redundant with an empty `missingAttr`, `pass` is pointless, etc. Look at the `FileWriter` example mentioned above.

src/cclib/method/orbitals.py
```diff
@@ -22,12 +22,14 @@
 class Orbitals(Method):
     """A class for orbital related methods."""
 
-    def __init__(self, data, progress=None, \
+    def __init__(self, data, requiredAttr = ['mocoeffs','moenergies','homos'], progress=None, \
```
ATenderholt: By providing a default for `requiredAttr`, you're essentially implying that `requiredAttr` may contain more (or less!) than those three attributes, but it's up to the user of the `Orbitals` method to known which attributes are required.

src/cclib/method/orbitals.py
```diff
@@ -22,12 +22,14 @@
...
                  loglevel=logging.INFO, logname="Log"):
 
         # Call the __init__ method of the superclass.
-        super(Orbitals, self).__init__(data, progress, loglevel, logname)
+        super(Orbitals, self).__init__(data,requiredAttr, progress, loglevel, logname)
         self.fragresults = None
+        if self.initFlag:
+            print('Missing attributes: %s' % (self.attrErrorMessage))
```
ATenderholt: Why should this object even be valid if it's missing the required attributes?

src/cclib/method/orbitals.py
```diff
@@ -39,16 +41,17 @@ def __repr__(self):
 
     def closed_shell(self):
         """Return Boolean indicating if system is closed shell."""
-
+        if self.initFlag:
+            return None
```
ATenderholt: This contradicts the current docstring. Also, by returning `None`, you've introduced a route for a very subtle bug since usage is probably more akin to `if orbitals.closed_shell`:

```
>>> values = [ None, True, False ]
>>> for value in values:
...     if value:
...             print("Value is true")
...     else:
...             print("Value is false")
... 
Value is false
Value is true
Value is false
```

src/cclib/parser/data.py
```diff
@@ -75,8 +75,8 @@ class ccData(object):
         scftargets -- targets for convergence of the SCF (array[2])
         scfvalues -- current values for convergence of the SCF (list of arrays[2])
         temperature -- temperature used for Thermochemistry (float, kelvin)
-        transprop -- all absorption and emission spectra (dictionary {name:(etenergies, etoscs)})
-            WARNING: this attribute is not standardized and is liable to change in cclib 2.0
+        transprop -- all absorption and emission spectra (dictionary {name:(etenergies, etoscs)})
+            WARNING: this attribute is not standardized and is liable to change in cclib 2.0
```
ATenderholt: Don't change whitespace in an unrelated Pull Request.

src/cclib/parser/data.py
```diff
@@ -382,7 +382,7 @@ def nelectrons(self):
 
     @property
     def closed_shell(self):
-        return orbitals.orbitals(self).closed_shell()
+        return orbitals.Orbitals(self).closed_shell()
```
ATenderholt: What's with `orbitals` being changed to `Orbitals`?

kunalsharma05: `orbitals` was mostly a typo, since the name of the class is `Orbitals` so I changed it accordingly. If you try to run closed_shell with `orbitals.orbitals`, it gives the following error:
`AttributeError: 'module' object has no attribute 'orbitals'`

src/cclib/method/calculationmethod.py
kunalsharma05: Got it! Thanks a lot!

src/cclib/method/orbitals.py
kunalsharma05: So the user would know of the required attributes, they just wouldn't know if those attributes were parsed into the ccData object?

src/cclib/parser/data.py
kunalsharma05: My Bad. Sorry about that.

kunalsharma05:
@ATenderholt Thanks for the previous review. I have made the changes you asked for. Please let me know if this works. Thanks in advance!

src/cclib/parser/data.py
ATenderholt: Ok.

src/cclib/parser/data.py
ATenderholt: Am I looking at an outdated diff? It looks like you still have whitespace changes.

kunalsharma05:
I made a mess of the branch while re-basing. I will submit a new PR.

Issue #472, opened 2018-03-15
Add tests for the main scripts (ccwrite, ccget)


pasenor:
At the moment there is no easy way to test the main scripts (`ccget`, `ccwrite`, probably `cda`) and no tests on them, which leads to bugs (#377, #466, probably more).

There are different ways to enable testing:
- move the scripts to a module and write a test that imports their main functions and invokes them. 
- write regression tests that run scripts using shell commands and look at the output

I prefer the first approach: it is faster, more granular (we are looking for errors in script files, not in the functions they invoke); it also allows us to use mocks and thus not deal with output files.

langner:
@pasenor Was there anything else you wanted to add for this?

langner:
Closing this, moved last two points for consideration to #54.

PR #473, opened 2018-03-15
Adding unit tests for the main scripts (ccwrite and ccget)

pasenor:
Adding unit tests for ccget and ccwrite, using the first approach outlined in #472 

I moved the scripts to `*py` files to enable a cleaner importing, since the `import` statement does not recognise files without extension and we would have to do something like

    imp.load_source('scripts/ccwrite)

which is arguably less clean and deprecated in python 3


test/io/testscripts.py
```diff
@@ -0,0 +1,74 @@
...
+"""Unit tests for main scripts (ccget, ccwrite)."""
+import os
+import unittest
+from io import StringIO
+
+try:
+    from unittest.mock import patch
+except ImportError:
+    from mock import patch
+
```
langner: You don't need that many blank lines between sections. Please follow the recommendations of PEP8 for formatting.

langner:
Travis is failing, so I presume this needs to wait for some other fixes?

pasenor:
> Travis is failing, so I presume this needs to wait for some other fixes?

This PR is tests only, not including the fixes to ccwrite from #466, so it fails predictably with

      File "test/io/testscripts.py", line 36, in setUp
        from scripts import ccget
      File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/cclib-1.5.2-py2.7.egg/scripts/__init__.py", line 9, in <module>
        from . import ccwrite
      File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/cclib-1.5.2-py2.7.egg/scripts/ccwrite.py", line 45
        type=int,
           ^
    SyntaxError: invalid syntax

ATenderholt:
It's also failing at another spot, which I think hints at another problem related to the changes in this PR.

```
  File "test/io/testscripts.py", line 50, in setUp
    from scripts import ccwrite
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/cclib-1.5.2-py2.7.egg/scripts/__init__.py", line 8, in <module>
    from . import ccget
ImportError: cannot import name ccget
```

test/io/testscripts.py
pasenor: fixed

pasenor:
> It's also failing at another spot, which I think hints at another problem related to the changes in this PR.

This is another facet of the same problem: due to syntax errors in the script file it cannot be imported. Now that the script is fixed it goes away, although I agree it was not the most sensible message to fail with. Added checks around import statements.

berquist:
I am ok with this. Is everyone else?

ATenderholt:
Looks good to me.

PR #477, opened 2018-03-15
Fix two bugs in ccwrite. 

szl0072:
1. Add a ',' in line 44 after '--index';
2. Remove `indices=` in line 96.


szl0072:
#466 already covered this. So I closed this up.

langner:
Thanks all the same for noticing!

PR #478, opened 2018-03-15
Fix links to docs (no tutorial anymore)

### LINKED ISSUE #470: The tutorial page is down
When I opened http://cclib.github.io/tutorial.html. It gave back a 404.

langner:
Fixes #470


ANNOUNCE
```diff
@@ -20,8 +20,7 @@ cclib also provides some calculation methods for interpreting the electronic pro
 
 (For a complete list see http://cclib.github.io/methods.html).
 
-For information on how to use cclib, see
-    http://cclib.github.io/tutorial.html
+For information on how to use cclib, see our documentation at http://cclib.github.io.
```
ATenderholt: What do you think HTTPS instead? Seems like that's become preferred and already works for us.

langner: Done, here and in README.

PR #479, opened 2018-03-16
Added required_attrs for electrons.py and nuclear.py

kunalsharma05:
This PR adds required attributes for Electrons and Nuclear method subclasses. These were similar to Orbitals subclass.


ATenderholt:
Looks good, thanks!

PR #480, opened 2018-03-17
Required attrs for subclasses of Population class

kunalsharma05:
This PR adds required attributes for MPA, LPA and CSPA method subclasses. They are child classes of Population class.

### DIFF
--- a/src/cclib/method/lpa.py
+++ b/src/cclib/method/lpa.py
@@ -16,7 +16,16 @@ from cclib.method.population import Population
 
 class LPA(Population):
     """The Löwdin population analysis"""
+
     def __init__(self, *args):
+
+        self.required_attrs = ('mocoeffs','homos','nbasis')
+
+        # If the parsed data does NOT have either of the Fragment orbital overlap 
+        # or the Atomic orbital overlap matrix
+        if not (hasattr(data, "aooverlaps") \
+             or hasattr(data, "fooverlaps") ):
+            self.required_attrs += ('aooverlaps or fooverlaps',)
 
         # Call the __init__ method of the superclass.
         super(LPA, self).__init__(logname="LPA", *args)
--- a/src/cclib/method/opa.py
+++ b/src/cclib/method/opa.py
@@ -11,7 +11,7 @@ import random
 
 import numpy
 
-from cclib.method.calculationmethod import Method
+from cclib.method.population import Population
 
 
 def func(x):
@@ -21,10 +21,18 @@ def func(x):
         return x+func(x-1)
 
 
-class OPA(Method):
+class OPA(Population):
     """Overlap population analysis."""
 
     def __init__(self, *args):
+
+        self.required_attrs = ('mocoeffs','homos','nbasis')
+
+        # If the parsed data does NOT have either of the Fragment orbital overlap 
+        # or the Atomic orbital overlap matrix
+        if not (hasattr(data, "aooverlaps") \
+             or hasattr(data, "fooverlaps") ):
+            self.required_attrs += ('aooverlaps or fooverlaps',)
 
         # Call the __init__ method of the superclass.
         super(OPA, self).__init__(logname="OPA", *args)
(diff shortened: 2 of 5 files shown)


src/cclib/method/population.py
```diff
@@ -20,6 +20,22 @@ class Population(Method):
...
 
+        # Do we have the needed attributes in the data object?
+        # Check for getting the method  being used e.g. Mulliken Population Analysis
+        if logname == 'MPA' or logname == 'LPA':
+
+            # If the parsed data does NOT have either of the Fragment orbital overlap 
+            # or the Atomic orbital overlap matrix
+            if not (hasattr(data, "aooverlaps") \
+                 or hasattr(data, "fooverlaps") ):
+                self.required_attrs = ('mocoeffs','aooverlaps','fooverlaps','homos','nbasis')
```
ATenderholt: This is the tricky part. A data instance can have either `aooverlaps` or `fooverlaps`, but not both. Perhaps these should be left out of `required_attrs` and a separate check is made in the constructor.

src/cclib/method/population.py
```diff
@@ -20,6 +20,22 @@ class Population(Method):
...
+            # If either of the two orbital overlaps is present 
+            else:
+                self.required_attrs = ('mocoeffs','homos','nbasis')
+
+        elif logname == 'CSPA':
+            self.required_attrs = ('mocoeffs','nbasis','homos')
+
```
ATenderholt: Instead of an if-else for MPA/LPA vs. CSPA, you should set the `required_attrs` in the appropriate subclass.

src/cclib/method/population.py
kunalsharma05: That makes sense. Will  do the required changes.

src/cclib/method/opa.py
```diff
@@ -21,11 +21,19 @@ def func(x):
         return x+func(x-1)
 
 
-class OPA(Method):
+class OPA(Population):
```
kunalsharma05: @ATenderholt This should be a subclass of `Population` and not of `Method` right?

langner:
Hi @kunalsharma05, do you intend to finish this one off?

kunalsharma05:
Hi @langner , I made some changes and was waiting for your review to move forward with the task.

langner:
@ATenderholt could you take a second look here? Were all you comments addressed?

@kunalsharma05 did you plan on doing something more here?

Need to move thing around after refactoring that happened during the summer. There are some formatting bits to fix, but we can do that after merging in the original change.

kunalsharma05:
@langner @ATenderholt Let me have a look real quick. Will get back to you on this by tomorrow.

src/cclib/method/lpa.py
```diff
@@ -16,8 +16,17 @@
...
+
     def __init__(self, *args):
 
+        self.required_attrs = ('mocoeffs','homos','nbasis')
+
+        # If the parsed data does NOT have either of the Fragment orbital overlap 
+        # or the Atomic orbital overlap matrix
+        if not (hasattr(data, "aooverlaps") \
+             or hasattr(data, "fooverlaps") ):
+            self.required_attrs += ('aooverlaps or fooverlaps',)
```
ATenderholt: I don't think this will work here or in the other parsers. Look at https://github.com/cclib/cclib/blob/829c0ec9ed1ca1cc335575440a7ee0ae926d5e1d/cclib/method/calculationmethod.py#L53.

src/cclib/method/population.py
langner: I think these are common to all subclasses of `Population`, which is where I put it in #642.

src/cclib/method/opa.py
langner: Done in #642.

src/cclib/method/lpa.py
langner: Done in #642.

langner:
Since Kunal is no longer active I'm closing this. Rewrote the change in #642, but I couldn't cherry pick Kunal's commits because it seems the branch they were on is gone (deleted?). Sorry to drop the history here.

PR #485, opened 2018-03-28
Promote doctests to full tests

berquist:
This is the first part of splitting https://github.com/cclib/cclib/pull/456.

I suggest reviewing by looking at each commit separately.


src/cclib/method/volume.py
```diff
@@ -15,16 +15,16 @@
 try:
     from PyQuante.CGBF import CGBF
     from cclib.bridge import cclib2pyquante
-    module_pyq = True
-except:
-    module_pyq = False
+    HAS_PYQUANTE = True
```
langner: Let's treat these as private constants, so with a leading underscore.

berquist:
Poke.

langner:
I'll merge this in for v1.5.3, just want to take a quick read.

berquist:
Ok. It's only that this, specifically leading to the imports and pytest, is a blocker for a bunch of other changes.

src/cclib/method/volume.py
```diff
@@ -38,41 +38,43 @@ class Volume(object):
        spacing -- the distance between the points in the cube
 
     Attributes:
-       data -- a numpy array of values for each point in the volume
+       data -- a NumPyarray of values for each point in the volume
```
berquist: TODO spacing typo

PR #491, opened 2018-04-05
Switch from plain unittest to pytest

berquist:
Following https://github.com/cclib/cclib/pull/485, this is the remainder of https://github.com/cclib/cclib/pull/456. The output is not pretty right now, especially `test_data` which is not expanded, but everything is being tested.


berquist:
I'm going to try and figure out how to have the output from `test_all` and the visual tests displayed. Then we should have basic parity with the current setup.

conftest.py
```diff
@@ -0,0 +1,39 @@
+import sys
+
+
+version_major = sys.version_info.major
+version_minor = sys.version_info.minor
+
+# Paths that should be ignored for all Python versions.
+paths_allver = [
```
langner: Maybe call this `paths_ignore_allver`?

conftest.py
```diff
@@ -0,0 +1,39 @@
+import sys
```
langner: Should this file go in the `test/` directory?

pytest.ini
```diff
@@ -0,0 +1,6 @@
+[pytest]
```
langner: Does this file need to go into the top level directory?

test/conftest.py
```diff
@@ -0,0 +1,39 @@
+import sys
```
langner: Where is this module used?

berquist: The version check on line 4,5: https://github.com/cclib/cclib/pull/491/files#diff-f57b4d3a4403d70b647003618269deeaR4

langner: No, I mean where is `conftest` used :)

langner: I see. Could you add some comment to the file (perhaps in the docstring) explaining this?

test/regression.py
```diff
@@ -1772,7 +1772,7 @@ def old_unit_test(logfile):
     return old_unit_test
 
 
-def main(which=[], opt_traceback=False, opt_status=False, regdir=__regression_dir__):
+def test_regressions(which=[], opt_traceback=False, opt_status=False, regdir=__regression_dir__):
```
langner: Why do we need to rename this?

test/data/testGeoOpt.py
```diff
@@ -44,7 +44,8 @@ def testatomnos(self):
         # This will work only for numpy
         #self.assertEquals(self.data.atomnos.dtype.char, 'i')
 
-        atomnos_types = [numpy.issubdtype(atomno,int) for atomno in self.data.atomnos]
+        atomnos_types = [numpy.issubdtype(atomno, np.signedinteger)
```
langner: This is fine, but for the future generally it's better to separate no-op formatting changes from actual changes.

test/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+]
+
+
+def match_path(path, partial_paths):
+    """Does the given path contain any of the stubs in partial_paths?"""
+    return any(partial_path in str(path)
+               for partial_path in partial_paths)
+
+
+def pytest_ignore_collect(path, config):
```
langner: I don't like this for one reason: we're moving the controls for exceptions away from the tests. How are we supposed to keep track of this, presuming we'd eventually like to get rid of these exceptions?

berquist:
This needs more work outside of the PR, so closing until it can be cleaned up with the problems fixed.

langner:
> This needs more work outside of the PR, so closing until it can be cleaned up with the problems fixed.

Could you summarize the remaining work needed in the original issue, if not already there? Or create a project...

berquist:
From highest to lowest importance:

0. There were a lot of unrelated changes that I'll just make small PRs for.
1. Output from `test_all` and the visual tests isn't displayed.
2. Something is wrong with the paths getting sent to the coverage checker, so it looks like the unit and regression tests are covering two different sets of source code.
3. Changing `main` to `test_regressions` allows pytest to automatically run regressions, but the current setup treats it as one big test. It would be nice if each file was its own individual test.

Issue #494, opened 2018-04-06
Separate psi3 parser into separate legacy code, and trim


langner:
Offshoot from #319

berquist:
I think this is an opportunity for some breaking changes regarding renames, at least of data files. This is because 4 is part of the program name, and part of the version is separate. Here is my proposal:

* Rename `Psi -> Psi4`, with an alias to `Psi`.
* The new parser is named `Psi3`.

For unit tests,
* `Psi/basicPsi4.0 -> Psi4/basicPsi4-1.0`

For regressions,
* `Psi/Psi3 -> Psi3/Psi3.4`
* `Psi/Psi4.0b5 -> Psi4/Psi4-beta5`
* `Psi/Psi4.0 -> Psi4/Psi4-0.5`

Just as an example:
* `Version 3.4 Alpha`
* `PSI 4.0.0-beta5 Driver`
* `Psi4 0.5 Driver`
* `Psi4 1.0.54 Driver`

berquist:
Done in https://github.com/cclib/cclib/pull/501 and https://github.com/cclib/cclib-data/pull/78.

PR #501, opened 2018-04-14
Psi: split into separate Psi3 and Psi4 parsers

berquist:
See https://github.com/cclib/cclib/issues/494.

See https://github.com/cclib/cclib-data/pull/78 for the renamed regression files. The regression PR must go in for this one to pass.

This also consolidates the analytic/numerical gradient parsing for Psi4 into one block.


langner:
I think there's still something to fix in regressions. Perhaps you can also separate the unrelated changes (consolidation of gradient parsing) into a different PR?

PR #502, opened 2018-04-14
Add file reader interface and XYZ reader implementation

berquist:
Attempt number 2 of https://github.com/cclib/cclib/pull/454. The previous PR was too different and large for the commits in a rebase to make sense, mostly due to the different directory structure from https://github.com/cclib/cclib/pull/498

* CJSON reader/writer code should be snake_case, even if JSON keys are
  camelCase.
* Initial XYZ file reader.
* Fix unclosed resource warnings in tests
* Replace `ccopen(source).parse() -> ccread(source)`


langner:
It seems like there are several unrelated changes here - could we break them up into smaller PRs?

PR #506, opened 2018-04-18
Adds metadata parsing for ORCA and enables metadata tests for many programs.

jevandezande:
Parses the metadata and input file (it is printed at the beginning of the orca output file).

Also enables metadata tests for other programs that were not running.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -105,7 +105,7 @@ class ORCA(logfileparser.Logfile):
         # ================================================================================
         if "INPUT FILE" == line.strip():
             self.skip_line(inputfile, '=')
-            name = next(inputfile).split()[-1]
+            self.metadata['input_file_name'] = next(inputfile).split()[-1]
 
             lines = []
             for line in inputfile:
@@ -113,7 +113,7 @@ class ORCA(logfileparser.Logfile):
                     break
                 lines.append(line[6:])
 
-            self.metadata['input_file'] = ''.join(lines[:-1])
+            self.metadata['input_file_contents'] = ''.join(lines[:-1])
             lines_iter = iter(lines[:-1])
 
             keywords = []
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -105,14 +105,23 @@ class ORCA(logfileparser.Logfile):
         # ================================================================================
         if "INPUT FILE" == line.strip():
             self.skip_line(inputfile, '=')
-            self.metadata['input_file'] = next(inputfile).split()[-1]
+            name = next(inputfile).split()[-1]
 
-            keywords = []
-            coords = []
+            lines = []
             for line in inputfile:
                 if line[0] != '|':
                     break
-                line = line[6:]
+                lines.append(line[6:])
+
+            self.metadata['input_file'] = ''.join(lines[:-1])
+            lines_iter = iter(lines[:-1])
+
+            keywords = []
+            coords = []
+            for line in lines_iter:
+                line = line.strip()
+                if not line:
+                    continue
 
                 # Keywords block
                 if line[0] == '!':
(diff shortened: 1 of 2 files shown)


cclib/parser/orcaparser.py
```diff
@@ -49,10 +49,120 @@ def before_parsing(self):
...
+                    if 'file' not in coord_type:
+                        while line[0] != '*':
+                            # Strip basis specification that can appear after coordinates
+                            line = line.split('newGTO')[0]
+                            coords.append(splitter(line))
+                            line = next(inputfile)[6:].strip()
+
+                line = next(inputfile)
+
+            self.metadata['methods'] = methods
```
berquist: I think that `keywords` would be a better name, since most entries there are not actually methods, though we don't currently have a way of separating the two. Do you have any ideas? My current idea is to do a search-and-replace after parsing with dictionaries of methods and keywords, maybe as another library...

cclib/parser/orcaparser.py
```diff
@@ -49,10 +49,120 @@ def before_parsing(self):
...
+        # |  3> ! Grid4
+        # |  4>
+        # |  5> *xyz 0 3
+        # |  6>     O   0   0   0
+        # |  7>     O   0   0   1.5
+        # |  8> *
+        # |  9>
+        # | 10>                          ****END OF INPUT****
+        # ================================================================================
+        if "INPUT FILE" == line.strip():
```
berquist: Could you also add `input_file` that contains the entire input as one string?

cclib/parser/orcaparser.py
```diff
@@ -49,10 +49,120 @@ def before_parsing(self):
...
+
+                # Keywords block
+                if line[0] == '!':
+                    methods += line[1:].split()
+
+                # Impossible to parse without knowing whether a keyword opens a new block
+                elif line[0] == '%':
+                    pass
+
+                # Geometry block
```
berquist: At first I wasn't sure how I felt about this, but I realize that the first set of displayed coordinates for many programs aren't necessarily the input coordinates: they may be symmetrized, rotated, etc. So +1.

cclib/parser/orcaparser.py
jevandezande: At first I thought you meant to store the name of the input file (which I implemented). However, upon rereading I presume you mean actually save all of the input as a string. I've corrected it to this in the latest commit.

berquist: Actually, now that you mention it, both would be good. Perhaps call it `input_file_name`? Maybe the other should then be `input_file_contents`.

test/data/testSP.py
```diff
@@ -254,20 +254,13 @@ def testmoments(self):
...
-    @skipForParser('QChem', 'Does not support metadata yet')
     def testmetadata(self):
         """Does metadata have expected keys and values?"""
         self.assertTrue(hasattr(self.data, "metadata"))
-        self.assertIn("basis_set", self.data.metadata)
+        if self.logfile.logname not in ['ORCA', 'Psi']:
+            self.assertIn("basis_set", self.data.metadata)
         self.assertIn("methods", self.data.metadata)
         self.assertIn("package", self.data.metadata)
         self.assertIn("package_version", self.data.metadata)
```
berquist: Can you add the new input file fields?

test/data/testSP.py
```diff
@@ -255,20 +255,16 @@ def testmoments(self):
...
-    @skipForParser('QChem', 'Does not support metadata yet')
     def testmetadata(self):
         """Does metadata have expected keys and values?"""
         self.assertTrue(hasattr(self.data, "metadata"))
-        self.assertIn("basis_set", self.data.metadata)
+        if self.logfile.logname not in ['ORCA', 'Psi']:
+            self.assertIn("basis_set", self.data.metadata)
+        if self.logfile.logname == 'ORCA':
+            self.assertIn("input_file_name", self.data.metadata)
+            self.assertIn("input_file_contents", self.data.metadata)
```
berquist: This isn't maintainable, so our next step can be to move metadata testing to its own file and have a test for each field.

PR #507, opened 2018-04-18
Correctly parses ORCA outputs with linear molecules

jevandezande:
Adds a GenericLinearIRTest that changes the number of vibration (linear molecules have 3*N - 5 vibrations).

Adds a CO2 vibrational test case for ORCA.


jevandezande:
The CO2 molecule linear vibrations test has been excised from this PR.

PR #509, opened 2018-05-16
initial Molcas Parser

kunalsharma05:
Currently supporting the following attributes:
1. atomnos
2. atomcharges
3. coreelectrons
4. charge
5. nbasis
6. metadata
7. atomcoords
8. natom
9. nmo
10. scfenergies

### DIFF
--- a/cclib/parser/molcasparser.py
+++ b/cclib/parser/molcasparser.py
@@ -1,14 +1,9 @@
 # -*- coding: utf-8 -*-
 #
-# This file is part of cclib (http://cclib.github.io), a library for parsing
-# and interpreting the results of computational chemistry packages.
+# Copyright (c) 2017, the cclib development team
 #
-# Copyright (C) 2014-2015, the cclib development team
-#
-# The library is free software, distributed under the terms of
-# the GNU Lesser General Public version 2.1 or later. You should have
-# received a copy of the license along with cclib. You can also access
-# the full license online at http://www.gnu.org/copyleft/lgpl.html.
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
 
 """Parser for Molcas output files"""
 


cclib/parser/molcasparser.py
```diff
@@ -0,0 +1,181 @@
...
+#
+# This file is part of cclib (http://cclib.github.io), a library for parsing
+# and interpreting the results of computational chemistry packages.
+#
+# Copyright (C) 2014-2015, the cclib development team
+#
+# The library is free software, distributed under the terms of
+# the GNU Lesser General Public version 2.1 or later. You should have
+# received a copy of the license along with cclib. You can also access
+# the full license online at http://www.gnu.org/copyleft/lgpl.html.
```
jevandezande: It looks like you put the wrong license here. It should reference the BSD 3-clause license.

cclib/parser/molcasparser.py
```diff
@@ -0,0 +1,181 @@
...
+    def __str__(self):
+        """Return a string representation of the object."""
+        return "Molcas log file %s" % (self.filename)
+
+    def __repr__(self):
+        """Return a representation of the object."""
+        return 'Molcas("%s")' % (self.filename)
+
+    def normalisesym(self, label):
+        """Does Molcas require symmetry label normalization?"""
```
jevandezande: There should be an explicit pass here (or even better, actual code)

cclib/parser/molcasparser.py
```diff
@@ -0,0 +1,181 @@
...
+            self.atomcoords.append(atomcoords)
+
+            if self.atomnos == []:
+                self.atomnos = [utils.PeriodicTable().number[atomelement] for atomelement in atomelements]
+
+            if not hasattr(self, 'natom'):
+                self.set_attribute('natom', len(self.atomnos))
+
+        ## This section is present when executing &SCF.
+        if line[0:29] == '++    Orbital specifications:':
```
jevandezande: No need for a zero, you can do `line[:29]`

cclib/parser/molcasparser.py
```diff
@@ -0,0 +1,181 @@
...
+
+                if line[6:30] == 'Total number of orbitals':
+                    self.set_attribute('nmo', int(line.split()[-1]))
+                if line[6:31] == 'Number of basis functions':
+                    self.set_attribute('nbasis', int(line.split()[-1]))
+
+                line = next(inputfile)
+
+        if line[6:23] == 'Molecular charge ':
+            self.set_attribute('charge', int(float(line.split()[-1])))
```
jevandezande: Are you sure this will round how you want it to? Calling `int` will round towards zero.

cclib/parser/molcasparser.py
```diff
@@ -0,0 +1,181 @@
...
+            self.skip_lines(inputfile, ['dashes',
+                                        'blank',
+                                        'SCF Algorithm',
+                                        'density differences',
+                                        'Number of density matrices in core',
+                                        'Maximum number of NDDO SCF iterations',
+                                        'Maximum number of HF  SCF iterations'])
+
+            line = next(inputfile)
+            assert line[6:37] == 'Threshold for SCF energy change'
```
jevandezande: This can be combined with the previous line, I presume it wasn't because you plan to actually parse this?

cclib/parser/molcasparser.py
```diff
@@ -0,0 +1,181 @@
...
+
+        ## This section is present when executing &SCF.
+        if line[0:24] == '++    Molecular charges:':
+
+            atomcharges = []
+
+            while line[6:29] != 'Total electronic charge':
+                line = next(inputfile)
+                if line[6:9] == 'N-E':
+                    atomcharges.extend(list(map(float, line.split()[1:])))
```
jevandezande: No need for explicit list cast.

cclib/parser/molcasparser.py
```diff
@@ -0,0 +1,181 @@
...
+            # Molcas only performs Mulliken population analysis.
+            self.set_attribute('atomcharges', {'mulliken': atomcharges})
+
+            # Ensure the charge printed here is identical to the
+            # charge printed before entering the SCF.
+            self.skip_line(inputfile, 'blank')
+            line = next(inputfile)
+            assert line[6:30] == 'Total            charge='
+            if hasattr(self, 'charge'):
+                assert int(float(line.split()[2])) == self.charge
```
jevandezande: Again, are you sure on the rounding?

cclib/parser/molcasparser.py
```diff
@@ -0,0 +1,181 @@
...
+    def normalisesym(self, label):
+        """Does Molcas require symmetry label normalization?"""
+
+    def before_parsing(self):
+        # Compile the regex for extracting the element symbol from the
+        # atom label in the "Molecular structure info" block.
+        self.re_atomelement = re.compile('([a-zA-Z]+)\d+')   
+
+        # Compile the dashes-and-or-spaces-only regex.
+        self.re_dashes_and_spaces = re.compile('^[\s-]+$')
```
jevandezande: There is minimal advantage to compiling regexes. Python internally compiles and caches regexes, so unless you have a huge number of them, explicitly compiling is unlikely to help. https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile

cclib/parser/molcasparser.py
kunalsharma05: Thanks!

langner: Yes, please use the same header as the other parsers: https://github.com/cclib/cclib/blob/master/cclib/parser/gaussianparser.py#L1

PR #512, opened 2018-05-18
Adding skipforMolcas on all test methods

kunalsharma05:
Just adding the skipforParser decorator on the methods. These will be removed as we keep on adding more functionality to the parser.

### DIFF
--- a/test/data/testBasis.py
+++ b/test/data/testBasis.py
@@ -34,10 +34,12 @@ class GenericBasisTest(unittest.TestCase):
     gbasis_C_2s_func0 = [2.9412, -0.1000]
     gbasis_C_2p_func0 = [2.9412, 0.1559]
 
+    @skipForParser('Molcas','The parser is still being developed so we skip this test')
     def testgbasis(self):
         """Is gbasis the right length?"""
         self.assertEquals(self.data.natom, len(self.data.gbasis))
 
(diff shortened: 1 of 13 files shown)


TDI-tenderholt:
Looks good to me so far. Will you create a separate PR for the test files?

kunalsharma05:
@TDI-tenderholt Yes, that's the plan.

berquist:
Do we need this? Tests can simply not be added to the `testdata` file.

ATenderholt:
The changes here were suggested by me (and maybe @langner) as a way to keep PRs small and targeting one change, while maintaining ability to merge directly into `master` without breaking the build. (This is almost like TDD, but instead of trying to fix failed tests, the skipped tests are being fixed.)

If tests are not added to `testdata` until that portion of the code is written, there's a chance that the PR will become fairly large before the tests pass. As a concrete example, there are 25 test cases in `testSP`. I think the current parser in #514 would fail in multiple places (`testcharge_and_mult`: no `mult`, `test_homos`: no `homos`, `testscftargetdim`: no `scftargets`, etc.).

langner:
Merging this in...

I think in the long run it would be nice to get rid of the `testdata` file altogether, and do some combination of logfile discovery and annotating the test cases even more. Keeping all the information about a test case in one place is probably a good thing. I don't know if all the skips need to be added at once, but it doesn't hurt.

Issue #435, opened 2017-09-11
Successfully completed boolean


jevandezande:
It would be nice to have a boolean indicating that a logfile contains a successfully completed job(s). Many programs indicate a successful job in some manner: ORCA prints out the timings `TOTAL RUN TIME: 0 days 0 hours 10 minutes 49 seconds 808 msec`, PSI4 asks for a beer `*** Psi4 exiting successfully. Buy a developer a beer!`. I presume other programs print something similar. I'm not sure what we would want to call such a variable, and I don't have much experience with other programs.

langner:
This is a good idea. I suggest we put it in `metadata` for now. For example, `metadata['success']`.

jevandezande:
Should this be set to `False` by default or should it not be available at all if the file terminated abnormally or is incomplete? The former is easier for the user, as they don't have to do a separate check for `success` in `metada`, but then it requires the value to be set at the beginning of `extract()`. However, no other variable seems to be set this way.

langner:
Sounds good. I think we'd like it to be `False` when a file is abruptly cut, for example.

langner:
This was finished.

PR #513, opened 2018-05-18
Adds success flag to metadata

jevandezande:
If a computation is completed successfully, sets `metadata['success'] = True` otherwise it is `False`. As discussed in #435.

Implemented  for all parsers and tested in `testGeoOpt.py`.

Question?
------
- [x] Should I test this in other places?
    Looks like we'll only test in `testGeoOpt.py`

### DIFF
--- a/cclib/parser/adfparser.py
+++ b/cclib/parser/adfparser.py
@@ -1151,3 +1151,6 @@ class ADF(logfileparser.Logfile):
                     self.polarizabilities.append(polarizability)
 
                 line = next(inputfile)
+
+        if line[:24] == ' Buffered I/O statistics':
+            self.metadata['success'] = True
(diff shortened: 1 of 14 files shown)


berquist:
> Should I test this in other places?

At first glance, no, since the geometry optimization test is called for each parser.

berquist:
For the failures, here are the actual terminating strings:

* `regression/GAMESS/PCGAMESS/dvb_gopt_b.out`: `EXECUTION OF GAMESS TERMINATED NORMALLY 11:59:47 LT  19-MAR-2006`
* `regression/Jaguar/Jaguar4.2/dvb_gopt.out`: `Job dvb_gopt completed on`, with `Total cpu seconds` above
* ``: `*** PSI4 exiting successfully. Buy a developer a beer!`

For GAMESS, `EXECUTION OF GAMESS TERMINATED NORMALLY` should work. For Psi4, `Buy a developer a beer!` (failure -> coffee). For Jaguar, probably a regex to handle `cpu` or `elapsed`.

berquist:
Great, thanks! If you have any other suggestions for what should go in `metadata`, let us know.

PR #517, opened 2018-05-28
Add file reader interface and XYZ reader implementation, take 3

berquist:
Continuing work from https://github.com/cclib/cclib/pull/454 and https://github.com/cclib/cclib/pull/502.


cclib/io/__init__.py
```diff
@@ -7,12 +7,14 @@
 
 """Contains all writers for standard chemical representations"""
 
+
```
langner: nit: maybe just one line

cclib/io/filereader.py
```diff
@@ -0,0 +1,44 @@
...
+# Copyright (c) 2017, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
+
+"""Generic file reader and related tools"""
+
+
+class Reader(object):
+    """Abstract class for reader objects."""
```
langner: Do we really want it to be abstract? If so, let's use `abc`.

cclib/io/filereader.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __init__(self, source, *args, **kwargs):
+        """Initialize the Reader object.
+
+        This should be called by a subclass in its own __init__ method.
+
+        Inputs:
+          source - A single filename, stream [TODO], or list of filenames/streams [TODO].
+        """
+
```
langner: nit: don't need the blank line here

cclib/io/xyzreader.py
```diff
@@ -0,0 +1,83 @@
...
+        #    repeat for numver of atoms
+        # (4. optional blank line)
+        # repeat for multiple sets of coordinates
+
+        all_atomcoords = []
+
+        while True:
+
+            try:
+
```
langner: nit: unneeded extra blank line

test/io/testxyzreader.py
```diff
@@ -0,0 +1,116 @@
...
+            self.assertTrue(hasattr(data, attr))
+
+        self.assertEqual(data.natom, 12)
+
+        atomnos = np.array([7, 6, 1, 8, 7, 6, 8, 6, 6, 1, 1, 1], dtype=int)
+
+        self.assertEqual(data.atomnos.shape, atomnos.shape)
+        self.assertTrue(np.all(np.equal(data.atomnos, atomnos)))
+
+        atomcoords = np.array([[
```
langner: Can the test be formulated in a way that doesn't require replicating the data?

test/io/testxyzreader.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def test_attributes_single(self):
+        """Is an XYZ file with a single geometry read into a ccData properly?
+        """
+        fpath = os.path.join(__datadir__, "test/bridge/uracil.xyz")
+        assert os.path.exists(fpath)
+        xyz = cclib.io.xyzreader.XYZ(fpath)
+        data = xyz.parse()
+
+        attrs = ('natom', 'atomnos', 'atomcoords')
```
langner: You can say the same thing on 3 lines explicitly. Also, since you're actually testing these attribute below, I don't think you need to assert their existence.

test/io/testxyzreader.py
```diff
@@ -0,0 +1,116 @@
...
+                [5.91213, 3.09904, -2.21101],
+                [2.25340, 3.12801, -1.51035],
+                [5.42415, 2.83814, -4.27067],
+                [6.99150, 3.10206, -2.32366]
+            ]
+        ])
+
+        self.assertEqual(data.atomcoords.shape, atomcoords.shape)
+        self.assertTrue(np.all(np.equal(data.atomcoords, atomcoords)))
+
```
langner: nit: maybe 2 blank lines before invocation of main?

cclib/io/filereader.py
berquist: I think that should be in a separate PR, since it needs to be done for the writers as well, along with tests for each.

test/io/testxyzreader.py
berquist: Not if you want to test for explicit correctness, no.

test/io/testxyzreader.py
```diff
@@ -0,0 +1,109 @@
...
+                [5.08640, 2.95832, -3.25044],
+                [5.91213, 3.09904, -2.21101],
+                [2.25340, 3.12801, -1.51035],
+                [5.42415, 2.83814, -4.27067],
+                [6.99150, 3.10206, -2.32366]
+            ]
+        ])
+
+        self.assertEqual(data.atomcoords.shape, atomcoords.shape)
+        self.assertTrue(np.all(np.equal(data.atomcoords, atomcoords)))
```
langner: I don't think you need to test the shape if you're checking equivalence. Also, perhaps we should use something from `numpy.testing` here?

test/io/testxyzreader.py
langner: Well, what if we wrote it back out and made sure the resulting file was the same?

test/io/testxyzreader.py
```diff
@@ -0,0 +1,109 @@
...
+        ]])
+
+        self.assertEqual(data.atomcoords.shape, atomcoords.shape)
+        self.assertTrue(np.all(np.equal(data.atomcoords, atomcoords)))
+
+    def test_attributes_two(self):
+        """Is an XYZ file with a two geometries read into a ccData properly?
+        """
+        fpath = os.path.join(__filedir__, "data/uracil_two.xyz")
+        assert os.path.exists(fpath)
```
langner: I don't think we need to assert additionally here. If the file doesn't exist, the test already fails.

test/io/testxyzreader.py
berquist: I started to write a roundtrip test, and I still think it's important to have, but I don't know that it solves this problem. The reason is because we need to fix the precision in the output file, and there are many reasons why something like the MD5 of the input and output files would differ. In that case, you could compare the generated output to a reference generated output.

Is that what you're referring to?

langner: The numbers in that example don't look weird to get mangled due to rounding or whatever. I wouldn't compare MD5's myself, just the exact strings, but I think it can be done in many ways. I do think not having the data in the actual test would be better, though. If you think it's not worth the effort at this time, that's OK with me as well.

berquist:
I added the round-trip tests against reference XYZ outputs and removed the hard-coded coordinates, but left in the atomic numbers in `testxyzreader`. Let me know if I should remove those also.

I think that only having the round-tripping tests is closer to integration testing than unit testing, so we should also have something simpler.

langner:
> I added the round-trip tests against reference XYZ outputs and removed the hard-coded coordinates, but left in the atomic numbers in testxyzreader. Let me know if I should remove those also.
> 
> I think that only having the round-tripping tests is closer to integration testing than unit testing, so we should also have something simpler

Great, I think this is a good compromise. This way, even if the the specific coordinates get changed in the file, the test won't need to be updated.

PR #529, opened 2018-06-01
Implements vibfreqs, vibrirs and vibdisps for Molcas

kunalsharma05:
Similar to the last PR. Made the requested changes.


cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+
+                if 'x' in line.split():
+                    if not hasattr(self, 'vibdisps'):
+                        self.vibdisps = []
+                    disps = []
+                    for n in range(3*self.natom):
+                        numbers = [float(s) for s in line[17:].split()]
+                        atomindex = int(re.search(r'\d+$',line.split()[0]).group()) - 1   # atom index, starting at zero
+                        numbermodes = len(numbers)
+                        if not disps:
```
ATenderholt: I didn't know this could be done. Is there a reason to prefer it over `if len(disps) == 0`?

cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+        #     C2         z   -0.03703  -0.03449  -0.07269   0.00000  -0.07416  -0.00001
+        #     C3         x   -0.00000   0.00001   0.00000  -0.06409  -0.00001   0.05110
+        #     C3         y   -0.00000   0.00001   0.00000   0.00152   0.00000  -0.03263
+        #     C3         z   -0.03808  -0.08037  -0.07267  -0.00001   0.07305   0.00000
+        if line[1:29] == 'Harmonic frequencies in cm-1':
+
+            self.skip_line(inputfile,'blank')
+            line = next(inputfile)
+
+            if 'IR' in line.split():
```
ATenderholt: Wouldn't `if 'IR' in line:` also work?

cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+                freqmode = 'IR'
+
+            while 'Thermochemistry' not in line.split():
```
ATenderholt: Why use `line.split()` instead of just `line`?

cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+
+                if 'Frequency:' in line.split():
```
ATenderholt: Similarly: why use `line.split()` instead of just `line`?

cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+                        if not hasattr(self, 'vibfreqs'):
+                            self.vibfreqs = []
+                        vibfreqs = [float(i.replace('i','-')) for i in line.split()[1:]]
+                        self.vibfreqs.extend(vibfreqs)
+
+                if 'Intensity:' in line.split():
```
ATenderholt: Similarly: why use `line.split()` instead of just `line`?

cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+                    if freqmode == 'IR':
+                        if not hasattr(self, 'vibirs'):
+                            self.vibirs = []
+                        vibirs = map(float, line.split()[1:])
+                        self.vibirs.extend(vibirs)
+
+                if 'x' in line.split():
```
ATenderholt: Similarly: why use `line.split()` instead of just `line`?

cclib/parser/molcasparser.py
kunalsharma05: Yes, it would.

cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+        #
+        #     C1         x   -0.00000   0.00000   0.00000  -0.05921   0.00000  -0.06807
+        #     C1         y    0.00001  -0.00001  -0.00001   0.00889   0.00001  -0.02479
+        #     C1         z   -0.03190   0.04096  -0.03872   0.00001  -0.12398  -0.00002
+        #     C2         x   -0.00000   0.00001   0.00000  -0.06504   0.00000  -0.03487
+        #     C2         y    0.00000  -0.00000  -0.00000   0.01045   0.00001  -0.05659
+        #     C2         z   -0.03703  -0.03449  -0.07269   0.00000  -0.07416  -0.00001
+        #     C3         x   -0.00000   0.00001   0.00000  -0.06409  -0.00001   0.05110
+        #     C3         y   -0.00000   0.00001   0.00000   0.00152   0.00000  -0.03263
+        #     C3         z   -0.03808  -0.08037  -0.07267  -0.00001   0.07305   0.00000
```
langner: Could you add (...) here and add the end snippet with 'Thermochemistry'.

cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+            self.skip_line(inputfile,'blank')
+            line = next(inputfile)
+
+            if 'IR' in line:
+                freqmode = 'IR'
+
+            while 'Thermochemistry' not in line:
+
+                if 'Frequency:' in line:
+                        if not hasattr(self, 'vibfreqs'):
```
langner: In `LogFile` there is a method called `set_attribute`, which is extremely useful but doesn't work for extending attributes like here. I was wondering whether you'd be up for udpating that method to work for extending things or adding another method for extending, whichever you think is more appropriate.

cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+                        vibirs = map(float, line.split()[1:])
+                        self.vibirs.extend(vibirs)
+
+                if 'x' in line:
+                    if not hasattr(self, 'vibdisps'):
+                        self.vibdisps = []
+                    disps = []
+                    for n in range(3*self.natom):
+                        numbers = [float(s) for s in line[17:].split()]
+                        atomindex = int(re.search(r'\d+$',line.split()[0]).group()) - 1   # atom index, starting at zero
```
langner: Consider putting the comment on the line above to keep the line length reasonable, and please make the comment a proper sentence.

cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+                        self.vibfreqs.extend(vibfreqs)
+
+                if 'Intensity:' in line:
+                    if freqmode == 'IR':
+                        if not hasattr(self, 'vibirs'):
+                            self.vibirs = []
+                        vibirs = map(float, line.split()[1:])
+                        self.vibirs.extend(vibirs)
+
+                if 'x' in line:
```
langner: Can this be more specific. There could feasibly be a word on some line in the future that contains 'x' but is different than what you expect.

cclib/parser/molcasparser.py
```diff
@@ -279,6 +279,68 @@ def extract(self, inputfile, line):
...
+                    if not hasattr(self, 'vibdisps'):
+                        self.vibdisps = []
+                    disps = []
+                    for n in range(3*self.natom):
+                        numbers = [float(s) for s in line[17:].split()]
+                        atomindex = int(re.search(r'\d+$',line.split()[0]).group()) - 1   # atom index, starting at zero
+                        numbermodes = len(numbers)
+                        if len(disps) == 0:
+                            for mode in range(numbermodes):
+                                # For each mode, make list of list [atom][coord_index]
```
langner: nit: please end sentences with periods.

cclib/parser/molcasparser.py
langner: nit: there's a space missing after a comma here

cclib/parser/molcasparser.py
kunalsharma05: Do we really need to update `set_attribute`. What are the problems associated with using built-in functions?
And if we really want to change the function, what exactly do we need it to do?

berquist: Well, we've been very inconsistent with using `set_attribute` versus `hasattr`; we don't have good standards for it. The advantage to using `set_attribute` (when not `extend`ing a quantity) is that the logger will print a warning when overwriting an already set attribute.

Also, any code written using `extend` can be rewritten to use `append` if you add an intermediate temporary list.

kunalsharma05: So maybe I can use `set_attribute` for assigning a value, but standard built-in functions for extending?

kunalsharma05:
@langner @berquist I have made most of the changes. Even added the Thermochemistry commits in this PR itself, since making a new one was mostly going to cause merge errors. All attributes that are parsed separately are in their individual commits, please take a look and let me know.

cclib/parser/molcasparser.py
langner: Well, you don't have to do this. This is fine. I'm just suggesting this would be a nice thing to have. Consider it thinking out loud :)

I would say the main advantage of having our own attribute setters is that we can change them in the future, so it is an opportunity gain in my mind.

Now that I think about it, perhaps better and more Pythonic would be to just override __setattr__ for the object we set attributes on and do all of this transparently.

cclib/parser/molcasparser.py
langner: Should be good, too. Please use your best judgement.

PR #531, opened 2018-06-11
Add basic EditorConfig file.

berquist:
See https://editorconfig.org/ for more information. Assuming your editor supports it either by default or using a plugin, these settings are automatically obeyed.

I didn't include `end_of_line = lf` because although we ultimately want that, I suspect it will automatically convert an entire Windows-terminated file once you open and save it, and if we want to fix line endings it should be done all at once as a separate change.


langner:
Created #536 to deal with line ending, since I don't think we have an issue currently open for that.

PR #532, opened 2018-06-11
Enable testing single files by passing relative paths as arguments

langner:
This makes it faster to iterate on changes to regression tests.


berquist:
Can you give an example of how to run it, so I can test it?

langner:
> Can you give an example of how to run it, so I can test it?

Excellent point. I've updated and expanded the module docstring with examples.

PR #538, opened 2018-06-13
Turbomole output files

kunalsharma05:
Turbomole output files for:
```dvb_gopt
dvb_ir
dvb_sp
dvb_sp_symm
dvb_un_sp


kunalsharma05:
@ATenderholt @langner @berquist These are the same output files as the ones we had in the older branch. Please review and let me know if there are changes needed.

berquist:
1. Why don't we merge https://github.com/cclib/cclib/tree/turbomoleparser into master? Are we fine with that history going away? Just a question.
2. The version is missing from the directory path. It looks like most files were run with 5.9, so please add that.

langner:
> Why don't we merge https://github.com/cclib/cclib/tree/turbomoleparser into master? Are we fine with that history going away? Just a question.

I'm OK with that, under the assumption that Kunal will work off of that merged code and he thinks it makes sense to do that. Keeping the history in `master` would be great.

kunalsharma05:
@langner @berquist I agree, since a lot has been implemented in that branch I would say we merge it but please take care in removing the `turbo.out` files from the outputs.

langner:
Could you also be more specific in the PR and commit description about the specifics (TM version for these files job type).

PR #539, opened 2018-06-13
Exit with status 1 when there are any failures as well

langner:
Note that this causes Travis to fail, intentionally. This should be followed by a fix for the failing Molcas unit test.


berquist:
For reference, https://github.com/cclib/cclib/pull/540/commits/047ca7fd1f6bf0a048b20a6dd7e7de2aca11afcd is a temporary fix for the failing Molcas unit test.

langner:
> For reference, 047ca7f is a temporary fix for the failing Molcas unit test.

I see. Let's get the failing tests fixed soon, of course.

PR #541, opened 2018-06-14
Parsing geotargets, geovalues correctly and enabling optdone for Molcas

kunalsharma05:
Currently we were parsing the `geotargets` and `geovalues` which included energy changes in `geovalues` and energy change threshold in `geotargets`. But since energy change threshold can be zero (Default for Gaussian method of &slapaf), which means that energy change is not considered a criteria for optimization. In such cases `testoptdone` in testGeoOpt fails as the geotarget is zero and geovalues cannot be less than zero.
In such cases we parse the `energy change threshold` target as `numpy.inf`

### DIFF
--- a/cclib/parser/molcasparser.py
+++ b/cclib/parser/molcasparser.py
@@ -11,6 +11,8 @@ from __future__ import print_function
 
 import re
 
+import numpy
+
 from cclib.parser import logfileparser
 from cclib.parser import utils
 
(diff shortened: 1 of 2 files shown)


cclib/parser/molcasparser.py
```diff
@@ -237,6 +243,14 @@ def extract(self, inputfile, line):
...
                 scftargets.append(target)
 
+            self.skip_lines(inputfile,['Threshold for linear dependence', 'Threshold at which DIIS is turned on', \
+                                         'Threshold at which QNR/C2DIIS is turned on'])
+            line = next(inputfile)
+
+            if line[6:31] == 'Threshold for Norm(delta)':
+                target = float(line.split()[-1])
+                scftargets.append(target)
+
```
kunalsharma05: @berquist @langner @ATenderholt This section is included because after properly reading the Molcas manual I got to know that there was another criteria for convergence of SCF Energy. So I included it in `scftargets` and `scfvalues`.

langner: nit: I would prefer the SCF stuff to be done in a separate change

If left here, please update the PR description to reflect this.

kunalsharma05: @langner I will make them in separate PR.

kunalsharma05:
@langner @ATenderholt I have created individual PRs for these changes and therefore I have closed this PR.

langner:
Thanks for doing that!

PR #543, opened 2018-06-16
Parsing geotargets, geovalues correctly and enabling optdone for Molcas

kunalsharma05:
Currently we were parsing the `geotargets` and `geovalues` which included energy changes in `geovalues` and energy change threshold in `geotargets`. But since energy change threshold can be zero (Default for Gaussian method of &slapaf), which means that energy change is not considered a criteria for optimization. In such cases `testoptdone` in testGeoOpt fails as the geotarget is zero and geovalues cannot be less than zero.
In such cases we parse the `energy change threshold` target as `numpy.inf`

### DIFF
--- a/cclib/parser/molcasparser.py
+++ b/cclib/parser/molcasparser.py
@@ -11,6 +11,8 @@ from __future__ import print_function
 
 import re
 
+import numpy
+
 from cclib.parser import logfileparser
 from cclib.parser import utils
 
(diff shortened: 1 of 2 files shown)


cclib/parser/molcasparser.py
```diff
@@ -500,11 +517,11 @@ def extract(self, inputfile, line):
             line_max = next(inputfile).split()
             if not hasattr(self, 'geotargets'):
                 # The attribute geotargets is an array consisting of the following
-                # values: [Energy threshold, RMS Displacements threshold, RMS Gradient threshold, \
-                #          Max Displacements threshold, Max Gradient threshold].
-                self.geotargets = [self.energy_threshold, float(line_rms[4]), float(line_rms[8]), float(line_max[4]), float(line_max[8])]
+                # values: [Energy threshold, Max Gradient threshold, RMS Gradient threshold, \
+                #          Max Displacements threshold, RMS Displacements threshold].
+                self.geotargets = [self.energy_threshold, float(line_max[8]), float(line_rms[8]), float(line_max[4]), float(line_rms[4])]
```
langner: It would help readability if you could use tmp vars with intuitive names here, instead of line_max[8], line_rms[8], and do on.

kunalsharma05:
@langner @berquist Made the requested changes. As I mentioned in #544, I think we should merge #544 before merging this PR.

PR #544, opened 2018-06-16
Parsing dnorm threshold and values into scftargets and scfvalues

kunalsharma05:
After properly reading the Molcas manual I got to know that there was another criterion for convergence of SCF Energy, `Delta norm`. So I included it in `scftargets` and `scfvalues`.

### DIFF
--- a/cclib/parser/molcasparser.py
+++ b/cclib/parser/molcasparser.py
@@ -215,6 +215,10 @@ class Molcas(logfileparser.Logfile):
         #  Threshold for SCF energy change            0.10E-08
         #  Threshold for density matrix               0.10E-03
         #  Threshold for Fock matrix                  0.15E-03
+        #  Threshold for linear dependence            0.10E-08
+        #  Threshold at which DIIS is turned on       0.15E+00
+        #  Threshold at which QNR/C2DIIS is turned on 0.75E-01
+        #  Threshold for Norm(delta) (QNR/C2DIIS)     0.20E-04
         if line[:34] == '++    Optimization specifications:':
 
             scftargets = []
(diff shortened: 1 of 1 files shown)


cclib/parser/molcasparser.py
```diff
@@ -237,6 +241,14 @@ def extract(self, inputfile, line):
                 target = float(line.split()[-1])
                 scftargets.append(target)
 
+            self.skip_lines(inputfile,['Threshold for linear dependence', 'Threshold at which DIIS is turned on', \
+                                         'Threshold at which QNR/C2DIIS is turned on'])
+            line = next(inputfile)
+
+            if line[6:31] == 'Threshold for Norm(delta)':
+                target = float(line.split()[-1])
+                scftargets.append(target)
```
langner: Is there a unit test that you can update to reflect the changes in `scftargets` and `scfvalues`?

kunalsharma05: I don't think that there is a test that would reflect this change. We were earlier parsing 3 targets (or thresholds)
```
scftargets = [Threshold for SCF energy change, Threshold for density matrix, \
Threshold for Fock matrix]
```
And similarly we were parsing the respective `scfvalues`.

Now we are similarly parsing 4 such `scftargets` and respective `scfvalues`.

There are tests that checks for dimensions of `geotargets` and `geovalues`,
```python
    def testscftargetdim(self):
        """Do the scf targets have the right dimensions?"""
        self.assertEquals(self.data.scftargets.shape, (len(self.data.scfvalues), len(self.data.scfvalues[0][0]))) # equals 4
```
But there is no test that actually checks the `scftargets` and `scfvalues` since they differ for each software.

kunalsharma05:
@langner @berquist @ATenderholt I think that we should merge this PR before #543 because that PR has deletions which might cause errors when we try to merge this one.

cclib/parser/molcasparser.py
berquist: @langner Is there a problem then? I don't understand.

kunalsharma05: No, no problem. I am just saying we don't have a test to actually check the values of `scftargets` and `scfvalues` since they are different for each parser and even for each job (Since `scftargets` can be specified by the user).

ATenderholt: @langner Does #545 satisfy your concern about the unit test here?

langner: Yes, I think the new tests in #545 are the way to go. Thanks!

PR #545, opened 2018-06-19
Added test to check number of SCF targets for each software

kunalsharma05:
Each software has different criteria which need to be satisfied in order for the `SCF Energy value` to be converged. This test checks the whether the parsed values of `scftargets[0]` is equal to the number of SCF criteria (`num_scf_criteria`). E.g. Molcas has 4 SCF convergence criteria so we check if 4 values are being parsed into `scftargets`.


test/data/testSP.py
```diff
@@ -151,6 +154,10 @@ def testscftargetdim(self):
         """Do the scf targets have the right dimensions?"""
         self.assertEquals(self.data.scftargets.shape, (len(self.data.scfvalues), len(self.data.scfvalues[0][0])))
 
+    def testscftargets(self):
+        """Are correct number of SCF convergence criteria being parsed?"""
+        self.assertEquals(len(self.data.scftargets[0]), self.num_scf_criteria)
```
langner: You could check `scftargets[i]` for all possible `i`, no?

kunalsharma05: In SP calculations we get `len(scftargets) = 1` since we only have one iteration of SCF energy calculation. So `i = 1` and we only need to check once.

langner: Duh, this is for SP. Perhaps check the length for good form.

kunalsharma05: Yeah. That would make sense.

PR #549, opened 2018-06-25
Implementing initial turbomoleparser

kunalsharma05:
Added functionality to support:
```
natom
nbasis
nmo
atomcoords
```


langner:
So does this make any of the unit tests pass?

Issue #551, opened 2018-06-26
Testing directory should not be installed


berquist:
The fix presented in https://github.com/cclib/cclib/issues/527 for https://github.com/cclib/cclib/pull/548 causes the `tests/` directory to be installed.

PR #552, opened 2018-06-26
Don't install the test directory

berquist:
Fixes https://github.com/cclib/cclib/issues/551 due to silly typo.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ def setup_cclib():
         long_description="\n".join(doclines[2:]),
         classifiers=classifiers.split("\n"),
         platforms=["Any."],
-        packages=setuptools.find_packages(exclude=['tests']),
+        packages=setuptools.find_packages(exclude=['*test*']),
         entry_points={
             'console_scripts': [
                 'ccget=cclib.scripts.ccget:ccget',


berquist:
Btw, @langner, why do you have your commit (`.gitconfig`?) tool set up to do "berquist authored and langner committed"? I think we should have explicit merge commits, otherwise we lose that branch point, and the commit hashes get overwritten.

langner:
> Btw, @langner, why do you have your commit (.gitconfig?) tool set up to do "berquist authored and langner committed"? I think we should have explicit merge commits, otherwise we lose that branch point, and the commit hashes get overwritten.

Well, that's because I choose "rebase and merge" instead of "create a merge commit" for cclib on GitHub. Local defaults are irrelevant.

There's a tradeoff here, and in our case I think a linear commit history is more valuable. In the past we always merged and the commit history was a pain to read ("train tracks") because people did not sync/rebase their branches locally. And we don't have that many PRs/contributors to make branch traceability an issue.

I'm not opposed to doing explicit merges as a rule, as it does carry more information and makes tracing/debugging a little easier (although I've never benefited from this in cclib). If we do make a policy like that, let's write it down in the docs ("developer guidelines" or similar). Also, I would ask in that case that we expect contributors to rebase locally on top of a reasonably recent commit in `master`, so that the train track effect is minimzed.

@ATenderholt thoughts?

langner:
Starting the conversation about dev guidelines in #555

berquist:
I shouldn't have asked about that here, so thanks for opening the issue. At least for the time being I'm fine with how things are working.

In any case, can you confirm that this works with your setup? `pip install -e .` and `pip install .` now seem to be fixed.

langner:
Confirmed that `test` is not installed anymore.

PR #553, opened 2018-06-26
fix python3-incompatible prints in turbomole parser

grzegorzmazur:


berquist:
Hi Grzegorz, thanks for the PR! Normally, I would accept this, because of course we want Python 3 compatibility, but this file is a bit of a special case. We have a Google Summer of Code student (@kunalsharma05) working on the Turbomole parser right now, and he's probably going to rewrite most of it from scratch. Currently, I don't think it's usable at all, or connected to the rest of the code, so any installation errors you see won't affect other functionality.

I'll leave this PR open until the installation errors under 3.x go away.

grzegorzmazur:
Hi Eric,

I'm just working on our old parser to the new cclib, planning to submit it upstream shortly, and the turbomole parser fails during was the first obstacle I've encountered, that's why I've submitted the fix as a stop-gap measure. But OK, it's a simple fix which I can keep applied locally for now.

Thanks,
Grzesiek

langner:
It seems the problem will go away with #554

berquist:
Yes, this is now fixed.

PR #554, opened 2018-06-27
Initial Turbomole Parser

kunalsharma05:
In this PR the diff is a little bit cleaner. The old code has been put in a temporary class while the new one implements the following attributes:
```
1. natom
2. nbasis
3. nmo
4. atomcoords
```

### DIFF
--- a/cclib/parser/turbomoleparser.py
+++ b/cclib/parser/turbomoleparser.py
@@ -1,20 +1,20 @@
-"""
-cclib (http://cclib.sf.net) is (c) 2006, the cclib development team
-and licensed under the LGPL (http://www.gnu.org/copyleft/lgpl.html).
-"""
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2018, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
 
-__revision__ = "$Revision: 635 $"
+"""Parser for Turbomole output files."""
+
+from __future__ import print_function
 
 import re
 
-# If numpy is not installed, try to import Numeric instead.
-try:
-    import numpy
-except ImportError:
-    import Numeric as numpy
+import numpy
 
-import utils
-import logfileparser
+from cclib.parser import logfileparser
+from cclib.parser import utils
 
 class AtomBasis:
     def __init__(self, atname, basis_name, inputfile):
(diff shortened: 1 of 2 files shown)


cclib/parser/turbomoleparser.py
```diff
@@ -536,5 +536,127 @@ def after_parsing(self):
...
+        #     0.92683848   -0.00007460    2.49592179    c      3    6.000    0     0
+        #     2.69176331   -0.00007127    0.44712612    c      3    6.000    0     0
+        #     1.69851645   -0.00007331   -2.06488947    c      3    6.000    0     0
+        #...
+        #    -7.04373606    0.00092244    2.74543891    h      1    1.000    0     0
+        #    -9.36352819    0.00017229    0.07445322    h      1    1.000    0     0
+        #    -0.92683849   -0.00007461   -2.49592179    c      3    6.000    0     0
+        #    -1.65164853   -0.00009927   -4.45456858    h      1    1.000    0     0
+        if 'Atomic coordinate, charge and isotop information' in line:
+            self.skip_lines(inputfile,['border','b','b'])
```
langner: There should normally be [a space after commas](https://stackoverflow.com/questions/31741418/pep8-space-after-a-comma).

cclib/parser/turbomoleparser.py
```diff
@@ -536,5 +536,127 @@ def after_parsing(self):
...
+                line = next(inputfile)
+               
+                while len(line) > 2:
+                    atomnos.append(line.split()[3].upper())
+                    atomcoords.append([utils.convertor(float(x), "bohr", "Angstrom") for x in line.split()[0:3]])
+                    line = next(inputfile)
+
+                self.atomcoords.append(atomcoords)
+
+                if self.atomnos == []:
```
langner: Why this reluctance to not overwrite the attribute?

Also, would it be possible to use `self.setattr` in this block?

cclib/parser/turbomoleparser.py
```diff
@@ -536,5 +536,127 @@ def after_parsing(self):
...
+        #    1.69851645303760     -0.00007330575699     -2.06488946767951      c
+        #...
+        #   -7.04373605585274      0.00092243787879      2.74543890990978      h
+        #   -9.36352819434217      0.00017228707094      0.07445321997922      h
+        #   -0.92683848797451     -0.00007460625018     -2.49592178685491      c
+        #   -1.65164852640697     -0.00009927259342     -4.45456857763148      h
+        #$redundant
+        if line[0:6] == "$coord":
+            if '$user' not in next(inputfile) and '$coordinate' not in line:
+                if not hasattr(self, 'atomcoords'):
```
langner: Again, why the check? I would say either overwrite the coordinates or at least check that they haven't changed.

kunalsharma05: Here we are just checking whether there are initial coordinates (given as a part of the input) in the `job.last` file. If they are found then we don't need the ones from the `coord` file. But if for some reason they are not found in the `job.last` file then we pick them up from the `coord` file.

cclib/parser/turbomoleparser.py
```diff
@@ -536,5 +536,127 @@ def after_parsing(self):
...
+        #   -0.92683848797451     -0.00007460625018     -2.49592178685491      c
+        #   -1.65164852640697     -0.00009927259342     -4.45456857763148      h
+        #$redundant
+        if line[0:6] == "$coord":
+            if '$user' not in next(inputfile) and '$coordinate' not in line:
+                if not hasattr(self, 'atomcoords'): 
+                    self.atomcoords = []
+                    if not hasattr(self, 'atomnos'):
+                        self.atomnos = []
+
```
kunalsharma05: @langner @ATenderholt Here, the variable line has value, `line = '$coord'`. You can check it using print.

cclib/parser/turbomoleparser.py
```diff
@@ -536,5 +536,127 @@ def after_parsing(self):
...
+                    atomcoords = []
+                    atomnos = []
+                    line = next(inputfile)
```
kunalsharma05: @langner @ATenderholt  But using `line = next(inputfile)`, it skips over, `   -2.69176330280845     -0.00007129445712     -0.44712612093731      c`
and makes `line =    -1.69851644615005     -0.00007332282157      2.06488947265450      c`.
Which causes the `atomcoords` and `atomnos` to be parsed wrongly

cclib/parser/turbomoleparser.py
```diff
@@ -536,5 +536,127 @@ def after_parsing(self):
                     i -= 1
                 i += 1
 
-#EOF
+
+# Implementation for the new code begins here.
```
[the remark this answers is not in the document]
kunalsharma05: This will cause the diff to look a little ugly. Still I have made the changes to reflect this and other issues as well.

cclib/parser/turbomoleparser.py
```diff
@@ -45,11 +45,121 @@ def parse_basis(self, inputfile):
             self.coefficients.append(coeff_arr)
             line=inputfile.next()
 
+# Implementation for the new code begins here.
```
langner: I don't think you need this comment.

cclib/parser/turbomoleparser.py
```diff
@@ -45,11 +45,121 @@ def parse_basis(self, inputfile):
...
+                        atomcoords.append([utils.convertor(float(x), "bohr", "Angstrom") for x in line.split()[0:3]])
+                        line = next(inputfile)
+
+                    self.atomcoords.append(atomcoords)
+                    self.set_attribute(atomnos, [utils.PeriodicTable().number[atomelement] for atomelement in atomnos])
+
+
+    def after_parsing(self):
+        pass
+
```
langner: Normally there are two blank lines between classes in Python.

cclib/parser/turbomoleparser.py
```diff
@@ -45,11 +45,121 @@ def parse_basis(self, inputfile):
...
+                line = next(inputfile)
+                if '$user' not in line: 
+                    atomcoords = []
+                    atomnos = []
+                    while line[0] != "$":
+                        atomnos.append(line.split()[3].upper())
+                        atomcoords.append([utils.convertor(float(x), "bohr", "Angstrom") for x in line.split()[0:3]])
+                        line = next(inputfile)
+
+                    self.atomcoords.append(atomcoords)
```
langner: Another typical thing we do it accumulate the coordinates (or other elements of arrays) in the loop, and then set the attribute using `set_attribute` at the very end.

cclib/parser/turbomoleparser.py
kunalsharma05: Added an extra line.

cclib/parser/turbomoleparser.py
langner: I think our modus operandi has been to parse every wherever available, even it was parsed before and we don't use it. Besides being able to check for consistency in the output, this way we make sure the parser code is not broken for rarely visited branches. Adding this to #555 so it end up in our docs.

cclib/parser/turbomoleparser.py
```diff
@@ -46,10 +46,118 @@ def parse_basis(self, inputfile):
...
+        #    0.92683848474542     -0.00007460039817      2.49592179180606      c
+        #    2.69176330586455     -0.00007127328394      0.44712611937145      c
+        #    1.69851645303760     -0.00007330575699     -2.06488946767951      c
+        #...
+        #   -7.04373605585274      0.00092243787879      2.74543890990978      h
+        #   -9.36352819434217      0.00017228707094      0.07445321997922      h
+        #   -0.92683848797451     -0.00007460625018     -2.49592178685491      c
+        #   -1.65164852640697     -0.00009927259342     -4.45456857763148      h
+        #$redundant
+        if line[0:6] == "$coord":
```
langner: In these situations, it's probably better to use `line.startswith` to avoid simple counting mistakes. Also more readable.

cclib/parser/turbomoleparser.py
```diff
@@ -46,10 +46,118 @@ def parse_basis(self, inputfile):
...
+            if '$coordinate' not in line:
+                line = next(inputfile)
+                if '$user' not in line: 
+                    atomcoords = []
+                    atomnos = []
+                    while line[0] != "$":
+                        atomnos.append(line.split()[3].upper())
+                        atomcoords.append([utils.convertor(float(x), "bohr", "Angstrom") for x in line.split()[0:3]])
```
langner: You probably already know this, but when slicing `line.split()[:3]` is equivalent to `line.split()[0:3]`. It's good to get in the habit of using the former. More good details about slicing: https://stackoverflow.com/questions/509211/understanding-pythons-slice-notation

cclib/parser/turbomoleparser.py
```diff
@@ -46,10 +46,118 @@ def parse_basis(self, inputfile):
...
+                        line = next(inputfile)
+
+                    self.set_attribute('atomcoords', atomcoords)
+                    self.set_attribute('atomnos', [utils.PeriodicTable().number[atomelement] for atomelement in atomnos])
```
langner: Consider doing the string-to-int conversion inside the loop, so that this part is purely about attribute setting boilerplate.

cclib/parser/turbomoleparser.py
```diff
@@ -114,7 +222,7 @@ def extract(self, inputfile, line):
             homos=int(temp[1])
 
         if line[0:6] == "$basis":
-            print "Found basis"
+            print("Found basis")
```
langner: Is this left over debugging code?

kunalsharma05: Had to make this change, since the parser kept failing tests for python 3.

Issue #335, opened 2016-09-09
Refactor parsers


ATenderholt:
The ```extract``` methods of the parsers have become unruly now that we support so many attributes, so it would be good to refactor these into smaller chunks. One possibility would be to take advantage of decorators which would allow all wrapped functions to automatically update progress, have a method that returns which attributes are actually supported and any parser-specific doc-strings, and ability to parse just the attributes one cares about instead of whole file.

I've pushed some code to my repo for this decorator and how it would be used: https://github.com/ATenderholt/cclib/commit/59453355360e3a682f10160778507cfdbfe646a3

Here's an example:

```
>>> from cclib.parser import Gaussian
>>> Gaussian.get_supported_attributes().keys()
dict_keys(['newattr'])  # the only attribute parsed with decorated function is newattr
>>> Gaussian.get_supported_attributes()["newattr"].__name__
'parsing_test'             # the function is called 'parsing_test'
>>> Gaussian.get_supported_attributes()["newattr"].__doc__
'Parses newattr.'        # and here's its docstring
>>> parser = Gaussian("nonexistant.log")
>>> parser.fupdate = 0.5  # needed in this demo because fupdate isn't set unless parse() is called
>>> parser.parsing_test("some good line", None)  # lets parse a line that should actually enter this function; None is because I don't have an associated inputfile in this example.
some good line
>>> parser.parsing_test("some bad line", None)  # here's a block we should skip over
>>>
```

Only downside I see is the overhead of calling into the decorators, but I think simplified code and methods to help generate documentation would make it worth it.

langner:
We definitely need to break up the parsers into smaller pieces in some way. This is an interesting idea.

ATenderholt:
I had thought about the simple solution of just naming functions ```parse_*```, but some of the blocks caught by the if-statements find more than a single attribute.

After looking at the Gaussian parser a bit more, the predicate parsed to the decorator should probably also allow an argument list or keywords which complicates things slightly.

langner:
Perhaps stacking multiple decorators is an option?

ATenderholt:
Stacking multiple decorators would definitely make the predicate argument shorter, but I'm not sure it simplifies things that much or we'd need a slew of decorators. For instance, here's an entry point:

```
if self.BOMD and line.startswith(' Summary information for step')
```

I could imagine the following:

```
# decorators get specified in reverse direction of when they are applied, right?
@parseattribute.parser(lambda attr: attr, "object attribute name")
[...]

PR #561, opened 2018-07-06
Psi4: consolidate gradient parsing

berquist:
The only difference between an analytic and numerical gradient calculation is the number of blank lines, so remove all the duplicate code.

The numerical gradient regression test is also slightly more robust.

### DIFF
--- a/cclib/parser/psi4parser.py
+++ b/cclib/parser/psi4parser.py
@@ -7,6 +7,8 @@
 
 """Parser for Psi4 output files."""
 
+from collections import namedtuple
+
 import numpy
 
 from cclib.parser import data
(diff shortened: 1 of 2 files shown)


cclib/parser/psi4parser.py
```diff
@@ -40,6 +40,17 @@ def before_parsing(self):
         # with changes triggered by ==> things like this <== (Psi3 does not have this)
         self.section = None
 
+        # Match the number of skipped lines required based on the type
+        # of gradient present (determined from the header), as
+        # otherwise the parsing is identical.
+        self.gradient_headers = {
```
langner: Why not make this one dictionary with analytic/numerical as the keys, or perhaps a namedtuple? I don't think we need to care about the performance here, so looping through a list/keys is fine.

cclib/parser/psi4parser.py
```diff
@@ -40,6 +40,17 @@ def before_parsing(self):
         # with changes triggered by ==> things like this <== (Psi3 does not have this)
         self.section = None
 
+        # Match the number of skipped lines required based on the type
```
langner: Why not make these class or module constants?

cclib/parser/psi4parser.py
```diff
@@ -824,11 +820,9 @@ def get_normalization_factor(exp, lx, ly, lz):
         #     2     0.00000000000000    -0.00979709321487     0.01460651641258
         #     3     0.00000000000000     0.00979709321487     0.01460651641258
 
-        if line.strip() == '## F-D gradient (Symmetry 0) ##':
-            next(inputfile)
-            self.skip_lines(inputfile, ['b'])
-            next(inputfile)
-            self.skip_lines(inputfile, ['b'])
+        if line.strip() in self.gradient_headers:
```
langner: I think reducing line count is great, but using the vars instead of strings here makes the parser less readable. Do you think it makes sense to put the parsing code in a helper method and call it differently in the two cases?

berquist:
> Why not make these class or module constants?

Done. I'm not sure if I put them in the best location (just before `extract()`).

> I think reducing line count is great, but using the vars instead of strings here makes the parser less readable. Do you think it makes sense to put the parsing code in a helper method and call it differently in the two cases?

The helper method is not difficult, but it doesn't solve the fundamental issue of using class constants versus keeping the data close to the parsing code. Passing in the `skip_lines` directly to the method would mean that data can't be reused somewhere else, if that matters at all. I'm not sure what's more important and could be swayed either way.

> Why not make this one dictionary with analytic/numerical as the keys, or perhaps a namedtuple? I don't think we need to care about the performance here, so looping through a list/keys is fine.

If you check the last commit (https://github.com/cclib/cclib/pull/561/commits/d900e73c0d4dcaddac78191fe1ed06a11f8c08f5), you can see the difference between a very compact representation (a single dict to map between header and lines to skip) and a more general one based on namedtuples. I don't like the extra logic above the gradient parse call, but using a simple class to hold metadata may be useful for other things. What do you think?

A lot of this depends on whether or not storing "triggers" for reuse is desirable.

langner:
I dunno... I'm not happy with how the code reads now. It's jumping all over and the parsing flow is not obvious with that list in there. But I'm just going to merge in the interest of moving forward. This is a good case to consider when designing how parsers are supposed to work (#335). Of course, that refactoring can happen before 2.x but it'd be good to tie the two topics..

PR #563, opened 2018-07-10
Turbomole: Added implementation for frequency attributes

kunalsharma05:
**1. Added the implementation for following attributes:**
```
1. vibfreqs
2. vibdisps
3. vibirs
4. vibsyms
```
**2. Removed skipfor from the following tests:**
```
1. testvibdisps
2. testlengths
3. testfreqval
4. testirintens
```

### DIFF
--- a/cclib/parser/turbomoleparser.py
+++ b/cclib/parser/turbomoleparser.py
@@ -166,58 +166,62 @@ class Turbomole(logfileparser.Logfile):
         #   1   c           x   0.00000  0.00001  0.00000 -0.01968 -0.04257  0.00001
         #                   y  -0.08246 -0.08792  0.02675 -0.00010  0.00000  0.17930
         #                   z   0.00001  0.00003  0.00004 -0.10350  0.11992 -0.00003
-        if line[5:14] == 'frequency':
-            if not hasattr(self, 'vibfreqs'):
-                self.vibfreqs = []
+        if 'NORMAL MODES and VIBRATIONAL FREQUENCIES (cm**(-1))' in line:
+            vibfreqs, vibsyms, vibirs, vibdisps = [], [], [], []
+            while '****  force : all done  ****' not in line:
+                if line.strip().startswith('frequency'):
+                    freqs = [float(i.replace('i', '-')) for i in line.split()[1:]]
+                    vibfreqs.extend(freqs)
+                    self.skip_line(inputfile, ['b'])
+                    line = next(inputfile)
+                    if line.strip().startswith('symmetry'):
+                        syms = line.split()[1:]
+                        vibsyms.extend(syms)
 
-            if not hasattr(self, 'vibdisps'):
-                self.vibdisps = []
+                    self.skip_lines(inputfile, ['b', 'IR', 'dQIP'])
+                    line = next(inputfile)
+                    if line.strip().startswith('intensity (km/mol)'):
+                        irs = [self.float(f) for f in line.split()[2:]]
+                        vibirs.extend(irs)
 
-            vibfreqs = [float(i.replace('i', '-')) for i in line.split()[1:]]
-            self.vibfreqs.extend(vibfreqs)
-            self.skip_line(inputfile, ['b'])
-            line = next(inputfile)
-            if line[5:13] == 'symmetry':
-                if not hasattr(self, 'vibsyms'):
-                    self.vibsyms = []
+                    self.skip_lines(inputfile, ['intensity', 'b', 'raman', 'b'])
+                    line = next(inputfile)
+                    x, y, z = [], [], []
+                    while line.split():
+                        x.append([float(i) for i in line.split()[3:]])
+                        line = next(inputfile)
+                        y.append([float(i) for i in line.split()[1:]])
+                        line = next(inputfile)
+                        z.append([float(i) for i in line.split()[1:]])
+                        line = next(inputfile)
 
-                vibsyms = line.split()[1:]
-                self.vibsyms.extend(vibsyms)
+                    for j in range(len(x[0])):
+                        disps = []
+                        for i in range(len(x)):
+                            disps.append([x[i][j], y[i][j], z[i][j]])
+                        vibdisps.append(disps)
 
-            self.skip_lines(inputfile, ['b', 'IR', 'dQIP'])
-            line = next(inputfile)
-            if line.split()[0] == 'intensity':
-                if not hasattr(self, 'vibirs'):
-                    self.vibirs = []
-
-                vibirs = [self.float(f) for f in line.split()[2:]]
-                self.vibirs.extend(vibirs)
-
-            self.skip_lines(inputfile, ['intensity', 'b', 'raman', 'b'])
-            line = next(inputfile)
-            x, y, z = [], [], []
-            while line.split() != []:
-                x.append([float(i) for i in line.split()[3:]])
                 line = next(inputfile)
-                y.append([float(i) for i in line.split()[1:]])
-                line = next(inputfile)
-                z.append([float(i) for i in line.split()[1:]])
-                line = next(inputfile)
 
-            for j in range(len(x[0])):
-                vibdisps = []
-                for i in range(len(x)):
-                    vibdisps.append([x[i][j], y[i][j], z[i][j]])
-                self.vibdisps.append(vibdisps)
+            self.set_attribute('vibfreqs', vibfreqs)
+            self.set_attribute('vibsyms', vibsyms)
+            self.set_attribute('vibirs', vibirs)
+            self.set_attribute('vibdisps', vibdisps)
+
+    def deleting_modes(self, vibfreqs, vibdisps, vibirs):
+        """Deleting frequencies relating to translations or rotations"""
+        i = 0
+        while i < len(vibfreqs):
+            if vibfreqs[i] == 0.0:
+                # Deleting frequencies that have value 0 since they
+                # do not correspond to vibrations.
+                del vibfreqs[i], vibdisps[i], vibirs[i]
+                i -= 1
+            i += 1
 
     def after_parsing(self):
         if hasattr(self, 'vibfreqs'):
-            i = 0
-            while i < len(self.vibfreqs):
-                if self.vibfreqs[i] == 0.0:
-                    del self.vibfreqs[i], self.vibdisps[i], self.vibirs[i]
-                    i -= 1
-                i += 1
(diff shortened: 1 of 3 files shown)


cclib/parser/turbomoleparser.py
```diff
@@ -149,9 +149,75 @@ def extract(self, inputfile, line):
...
+        # |dDIP/dQ|   (a.u.)     0.0002   0.0000   0.0005   0.0004   0.0000   0.0000
+        # intensity (km/mol)       0.05     0.00     0.39     0.28     0.00     0.00
+        # intensity (  %   )       0.05     0.00     0.40     0.28     0.00     0.00
+        #
+        #        RAMAN             YES      YES      YES      YES      YES      YES
+        #
+        #   1   c           x   0.00000  0.00001  0.00000 -0.01968 -0.04257  0.00001
+        #                   y  -0.08246 -0.08792  0.02675 -0.00010  0.00000  0.17930
+        #                   z   0.00001  0.00003  0.00004 -0.10350  0.11992 -0.00003
+        if line[5:14] == 'frequency':
```
langner: It would be more specific, and perhaps more prudent, to trigger this section with "Frequency value in ..." instead of the more generic "frequency". What do you think?

cclib/parser/turbomoleparser.py
```diff
@@ -149,9 +149,75 @@ def extract(self, inputfile, line):
...
+            if not hasattr(self, 'vibfreqs'):
```
langner: Can you do the tmp var & self.set_attribute dance here?

cclib/parser/turbomoleparser.py
```diff
@@ -149,9 +149,75 @@ def extract(self, inputfile, line):
...
+                self.vibfreqs = []
+
+            if not hasattr(self, 'vibdisps'):
+                self.vibdisps = []
+
+            vibfreqs = [float(i.replace('i', '-')) for i in line.split()[1:]]
+            self.vibfreqs.extend(vibfreqs)
+            self.skip_line(inputfile, ['b'])
+            line = next(inputfile)
+            if line[5:13] == 'symmetry':
```
langner: Is there a reason we need to use indices here, and not something more flexible like line.strip().startswith?

cclib/parser/turbomoleparser.py
```diff
@@ -149,9 +149,75 @@ def extract(self, inputfile, line):
...
+                if not hasattr(self, 'vibirs'):
+                    self.vibirs = []
+
+                vibirs = [self.float(f) for f in line.split()[2:]]
+                self.vibirs.extend(vibirs)
+
+            self.skip_lines(inputfile, ['intensity', 'b', 'raman', 'b'])
+            line = next(inputfile)
+            x, y, z = [], [], []
+            while line.split() != []:
```
langner: Is there an advantage of comparing to a list instead of just `line.strip()`?

cclib/parser/turbomoleparser.py
```diff
@@ -149,9 +149,75 @@ def extract(self, inputfile, line):
...
+
+            for j in range(len(x[0])):
+                vibdisps = []
+                for i in range(len(x)):
+                    vibdisps.append([x[i][j], y[i][j], z[i][j]])
+                self.vibdisps.append(vibdisps)
 
     def after_parsing(self):
-        pass
+        if hasattr(self, 'vibfreqs'):
```
langner: Could you perhaps add a comment describing what is being done here? Or better yet, add a little function that is responsible for it.

cclib/parser/turbomoleparser.py
kunalsharma05: I did this so that each time we encountered frequency we would get to parse the relevant information. If we used some other trigger we would have to use a while loop and the would need a loop ending trigger which I am not very sure about in turbomole.

cclib/parser/turbomoleparser.py
kunalsharma05: @langner Yeah, but again for this we will need a trigger that can enter/end a loop over the frequency section and then `set_attribute` the parsed data of the temp variable into the main thing. Otherwise we will have to create a tmp variable before the if condition and in case we don''t have a frequency section we will just have a initialized unused array.

cclib/parser/turbomoleparser.py
kunalsharma05: Fair point, will change.

cclib/parser/turbomoleparser.py
kunalsharma05: Alright!

cclib/parser/turbomoleparser.py
ATenderholt: If `extract()` is refactored in the future, how would this part of the code be handled?

cclib/parser/turbomoleparser.py
```diff
@@ -149,9 +149,82 @@ def extract(self, inputfile, line):
...
+                    x, y, z = [], [], []
+                    while line.split():
+                        x.append([float(i) for i in line.split()[3:]])
+                        line = next(inputfile)
+                        y.append([float(i) for i in line.split()[1:]])
+                        line = next(inputfile)
+                        z.append([float(i) for i in line.split()[1:]])
+                        line = next(inputfile)
+
+                    for j in range(len(x[0])):
```
langner: Is this equivalent to `vibdisps.append(zip(x, y, z))`?

cclib/parser/turbomoleparser.py
```diff
@@ -149,9 +149,82 @@ def extract(self, inputfile, line):
...
+        # intensity (km/mol)       0.05     0.00     0.39     0.28     0.00     0.00
+        # intensity (  %   )       0.05     0.00     0.40     0.28     0.00     0.00
+        #
+        #        RAMAN             YES      YES      YES      YES      YES      YES
+        #
+        #   1   c           x   0.00000  0.00001  0.00000 -0.01968 -0.04257  0.00001
+        #                   y  -0.08246 -0.08792  0.02675 -0.00010  0.00000  0.17930
+        #                   z   0.00001  0.00003  0.00004 -0.10350  0.11992 -0.00003
+        if 'NORMAL MODES and VIBRATIONAL FREQUENCIES (cm**(-1))' in line:
+            vibfreqs = []
```
langner: You can initialize all of these on one line.

cclib/parser/turbomoleparser.py
```diff
@@ -149,9 +149,82 @@ def extract(self, inputfile, line):
...
+            vibsyms = []
+            vibirs = []
+            vibdisps = []
+            while '****  force : all done  ****' not in line:
+                if line.strip().startswith('frequency'):
+                    freqs = [float(i.replace('i','-')) for i in line.split()[1:]]
```
berquist: Here and in other places, there's a missing space after each comma (PEP 8). Using a linter like `pylint` will help find these.

cclib/parser/turbomoleparser.py
```diff
@@ -149,9 +149,82 @@ def extract(self, inputfile, line):
...
+            self.set_attribute('vibfreqs',vibfreqs)
+            self.set_attribute('vibsyms',vibsyms)
+            self.set_attribute('vibirs',vibirs)
+            self.set_attribute('vibdisps',vibdisps)
+
+    def deleting_modes(self, vibfreqs, vibdisps, vibirs):
+        """Deleting frequencies relating to translations or rotations"""
+        i = 0
+        while i < len(vibfreqs):
+            if vibfreqs[i]==0.0:
```
berquist: Ditto PEP 8.

cclib/parser/turbomoleparser.py
```diff
@@ -149,9 +149,82 @@ def extract(self, inputfile, line):
...
+        #
+        #   1   c           x   0.00000  0.00001  0.00000 -0.01968 -0.04257  0.00001
+        #                   y  -0.08246 -0.08792  0.02675 -0.00010  0.00000  0.17930
+        #                   z   0.00001  0.00003  0.00004 -0.10350  0.11992 -0.00003
+        if 'NORMAL MODES and VIBRATIONAL FREQUENCIES (cm**(-1))' in line:
+            vibfreqs = []
+            vibsyms = []
+            vibirs = []
+            vibdisps = []
+            while '****  force : all done  ****' not in line:
```
langner: Using `while` in this capacity can be problematic, although I'm not strongly opposed. An alternative and maybe more dependable stopping condition might be to count the number of frequencies/displacements parsed. We know how many there should be, right? You could also exit early if this line is there for some reason and not all were printed.

cclib/parser/turbomoleparser.py
kunalsharma05: Yeah, my bad. I have started using it.

cclib/parser/turbomoleparser.py
kunalsharma05: Well, we can stop once we get the number of frequencies. But the thing is that Turbomole prints frequencies with value 0.0 as well, which get parsed and need to be deleted later and all the corresponding `vibdisps` and `vibirs` need to be deleted too. So I am not sure if counting till the number is the best way to go.

langner: I understand these are at the end only to fill up a line?

cclib/parser/turbomoleparser.py
kunalsharma05: No, `x[0] = [-0.01201, 0.00079, 0.01032, 0.01229, -0.22313, -0.00142]` so `x[0]` is basically an array of the x values (like in this [line](https://github.com/cclib/cclib/blob/master/data/Turbomole/basicTurbomole5.9/dvb_ir/aoforce.out#L870)) and for an entire block of 20 atoms we have `x` as an array consisting of 20 arrays `x[0],x[1] ... x[19]` which have the 6 values for 6 modes that are printed in a line. `y and z` are similar in this aspect. So zip(x,y,z) would just make an array where the zipped elements would themselves be arrays with x,y,z coordinates for different modes.

cclib/parser/turbomoleparser.py
kunalsharma05: Yes. Making the change.

cclib/parser/turbomoleparser.py
langner: We discussed this offline, and opted for keeping the while in place here.

cclib/parser/turbomoleparser.py
langner: True, but you could use it in the inner loop?

PR #566, opened 2018-07-11
Turbomole: remove duplicate unit test data files

berquist:


kunalsharma05:
@berquist Was looking to open a PR for this. Don't know how this got added, maybe something to do with my merging. Thanks for handling it.

PR #567, opened 2018-07-16
Turbomole: Added functionality for sorting the inputs to Turbomole parser

kunalsharma05:
The order for sorting the inputs looks something like this:
```python
    sorting_order = {
        'basis' : 0,
        'control' : 1,
        'mos' : 2,
        'alpha' : 3,
        'beta' : 4,
        'job.last' : 5,
        'coord' : 6,
        'gradient' : 7,
        'aoforce' : 8,
    }
```

### DIFF
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -292,6 +292,9 @@ def ccopen(source, *args, **kargs):
         else:
             inputfile.seek(0, 0)
         if not is_stream:
+            if is_listofstrings:
+                if filetype == Turbomole:
+                    source = sort_turbomole_outputs(source)
             inputfile.close()
             return filetype(source, *args, **kargs)
         return filetype(inputfile, *args, **kargs)
@@ -416,3 +419,55 @@ def _determine_output_format(outputtype, outputdest):
             raise UnknownOutputFormatError(extension)
 
     return outputclass
+
+def path_leaf(path):
+    """
+    Splits the path to give the filename. Works irrespective of '\'
+    or '/' appearing in the path and also with path ending with '/' or '\'.
+
+    Inputs:
+      path - a string path of a logfile.
+    Returns:
+      tail - 'directory/subdirectory/logfilename' will return 'logfilename'.
+      ntpath.basename(head) - 'directory/subdirectory/logfilename/' will return 'logfilename'.
+    """
+    head, tail = os.path.split(path)
+    return tail or os.path.basename(head)
+
+def sort_turbomole_outputs(filelist):
+    """
+    Sorts a list of inputs (or list of log files) according to the order
+    defined below. Just appends the unknown files in the end of the sorted list.
+
+    Inputs:
+      filelist - a list of Turbomole log files needed to be parsed.
+    Returns:
+      sorted_list - a sorted list of Turbomole files needed for proper parsing.
+    """
+    sorting_order = {
+        'basis' : 0,
+        'control' : 1,
+        'mos' : 2,
+        'alpha' : 3,
+        'beta' : 4,
+        'job.last' : 5,
+        'coord' : 6,
+        'gradient' : 7,
+        'aoforce' : 8,
+    }
+
+    known_files = []
+    unknown_files = []
+    sorted_list = []
+    for fname in filelist:
+        filename = path_leaf(fname)
+        if filename in sorting_order:
+            known_files.append([fname, sorting_order[filename]])
+        else:
+            unknown_files.append(fname)
+    for i in sorted(known_files, key=lambda x: x[1]):
+        sorted_list.append(i[0])
+    if unknown_files:
+        sorted_list.append(known_files)
+
+    return sorted_list


cclib/io/ccio.py
```diff
@@ -12,6 +12,7 @@
 
 import atexit
 import io
+import ntpath
```
berquist: Using `os.path` will remove the need for some of `path_leaf`. It's even recommended by `ntpath`:
```
In [1]: import ntpath

In [2]: ntpath?
Type:        module
String form: <module 'ntpath' from '/home/eric/virtualenvs/cclib/lib/python3.6/ntpath.py'>
File:        ~/virtualenvs/cclib/lib/python3.6/ntpath.py
Docstring:  
Common pathname manipulations, WindowsNT/95 version.

Instead of importing this module directly, import os and refer to this
module as os.path.
```

kunalsharma05:
@berquist So the error while using `basis` as the first argument here is that the `ccio.ccopen()` is not able to guess what parser the file belongs to.

cclib/io/ccio.py
```diff
@@ -292,6 +293,8 @@ def ccopen(source, *args, **kargs):
         else:
             inputfile.seek(0, 0)
         if not is_stream:
+            if filetype == Turbomole:
+                source = sort_turbomole_outputs(source)
```
kunalsharma05: @berquist A simple check `if is_listofstrings:` will remove the second failure where `job.last` is passed as the first parameter. Despite that I am a bit conufesd on how to make cclib guess that a file is a part of Turbomole outputs. When we use `ccopen()`  with a list as input then we are able to guess the correct parser since `job.last` is needed to find the correct trigger  but `ccget` seems to be looking at one file at a time and that is throwing it off since it only correctly predicts for `job.last` and not for the files (e.g. `basis`, `control`).

cclib/io/ccio.py
```diff
@@ -416,3 +419,55 @@ def _determine_output_format(outputtype, outputdest):
...
+    Splits the path to give the filename. Works irrespective of '\'
+    or '/' appearing in the path and also with path ending with '/' or '\'.
+
+    Inputs:
+      path - a string path of a logfile.
+    Returns:
+      tail - 'directory/subdirectory/logfilename' will return 'logfilename'.
+      ntpath.basename(head) - 'directory/subdirectory/logfilename/' will return 'logfilename'.
+    """
+    head, tail = ntpath.split(path)
```
kunalsharma05: @berquist So you mean that I should use `os.path.split()` here?

cclib/io/ccio.py
langner: As Eric alluded, we need to treat all the files as one buffered file object. This has come up in the past, and we've always come to the conclusion that we want to hide the particulars of looping over multiple files, streams or whatnot, and expose a consistent buffered file object to the parsers. We actually have a `FileWrapper` class in `logfileparser.py` which IIRC was meant to implement this kind of thing, so if it doesn't we need to extend it.

cclib/io/ccio.py
berquist: Without working it out myself, that sounds right.

berquist:
Ok, now this happens, which makes a lot more sense to me. I know this doesn't happen for Molpro because there are triggers in both files.
```
$ ccget --verbose --list job.last control basis coord energy mos statistics 
Attempting to read job.last
Identified logfile to be in Turbomole format
[Turbomole job.last WARNING] Attribute atomcoords changed value ([] -> [[-1.42441979549698, -3.7725043365061e-05, -0.23660895310213872], [-0.8988161976787694, -3.8799273103188e-05, 1.092692450551379], [0.4904618018011955, -3.947661993314e-05, 1.3207849314567357], [1.424419800788752, -3.7714459820843e-05, 0.23660895310213872], [0.8988161976787694, -3.8793981331079005e-05, -1.092692450551379], [-1.5897676498375695, -5.799782231464e-05, 1.9539105333108941], [0.874014810118434, -5.2579047675024e-05, 2.3572561610518576], [1.5897676286704812, -5.7976655226204005e-05, -1.9539105491862105], [2.890209400608588, -6.9412174753753e-05, 0.5369996407258266], [3.9015708275813177, 0.00017535345237593302, -0.36436857226200126], [3.7273846131301, 0.00048809718401783303, -1.452823705090136], [3.137403591295802, -0.000348928869323242, 1.6160099287336924], [4.954965731767726, 9.1166649893852e-05, -0.03939893671857988], [-2.89020940590036, -6.9417466525862e-05, -0.5369996407258266], [-3.137403591295802, -0.000348960619955896, -1.6160099340254646], [-3.9015708275813177, 0.000175364035920151, 0.36436857226200126], [-3.727384602546555, 0.000488134226422596, 1.452823705090136], [-4.954965731767726, 9.1171941665961e-05, 0.0393989473021241], [-0.49046180709296755, -3.9481911705249006e-05, -1.3207849314567357], [-0.874014762492485, -5.2531421726043006e-05, -2.357256176927174]])
cclib can parse the following attributes from job.last:
  atomcoords
  atomnos
  metadata
Attempting to read control
Attempting to use fallback mechanism to read file
Cannot figure out the format of 'control'
Report this to the cclib development team if you think it is an error.

Usage:  ccget <attribute> [<attribute>] <compchemlogfile> [<compchemlogfile>]
Try     ccget --help for more information
```
I am ok with this current ccget behavior if everyone else is, because it's more important that ccopen and ccread work properly.

Issue #256, opened 2016-04-07
Support for rotational constants


ghutchis:
I think there should be support for rotational constants - either in cm-1 or GHz.

These would be stored as an array of rank 1 (i.e., 3 of them). I can contribute a bunch of examples.

ghutchis:
I mention this because I'd really like to plot rotational spectra in Avogadro and PQR.pitt.edu.

berquist:
I'm not against this, but not every program calculates this even though it is easy enough to calculate separately: https://gist.github.com/c730ca8a2af4df112933dc4d310ac554

langner:
We already both parse and calculate some things (populations come to mind), so this is a matter of taste. Can we add a method for rotational constants based on your gist?

berquist:
> Can we add a method for rotational constants based on your gist?

Yes, I was thinking of this. I'll clean it up and add it.

ghutchis:
One nitpick - these should be be exact isotopic masses..

berquist:
That would explain why my numbers are slightly off. Now I think it's a good idea for this to be added to the parser.

ghutchis:
Have we defined an attribute name? I can add a few of these (e.g., Gaussian, MOPAC). I'll see if I have time to add the list of default isotopic masses too.

berquist:
It looks like I was going to call it `rotcons`.

Having a list of default isotopic masses would be a great help.

langner:
Why not `rotconst`?

If we ever get to version 2.0 I would advocate more verbose attribute names.

berquist:
No particular reason. I agree.

langner:
Pushing back.

berquist:
I have an implementation as part of the Nuclear method. Physical constants come from [`scipy.constants`](https://docs.scipy.org/doc/scipy/reference/constants.html) and isotopic masses come from [`periodictable`](https://pypi.org/project/periodictable/). For isotopic masses, the most abundant isotope is used.

Are there better choices?

berquist:
No, this is done, taken care of by https://github.com/cclib/cclib/pull/568.

Issue #555, opened 2018-06-27
Developer guidelines


langner:
It'd be useful to have some developer guidelines in our docs, so that we can point contributors to it, and also achieve consensus among ourselves and make our lives easier. This would include at least:
1. Git workflow policies, including merging/rebasing conventions or rules
1. Review workflow, including our unwritten rule that you can't merge your own code (with whatever exceptions we deem appropriate)
1. Style recommendations

langner:
4. Procedures for unit test files and code.
5. Procedures for regression tests and files in cclib/cclib-data

langner:
6. Parsing conventions, such as not predicating parsing on previously parsed attributes. So, parse data multiple times if available, even if we don't use the value for attributes. OTOH, prefer overwriting re-parsed attributes, or at least checking their consistency, via `set_attribute`.

berquist:
Rebasing formatting changes is making it too difficult to review inside a PR. If we are unhappy with the number of commits in a PR, we should either:
1. "Squash and merge"
2. Request an interactive rebase to squash those commits just before merging the PR.

langner:
7. Describe our approach to parsing files, streams and multiples files and/or directories.

langner:
Not for points 4 & 5: when updating unit test files, first copy the newest version of the test logfiles to a new directory and then update them with the actual new contents. That way the history in git shows the changes from the last version by default, which makes things sometimes easier to debug. This was done recently when updating ORCA in #575 and works nicely.

langner:
8. Describe our philosophy for methods: willing to include anything that relies on attributes and physical constants. Methods should also be pure Python and reasonable light.

berquist:
From https://github.com/cclib/cclib/pull/568#discussion_r212835180,

> let's have a convention to use only one tool if we want to auto-format. We have enough pep8 violations that it would make sense to format in bulk once if we want to go in that direction.

berquist:
Now that docs are easy to modify again, how would you like to do this? Will everything go in the already existing `development.rst`?

langner:
> Now that docs are easy to modify again, how would you like to do this? Will everything go in the already existing development.rst?

I dunno, haven't thought about the details yet. One thought: best to do it incrementally, so it's easier to review and have a discussion (where needed).

Also...

9. From cclib/cclib.github.io#21: I think we need to outline in the dev section what we expect from pull requests and point people to that when they send very large changes, etc.

langner:
[...]

PR #568, opened 2018-07-17
Method for calculating rotational constants

berquist:
Additionally,
* isotopic masses via `periodictable` package
* center of mass
* moment of inertia tensor
* principal moments of inertia

Handles https://github.com/cclib/cclib/issues/256.

Right now, each method returns multiple arrays, one for each set of units. Another way to do it would be to pass a string with the desired output units ("MHz", "cm-1") and only get back one array. That's probably better.

This also only considers the last set of `atomcoords`.


requirements.txt
```diff
@@ -1,5 +1,7 @@
 biopython ; python_version == '2.7' or python_version >= '3.4'
 numpy
+scipy
+periodictable
```
langner: We have a periodic table thing in util: https://github.com/cclib/cclib/blob/master/cclib/parser/utils.py#L124

Do you think we should replace that with this package, or standardize on our own version?

berquist: Our periodic table is not sufficient, because we need masses, specifically isotopic masses. I am not sure if `atommasses`, when present, are always isotopic masses. They should be when parsed from a frequency calculation. However I don't believe they're always available.

If there is a better Python package for this sort of data, and a better one for physical constants than scipy, I am open to suggestions.

We can eventually have our periodic table wrap an external package, once we pick an appropriate one, because holding our own atomic masses doesn't sound good.

berquist:
I am interested in @ghutchis's thoughts about what features are desired and if he knows about other packages, since he originally requested this feature.

berquist:
For some physical constants, [unyt](https://unyt.readthedocs.io/en/latest/modules/unyt.physical_constants.html) provides a few. One thing I like about it is that it provides the opportunity to handle units in the future. Something I dislike is the provenance of the units is not as clear as SciPy, which specifically says the CODATA version they originate from.

I cannot find a better atomic mass/isotope library than periodictable that is already in Python. There is an excellent one in DALTON, but we cannot use even a Python translation because it is GPL'ed.

berquist:
periodictable does not work with Python 3.2 due to Unicode strings, but is tested with 3.3.

berquist:
I just rediscovered [mendeleev](https://bitbucket.org/lukaszmentel/mendeleev), which may be even better than periodictable.

ATenderholt:
Mendeleev seems nice, but has some pretty heavy dependencies like sqlalchemy: 
https://bitbucket.org/lukaszmentel/mendeleev/src/fcbaa655eb406237ae465ccc44436248d4b8d145/setup.py?at=default&fileviewer=file-view-default#setup.py-34.

berquist:
- periodictable requires numpy and pyparsing.
- mendeleev requires six, python-dateutil, pytz, pandas, colorama, pyfiglet, and sqlalchemy.

I did a quick check and there's a small SQLite3 database (< 400 KB) that comes with the package. periodictable stores all its data in a massive comma-separated string. I have no strong opinion either way.

ATenderholt:
My vote is against mendeleev because it has so many dependencies completely unrelated to cclib. Since it's MIT licensed, couldn't we just take their database, write custom queries with the builtin sqlite module, and distribute it with cclib?

berquist:
I would rather choose one package or the other over taking a piece of one, especially since they are both being actively developed. I don't think that there's anything missing from periodictable, so it could replace our own implementation.

ATenderholt:
Fair point about not taking parts of a package if there are actively developed. I agree with going with periodictable.

langner:
I think adding a new dependency requires a bit more thought, a slightly more prolonged discussion. Is there a way to use our own stuff for this PR and have this discussion outside of the development cycle?

berquist:
Just a reminder that the build is failing because of 3.2; bumping to 3.4 solves this and is not really limited to just using `periodictable` and `scipy`.

A temporary solution would be to have the mass-dependent parts of this Method work like the bridges, where methods that do not require the external dependency are still usable. So things like the nuclear repulsion energy calculation would still work. It should already be written this way in the PR.

> Is there a way to use our own stuff for this PR

It depends on what you mean by this. We don't have masses, either average ones or the isotopic ones we need.

cclib/method/nuclear.py
```diff
@@ -31,16 +31,120 @@ def __repr__(self):
...
             zi = self.data.atomnos[i]
             for j in range(i+1, self.data.natom):
                 rj = self.data.atomcoords[0][j]
                 zj = self.data.atomnos[j]
-                d = numpy.linalg.norm(ri-rj)
+                d = np.linalg.norm(ri-rj)
                 nre += zi*zj/d
         return nre
+
+    @staticmethod
```
langner: Why shouldn't this be a module-level function?

cclib/method/nuclear.py
```diff
@@ -31,16 +31,120 @@ def __repr__(self):
...
+        """
+        most_abundant_isotope = element.isotopes[0]
+        abundance = 0
+        for iso in element:
+            if iso.abundance > abundance:
+                most_abundant_isotope = iso
+                abundance = iso.abundance
+        return most_abundant_isotope
+
+    @staticmethod
```
langner: Ditto.

requirements.txt
langner: OK, I'm OK with this choice. `periodictable` look reasonable.

berquist:
The advantage to having all these methods is that they live right next to data type they take as input, and are an example of what you can do with cclib. The hope is that you don't need to always reinvent the wheel when building an analysis on top of `ccData` objects.

It's nice that using the parser and readers/writers doesn't require using the methods. However, going the other direction, having the readers and writers close by means that you can take just an XYZ file and calculate the nuclear repulsion energy or rotational constants, for example.

An option is to split off all the methods into their own repository. My feeling is that many people don't know about or use the methods as it is, and moving them somewhere else would make that worse.

I think that we should allow any method that doesn't require more than attributes and physical constants, as long as its correctness can be verified.

langner:
Thanks for the thoughts, my opinion is roughly the same. Added this topic to #555 which should ideally go into docs for developers.

cclib/method/nuclear.py
```diff
@@ -9,12 +9,39 @@
...
 
-import numpy
+import numpy as np
 
 from cclib.method.calculationmethod import Method
 from cclib.parser.utils import PeriodicTable
 
 
+def get_most_abundant_isotope(element):
+    """Given a `periodictable` element, return the most abundant
```
langner: I'm not a fan of strictly keeping to the 80 column limit. Having slightly longer lines so that they don't wrap in a n ugly way is fine IMO, as long as the length is reasonable. What do you think? We could also cover this in #555.

cclib/method/nuclear.py
```diff
@@ -9,12 +9,39 @@
...
+            most_abundant_isotope = iso
+            abundance = iso.abundance
+    return most_abundant_isotope
+
+
+def get_isotopic_masses(charges):
+    """Return the masses for the given nuclei, respresented by their
+    nuclear charges.
+    """
+    import periodictable
```
langner: Shouldn't we import (softly) at top of module?

berquist: Yes, I'll make it consistent with the bridges for now.

cclib/method/nuclear.py
berquist: I think I'm in agreement; I have a habit of running `fill-paragraph` (or your editor's equivalent, `gq`?) for docstrings, and the default column width is 70. I can stop doing that.

langner: Yeah, let's have a convention to use only one tool if we want to auto-format. We have enough pep8 violations that it would make sense to format in bulk once if we want to go in that direction.

langner:
I think this is ready to go, and the Travis failure is for 3.2... @berquist probably need to rebase to get it tested in 3.4 (that was merged a moment ago).

PR #569, opened 2018-07-18
Turbomole: Added functionality for moenergies and mocoeffs

kunalsharma05:
mocoeffs and moenergies are usually parsed from `mos, alpha and beta` files.

Removed tests from testSP.py: testdimmocoeffs, testlengthmoenergies, testtypemoenergies
Removed tests from testSPun.py: testdimmocoeffs, testmoenergies

### DIFF
--- a/cclib/parser/turbomoleparser.py
+++ b/cclib/parser/turbomoleparser.py
@@ -66,6 +66,26 @@ class Turbomole(logfileparser.Logfile):
     def before_parsing(self):
         self.geoopt = False # Is this a GeoOpt? Needed for SCF targets/values.
 
+    @staticmethod
+    def split_molines(inline):
+        """Splits the lines containing mocoeffs (each of length 20)
+        and converts them to float correctly.
+        """
+        line = inline.replace("D", "E")
+        f1 = line[0:20]
+        f2 = line[20:40]
+        f3 = line[40:60]
+        f4 = line[60:80]
+
+        if(len(f4) > 1):
+            return [float(f1), float(f2), float(f3), float(f4)]
+        if(len(f3) > 1):
+            return [float(f1), float(f2), float(f3)]
+        if(len(f2) > 1):
+            return [float(f1), float(f2)]
+        if(len(f1) > 1):
+            return [float(f1)]
+
     def extract(self, inputfile, line):
         """Extract information from the file object inputfile."""
 
@@ -207,6 +227,58 @@ class Turbomole(logfileparser.Logfile):
             self.set_attribute('vibsyms', vibsyms)
             self.set_attribute('vibirs', vibirs)
             self.set_attribute('vibdisps', vibdisps)
+
+        # In this section we are parsing mocoeffs and moenergies from
+        # the files like: mos, alpha and beta.
+        # $scfmo    scfconv=6   format(4d20.14)
+        # # SCF total energy is     -382.3457535740 a.u.
+        # #
+        #      1  a      eigenvalue=-.97461484059799D+01   nsaos=60
+        # 0.69876828353937D+000.32405121159405D-010.87670894913921D-03-.85232349313288D-07
+        # 0.19361534257922D-04-.23841194890166D-01-.81711001390807D-020.13626356942047D-02
+        # ...
+        # ...
+        # $end
+        if (line.startswith('$scfmo') or line.startswith('$uhfmo')) and line.find('scfconv') > 0:
+            if line.strip().startswith('$uhfmo_alpha'):
+                self.unrestricted = True
+
+            # Need to skip the first line to start with lines starting with '#'.
+            line = next(inputfile)
+            while line.strip().startswith('#') and not line.find('eigenvalue') > 0:
+                line = next(inputfile)
+
+            moenergies = []
+            mocoeffs = []
+
+            while not line.strip().startswith('$'):
+                info = re.match(".*eigenvalue=(?P<moenergy>[0-9D\.+-]{20})\s+nsaos=(?P<count>\d+).*", line)
+                eigenvalue = self.float(info.group('moenergy'))
+                orbital_energy = utils.convertor(eigenvalue, 'hartree', 'eV')
+                moenergies.append(orbital_energy)
+                single_coeffs = []
+                nsaos = int(info.group('count'))
+
+                while(len(single_coeffs) < nsaos):
+                    line = next(inputfile)
+                    single_coeffs.extend(Turbomole.split_molines(line))
+
+                mocoeffs.append(single_coeffs)
+                line = next(inputfile)
+
+            max_nsaos = max([len(i) for i in mocoeffs])
+            for i in mocoeffs:
+                while len(i) < max_nsaos:
+                    i.append(numpy.nan)
+
+            if not hasattr(self, 'mocoeffs'):
+                self.mocoeffs = []
+
+            if not hasattr(self, 'moenergies'):
+                self.moenergies = []
+
+            self.mocoeffs.append(mocoeffs)
+            self.moenergies.append(moenergies)
 
     def deleting_modes(self, vibfreqs, vibdisps, vibirs):
         """Deleting frequencies relating to translations or rotations"""
--- a/test/data/testSP.py
+++ b/test/data/testSP.py
@@ -174,12 +174,10 @@ class GenericSPTest(unittest.TestCase):
         """Are correct number of SCF convergence criteria being parsed?"""
         self.assertEquals(len(self.data.scftargets[0]), self.num_scf_criteria)
 
-    @skipForParser('Turbomole','The parser is still being developed so we skip this test')
     def testlengthmoenergies(self):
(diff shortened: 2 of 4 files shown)


cclib/parser/turbomoleparser.py
```diff
@@ -66,6 +66,25 @@ def normalisesym(self, label):
     def before_parsing(self):
         self.geoopt = False # Is this a GeoOpt? Needed for SCF targets/values.
 
+    def split_molines(self, inline):
```
berquist: This should be a static method: https://stackoverflow.com/q/12179271/3249688

cclib/parser/turbomoleparser.py
```diff
@@ -208,6 +227,51 @@ def extract(self, inputfile, line):
...
+
+            while not (line.strip().startswith('1') and line.find('eigenvalue') > 0):
+                line = next(inputfile)
+
+            moenergies = []
+            mocoeffs = []
+
+            while not line.strip().startswith('$'):
+                info = line.split()
+                eigenvalue = float(info[2][11:].replace('D', 'E'))
```
berquist: You can use `self.float`: https://github.com/cclib/cclib/blob/180f02ccc556b6e96f5f0e9b4593ddeb7c383dcd/cclib/parser/logfileparser.py#L393

cclib/parser/turbomoleparser.py
```diff
@@ -208,6 +227,51 @@ def extract(self, inputfile, line):
...
+        # 0.69876828353937D+000.32405121159405D-010.87670894913921D-03-.85232349313288D-07
+        # 0.19361534257922D-04-.23841194890166D-01-.81711001390807D-020.13626356942047D-02
+        # ...
+        # ...
+        # $end
+        if (line.startswith('$scfmo') or line.startswith('$uhfmo')) and line.find('scfconv') > 0:
+            if line.strip().startswith('$uhfmo_alpha'):
+                self.unrestricted = True
+
+            while not (line.strip().startswith('1') and line.find('eigenvalue') > 0):
```
berquist: `startswith('1')` should not be there; that's the 1-based index of the MO.

cclib/parser/turbomoleparser.py
```diff
@@ -208,6 +227,51 @@ def extract(self, inputfile, line):
...
+            mocoeffs = []
+
+            while not line.strip().startswith('$'):
+                info = line.split()
+                eigenvalue = float(info[2][11:].replace('D', 'E'))
+                orbital_energy = utils.convertor(eigenvalue, 'hartree', 'eV')
+                moenergies.append(orbital_energy)
+                single_coeffs = []
+
+                while(len(single_coeffs) < self.nbasis):
```
ATenderholt: Does `nsaos=60` in the output indicate how many numbers in the coeffs block? If so, perhaps using that explicit number is better than relying on `self.nbasis`.

cclib/parser/turbomoleparser.py
kunalsharma05: Agreed!

cclib/parser/turbomoleparser.py
kunalsharma05: How is it different from just using `float()`?

cclib/parser/turbomoleparser.py
kunalsharma05: Yeah. I understood that after seeing the `water` mos file. Will make the changes.

cclib/parser/turbomoleparser.py
langner: Better, just make it a module-level function.

cclib/parser/turbomoleparser.py
```diff
@@ -208,6 +228,53 @@ def extract(self, inputfile, line):
             self.set_attribute('vibirs', vibirs)
             self.set_attribute('vibdisps', vibdisps)
 
+        # In this section we are parsing mocoeffs and moenergies from
+        # the files like: mos, alpha and beta.
+        # $scfmo    scfconv=6   format(4d20.14)
```
ATenderholt: The presence of `format(4d20.14)` makes me wonder if Turbomol has an option to change the format of the orbitals block. Is there anything in the documentation that talks about this?

cclib/parser/turbomoleparser.py
```diff
@@ -208,6 +228,53 @@ def extract(self, inputfile, line):
...
+            moenergies = []
+            mocoeffs = []
+
+            while not line.strip().startswith('$'):
+                info = line.split()
+                eigenvalue = float(info[2][11:].replace('D', 'E'))
+                orbital_energy = utils.convertor(eigenvalue, 'hartree', 'eV')
+                moenergies.append(orbital_energy)
+                single_coeffs = []
+                nsaos = int(info[3][6:])
```
ATenderholt: Perhaps parsing `eigenvalue` and `nsaos` can be more explicit with regular expressions. Python allows named matched groups:

```
>>> m = re.match(".*eigenvalue=(?P<moenergy>.{20})\s+nsaos=(?P<count>\d+).*", a)
>>> m.group("moenergy")
'-.97461484059799D+01'
>>> m.group("count")
'60'
```

I just quickly hacked the above regex, so it should be improved. `.{20}` should be replaced with the expected characters (`.`, `-`, `0-9`, `D`, etc.) instead of the match-any `.`.

cclib/parser/turbomoleparser.py
```diff
@@ -208,6 +228,53 @@ def extract(self, inputfile, line):
...
+
+                while(len(single_coeffs) < nsaos):
+                    line = next(inputfile)
+                    single_coeffs.extend(Turbomole.split_molines(line))
```
ATenderholt: Can't you just use `split_molines` without `Turbomol`?

cclib/parser/turbomoleparser.py
kunalsharma05: I didn't find it as such. What does this mean exactly?

cclib/parser/turbomoleparser.py
kunalsharma05: Alright! Will make an expression to get this. Makes more sense to use something dynamic.

cclib/parser/turbomoleparser.py
ATenderholt: 4 entries per column, and each entry has 20 characters with 14 decimal places. I'm guessing the d means using D instead of E for the exponent.

cclib/parser/turbomoleparser.py
ATenderholt: If you look at the code, `self.float` does basically everything you're doing here.

cclib/parser/turbomoleparser.py
berquist: If we are comfortable adding a dependency, I've used [fortranformat](https://pypi.org/project/fortranformat/) in the past and really like it.

cclib/parser/turbomoleparser.py
```diff
@@ -208,6 +228,58 @@ def extract(self, inputfile, line):
...
+            # Need to skip the first line to start with lines starting with '#'.
+            line = next(inputfile)
+            while line.strip().startswith('#') and not line.find('eigenvalue') > 0:
+                line = next(inputfile)
+
+            moenergies = []
+            mocoeffs = []
+
+            while not line.strip().startswith('$'):
+                info = re.match(".*eigenvalue=(?P<moenergy>[-]?((\d+(\.\d*)?)|\.\d+)([dD][+-]?[0-9]+)?)\s+nsaos=(?P<count>\d+).*", line)
```
ATenderholt: How about `re.match(".*eigenvalue=(?P<moenergy>[0-9D\.+-]{20})\s+nsaos=(?P<count>\d+).*", line)`? This way there are only 2 total groups instead of 6:

```
>>> a = "     1  a      eigenvalue=-.97461484059799D+01   nsaos=60"
>>> re.match(".*eigenvalue=(?P<moenergy>[0-9D\.+-]{20})\s+nsaos=(?P<count>\d+).*", a)
<re.Match object; span=(0, 57), match='     1  a      eigenvalue=-.97461484059799D+01   >
>>> _.groups()
('-.97461484059799D+01', '60')
>>> re.match(".*eigenvalue=(?P<moenergy>[-]?((\d+(\.\d*)?)|\.\d+)([dD][+-]?[0-9]+)?)\s+nsaos=(?P<count>\d+).*", a)
<re.Match object; span=(0, 57), match='     1  a      eigenvalue=-.97461484059799D+01   >
>>> _.groups()
('-.97461484059799D+01', '.97461484059799', None, None, 'D+01', '60')
```

cclib/parser/turbomoleparser.py
langner: I dunno, it doesn't feel like a compelling reason to add it as a dependency. We can easily add the functionality we need in one function, no?

cclib/parser/turbomoleparser.py
langner: So... then maybe it should go into `utils.py`?

PR #570, opened 2018-07-23
Turbomole: Added functionality for scfvalues, scftargets and scfenergies

kunalsharma05:
Added functionality for the following attributes:
```
1. scfvalues
2. scftargets
3. scfenergies
```

### DIFF
--- a/cclib/parser/turbomoleparser.py
+++ b/cclib/parser/turbomoleparser.py
@@ -140,34 +140,6 @@ class Turbomole(logfileparser.Logfile):
             self.append_attribute('atomcoords', atomcoords)
             self.set_attribute('atomnos', atomnos)
 
-        ## If we are unable to find coordinates in the job file, we will look for them
-        ## in the coord file.
-        #$coord
-        #   -2.69176330280845     -0.00007129445712     -0.44712612093731      c
-        #   -1.69851644615005     -0.00007332282157      2.06488947265450      c
-        #    0.92683848474542     -0.00007460039817      2.49592179180606      c
-        #    2.69176330586455     -0.00007127328394      0.44712611937145      c
-        #    1.69851645303760     -0.00007330575699     -2.06488946767951      c
-        #...
-        #   -7.04373605585274      0.00092243787879      2.74543890990978      h
-        #   -9.36352819434217      0.00017228707094      0.07445321997922      h
-        #   -0.92683848797451     -0.00007460625018     -2.49592178685491      c
-        #   -1.65164852640697     -0.00009927259342     -4.45456857763148      h
-        #$redundant
-        if line.startswith("$coord"):
-            if '$coordinate' not in line:
-                line = next(inputfile)
-                if '$user' not in line: 
-                    atomcoords = []
-                    atomnos = []
-                    while line[0] != "$":
-                        atomnos.append(utils.PeriodicTable().number[line.split()[3].upper()])
-                        atomcoords.append([utils.convertor(float(x), "bohr", "Angstrom") for x in line.split()[:3]])
-                        line = next(inputfile)
-
-                    self.set_attribute('atomcoords', atomcoords)
-                    self.set_attribute('atomnos', atomnos)
-
         # Frequency values in aoforce.out
         #        mode               7        8        9       10       11       12
         #
--- a/cclib/parser/turbomoleparser.py
+++ b/cclib/parser/turbomoleparser.py
@@ -65,6 +65,7 @@ class Turbomole(logfileparser.Logfile):
 
     def before_parsing(self):
         self.geoopt = False # Is this a GeoOpt? Needed for SCF targets/values.
+        self.periodic_table = utils.PeriodicTable()
 
     @staticmethod
     def split_molines(inline):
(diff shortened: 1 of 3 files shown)


cclib/parser/turbomoleparser.py
```diff
@@ -65,6 +65,7 @@ def normalisesym(self, label):
 
     def before_parsing(self):
         self.geoopt = False # Is this a GeoOpt? Needed for SCF targets/values.
+        self.scf_converged = True # Need a check to determine whether scf convergence fails.
```
ATenderholt: Shouldn't this be set to `False` here and be changed to `True` when it has converged?

cclib/parser/turbomoleparser.py
```diff
@@ -208,6 +209,75 @@ def extract(self, inputfile, line):
...
+                while 'convergence criteria satisfied' not in line:
+                    if 'ITERATION  ENERGY' in line:
+                        line = next(inputfile)
+                        info = line.split()
+                        iter_energy.append(utils.convertor(self.float(info[1]), 'hartree', 'eV'))
+                        iter_one_elec_energy.append(utils.convertor(self.float(info[2]), 'hartree', 'eV'))
+                    line = next(inputfile)
+
+                scfvalues = []
+                if len(iter_energy) == len(iter_one_elec_energy):
```
ATenderholt: Isn't this always true based on your code above? Suggest using `assert` to verify that condition holds instead of an `if`-statement.

cclib/parser/turbomoleparser.py
kunalsharma05: I am trying to find a trigger for finding if the convergence occurs but the problem is that Turbomole prints energy converged before the last iteration. I was thinking of parsing the data as is and change this to `true` and stop iterating when we find the `energy converged` keyword (something along these lines)

cclib/parser/turbomoleparser.py
```diff
@@ -208,6 +209,75 @@ def extract(self, inputfile, line):
...
+        #                 :  kinetic energy    =    375.67398458525  :
+        #                 :  potential energy  =   -758.01973815924  :
+        #                 :  virial theorem    =      1.98255043001  :
+        #                 :  wavefunction norm =      1.00000000000  :
+        #                  .......................................... 
+        if 'scf convergence criterion' in line:
+            total_energy_threshold = self.float(line.split()[-1])
+            one_electron_energy_threshold = self.float(next(inputfile).split()[-1])
+            scftargets = [total_energy_threshold, one_electron_energy_threshold]
+            if not hasattr(self, 'scftargets'):
```
langner: Do you think it would be possible to use `set_attribute` here? If not, are you able to add a new helper method that adds to an existing attribute?

kunalsharma05:
@langner @ATenderholt Made the requested changes. And removed some redundant code that was causing the tests to fail.

cclib/parser/turbomoleparser.py
```diff
@@ -124,50 +125,19 @@ def extract(self, inputfile, line):
...
-        #   -0.92683848797451     -0.00007460625018     -2.49592178685491      c
-        #   -1.65164852640697     -0.00009927259342     -4.45456857763148      h
-        #$redundant
-        if line.startswith("$coord"):
-            if '$coordinate' not in line:
+            atomcoords = []
+            atomnos = []
+            line = next(inputfile)
+            while len(line) > 2:
+                atomnos.append(utils.PeriodicTable().number[line.split()[3].upper()])
```
berquist: Rather than instantiate a new `PeriodicTable` every time this is called, could you have the class keep a copy of it?

cclib/parser/turbomoleparser.py
```diff
@@ -124,50 +125,19 @@ def extract(self, inputfile, line):
...
-        #   -1.65164852640697     -0.00009927259342     -4.45456857763148      h
-        #$redundant
-        if line.startswith("$coord"):
-            if '$coordinate' not in line:
+            atomcoords = []
+            atomnos = []
+            line = next(inputfile)
+            while len(line) > 2:
+                atomnos.append(self.periodic_table.number[line.split()[3].upper()])
+                atomcoords.append([utils.convertor(float(x), "bohr", "Angstrom") for x in line.split()[:3]])
```
berquist: Can you insert a line break?

cclib/parser/turbomoleparser.py
```diff
@@ -280,6 +250,63 @@ def extract(self, inputfile, line):
...
+            iter_one_elec_energy = []
+            while 'convergence criteria satisfied' not in line:
+                if 'ITERATION  ENERGY' in line:
+                    line = next(inputfile)
+                    info = line.split()
+                    iter_energy.append(self.float(info[1]))
+                    iter_one_elec_energy.append(self.float(info[2]))
+                line = next(inputfile)
+
+            assert len(iter_energy) == len(iter_one_elec_energy), 'Different number of values found for total energy and one electron energy.'
```
berquist: Do you know if it's possible to insert a line break? I've never used the assert message.

cclib/parser/turbomoleparser.py
```diff
@@ -280,6 +250,63 @@ def extract(self, inputfile, line):
...
+            scfvalues = [[x - y, a - b] for x, y, a, b in zip(iter_energy[1:],
+                                     iter_energy[:-1], iter_one_elec_energy[1:], iter_one_elec_energy[:-1])]
```
berquist: The indentation is weird here. Try inserting the line break before the `zip` instead.

kunalsharma05: Alright! I wasn't sure about the indentation.

PR #573, opened 2018-07-26
Molcas: Added functionality for mpenergies

kunalsharma05:
Also, removed tests from testMP.py:
```
1. testsizeandshape
2. testchange
```
@ATenderholt @berquist @langner Please review. The tests will fail as the files are not yet merged.

### DIFF
--- a/cclib/parser/molcasparser.py
+++ b/cclib/parser/molcasparser.py
@@ -696,6 +696,28 @@ class Molcas(logfileparser.Logfile):
                     self.mocoeffs = []
                 self.mocoeffs.append(mocoeffs)
 
+        ## Parsing MP energy from the &MBPT2 module.
+        #  Conventional algorithm used...
+        #
+        #         SCF energy                           =      -74.9644564043 a.u.
+        #         Second-order correlation energy      =       -0.0364237923 a.u.
+        #
+        #         Total energy                         =      -75.0008801966 a.u.
+        #         Reference weight ( Cref**2 )         =        0.98652
+        #
+        #  ::    Total MBPT2 energy                              -75.0008801966
+        #
+        #
+        #         Zeroth-order energy (E0)             =      -36.8202538520 a.u.
+        #
+        #         Shanks-type energy S1(E)             =      -75.0009150108 a.u.
+        if 'Total MBPT2 energy' in line:
+            mpenergies = []
+            mpenergies.append(utils.convertor(self.float(line.split()[4]), 'hartree', 'eV'))
+            if not hasattr(self, 'mpenergies'):
+                self.mpenergies = []
+            self.mpenergies.append(mpenergies)
+
 
 if __name__ == '__main__':
     import sys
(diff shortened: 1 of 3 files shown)


cclib/parser/molcasparser.py
```diff
@@ -696,6 +696,30 @@ def extract(self, inputfile, line):
...
+        #
+        #         Zeroth-order energy (E0)             =      -36.8202538520 a.u.
+        #
+        #         Shanks-type energy S1(E)             =      -75.0009150108 a.u.
+        if 'Total MBPT2 energy' in line:
+            mpenergies = []
+            mpenergies.append(utils.convertor(self.float(line.split()[4]), 'hartree', 'eV'))
+            self.skip_lines(inputfile, ['b', 'b', 'zeroth order energy', 'b'])
+            line = next(inputfile)
+            if not hasattr(self, 'mpenergies'):
```
langner: Is using `set_attribute` possible here?

cclib/parser/molcasparser.py
```diff
@@ -696,6 +696,30 @@ def extract(self, inputfile, line):
...
+        #  ::    Total MBPT2 energy                              -75.0008801966
+        #
+        #
+        #         Zeroth-order energy (E0)             =      -36.8202538520 a.u.
+        #
+        #         Shanks-type energy S1(E)             =      -75.0009150108 a.u.
+        if 'Total MBPT2 energy' in line:
+            mpenergies = []
+            mpenergies.append(utils.convertor(self.float(line.split()[4]), 'hartree', 'eV'))
+            self.skip_lines(inputfile, ['b', 'b', 'zeroth order energy', 'b'])
```
langner: Do you need to skip, and parse the additional line after this, after parsing what you need?

kunalsharma05: My bad. I thought that we needed to parse the Shanks type enregy as well so I wrote this. Will remove it.

cclib/parser/molcasparser.py
kunalsharma05: I initially used `set_attribute` but I reverted to more dynamic code since we don't know if we might need to parse mpenergies more than once e.g.if we use `&MBPT2` module in a loop like we use other modules during geometry optimization for each intermediate structure. I can use `set_attribute` by getting the data of that attribute (let's say `mpenergies` of one step), append it to the local variable that we are using to get the data and use that variable in the `set_attribute` method.

langner: Right, so you're saying you want to extend the attribute instead of overwriting it. That makes sense.

Is it reasonable to create a new helper function that will cover this use case? Or do you think `set_attribute` could perhaps be morphed to cover this case as well (optional argument)?

kunalsharma05: Maybe we can modify `set_attribute`. I will try to do that if during the end or maybe in a separate PR?

langner: OK

PR #574, opened 2018-07-26
Molcas: Added functionality for ccenergies 

kunalsharma05:
Also removed tests from testCC.py:
`1. testsign`

I also made some changes to the parsing of `++    Optimization specifications:` section as the `&RASSCF` section also has a `++    Optimization specifications:` section through which the while loop was unable to break, and the parsing was failing.
@ATenderholt @berquist @langner Please review. Thanks!
(The tests are failing as the files are not yet merged).

### DIFF
--- a/cclib/parser/molcasparser.py
+++ b/cclib/parser/molcasparser.py
@@ -222,36 +222,23 @@ class Molcas(logfileparser.Logfile):
         #  Threshold at which QNR/C2DIIS is turned on 0.75E-01
         #  Threshold for Norm(delta) (QNR/C2DIIS)     0.20E-04
         if line[:34] == '++    Optimization specifications:':
-
-            scftargets = []
-            while not line[6:37] == 'Threshold for SCF energy change':
-                line = next(inputfile)
-
-            if line[6:37] == 'Threshold for SCF energy change':
-                target = float(line.split()[-1])
-                scftargets.append(target)
-
+            self.skip_lines(inputfile, ['d', 'b'])
             line = next(inputfile)
+            if line.strip().startswith('SCF'):
+                scftargets = []
+                self.skip_lines(inputfile,
+                                ['Minimized', 'Number', 'Maximum', 'Maximum'])
+                lines = [next(inputfile) for i in range(7)]
+                targets = [
+                    'Threshold for SCF energy change',
+                    'Threshold for density matrix',
+                    'Threshold for Fock matrix',
+                    'Threshold for Norm(delta)',
+                ]
+                for y in targets:
+                    scftargets.extend([float(x.split()[-1]) for x in lines if y in x])
 
-            if line[6:34] == 'Threshold for density matrix':
-                target = float(line.split()[-1])
-                scftargets.append(target)
-
-            line = next(inputfile)
-
-            if line[6:31] == 'Threshold for Fock matrix':
-                target = float(line.split()[-1])
-                scftargets.append(target)
-
-            self.skip_lines(inputfile, ['Threshold for linear dependence', 'Threshold at which DIIS is turned on', \
-                                         'Threshold at which QNR/C2DIIS is turned on'])
-            line = next(inputfile)
-
-            if line[6:31] == 'Threshold for Norm(delta)':
-                target = float(line.split()[-1])
-                scftargets.append(target)
-
-            self.append_attribute('scftargets', scftargets)
+                self.append_attribute('scftargets', scftargets)
 
         #  ++ Convergence information
         #                                     SCF        iterations: Energy and convergence statistics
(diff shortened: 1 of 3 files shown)


cclib/parser/molcasparser.py
```diff
@@ -222,38 +222,41 @@ def extract(self, inputfile, line):
...
         if line[:34] == '++    Optimization specifications:':
-
-            scftargets = []
-            while not line[6:37] == 'Threshold for SCF energy change':
+            self.skip_lines(inputfile, ['d', 'b'])
+            line = next(inputfile)
+            if line.strip().startswith('SCF'):
+                scftargets = []
+                self.skip_lines(inputfile, ['Minimized', 'Number', \
+                                         'Maximum', 'Maximum'])
```
langner: I think that if you start the list on the next line you won't need to use the backslash.

cclib/parser/molcasparser.py
```diff
@@ -222,38 +222,41 @@ def extract(self, inputfile, line):
...
                 line = next(inputfile)
 
-            if line[6:37] == 'Threshold for SCF energy change':
-                target = float(line.split()[-1])
-                scftargets.append(target)
+                if line[6:37] == 'Threshold for SCF energy change':
```
langner: I question the robustness of using specific indices/slices when the string is so specific. It is unlikely to occur anywhere else, so maybe it is sufficient here to use `if 'very unique string' in line` in such cases?

cclib/parser/molcasparser.py
```diff
@@ -222,38 +222,41 @@ def extract(self, inputfile, line):
...
+                    target = float(line.split()[-1])
+                    scftargets.append(target)
 
-            line = next(inputfile)
+                line = next(inputfile)
 
-            if line[6:34] == 'Threshold for density matrix':
-                target = float(line.split()[-1])
-                scftargets.append(target)
+                if line[6:34] == 'Threshold for density matrix':
```
langner: Ditto.

cclib/parser/molcasparser.py
```diff
@@ -222,38 +222,41 @@ def extract(self, inputfile, line):
...
+                    target = float(line.split()[-1])
+                    scftargets.append(target)
 
-            line = next(inputfile)
+                line = next(inputfile)
 
-            if line[6:31] == 'Threshold for Fock matrix':
-                target = float(line.split()[-1])
-                scftargets.append(target)
+                if line[6:31] == 'Threshold for Fock matrix':
```
langner: Ditto, here and below.

cclib/parser/molcasparser.py
```diff
@@ -696,6 +699,32 @@ def extract(self, inputfile, line):
...
+        # ...
+        #          14          -75.01515915      -0.05070274      -0.00000029
+        #          15          -75.01515929      -0.05070289      -0.00000014
+        #          16          -75.01515936      -0.05070296      -0.00000007
+        #       Convergence after                    17  Iterations
+        #
+        #
+        #      Total energy (diff) :     -75.01515936      -0.00000007
+        #      Correlation energy  :        -0.0507029554992
+        if line.startswith('()'):
```
langner: Is it possible to trigger on the text in the previous line instead? That seems more appropriate and clearly related to CC.

cclib/parser/molcasparser.py
kunalsharma05: Yes, agreed!

cclib/parser/molcasparser.py
kunalsharma05: Yeah! Will make the change.

cclib/parser/molcasparser.py
kunalsharma05: Yeah. That way we can computation since all modules print `()()()`.

cclib/parser/molcasparser.py
```diff
@@ -222,38 +222,41 @@ def extract(self, inputfile, line):
...
+            if line.strip().startswith('SCF'):
+                scftargets = []
+                self.skip_lines(inputfile,
+                    ['Minimized', 'Number', 'Maximum', 'Maximum'])
                 line = next(inputfile)
 
-            if line[6:37] == 'Threshold for SCF energy change':
-                target = float(line.split()[-1])
-                scftargets.append(target)
+                if 'Threshold for SCF energy change' in line:
```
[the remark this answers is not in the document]
langner: +1

kunalsharma05:
@ATenderholt @langner Made the requested changes. Please have a look.

cclib/parser/molcasparser.py
```diff
@@ -222,36 +222,19 @@ def extract(self, inputfile, line):
...
-
-            if line[6:31] == 'Threshold for Norm(delta)':
-                target = float(line.split()[-1])
-                scftargets.append(target)
-
-            self.append_attribute('scftargets', scftargets)
+            if line.strip().startswith('SCF'):
+                scftargets = []
+                self.skip_lines(inputfile,
+                    ['Minimized', 'Number', 'Maximum', 'Maximum'])
```
berquist: Something is wrong with the indentation here.

cclib/parser/molcasparser.py
```diff
@@ -222,36 +222,19 @@ def extract(self, inputfile, line):
...
+                lines = [next(inputfile) for i in range(7)]
+                targets = ['Threshold for SCF energy change', 'Threshold for density matrix', 'Threshold for Fock matrix',
+                                    'Threshold for Norm(delta)']
```
berquist: Something is wrong with the indentation here. Also, the top line is too long. Try to either do
```python
targets = [
    'apple',
    'banana',
    'pear',
]
```
or
```python
targets = ['apple',
           'banana',
           'pear']
```
I don't think we've standardized on one or the other.

PR #575, opened 2018-07-27
ORCA: update unit test files to use v4.0 and move v2.9 to regressions

berquist:
Recreation of https://github.com/cclib/cclib/pull/394, which had aged out.


ATenderholt:
Is there a corresponding cclib-data PR for the v2.9 files?

ATenderholt:
Oh, I see it's already been merged.

langner:
Thanks for doing the move-update dance here, so that the history is clearer!

Issue #484, opened 2018-03-24
Get raw CJSON dict


jaimergp:
Would it be posible to obtain a raw CJSON dict out of `cclib.io.CJSON`? The current implementation can only return the dumped str with `json.dumps`, but I would like to obtain the raw Python object. Currently I have to call `json.loads` on the returned string.

I'd suggest a small refactor similar to this:

```
class CJSON(filewriter.Writer):

    [...]

    def _cjson_dict(self):
         cjson_dict = dict()
         # all the code present currently in self.generate_repr, sans the return statements
         # extracted below
         return cjson_dict

    def generate_repr(self):
        cjson_dict = self._cjson_dict()
        if self.terse:
            return json.dumps(cjson_dict, cls=NumpyAwareJSONEncoder)
        else:
            return json.dumps(cjson_dict, cls=JSONIndentEncoder, sort_keys=True, indent=4)

    [...]

```
Would that be OK? I can submit a PR if you want.

ATenderholt:
That seems reasonable to me, although `_cjson_dict` implies a non-public API according to PEP8. I suggest something like `as_dict()` or a property (`object`?) instead.

langner:
Agreed, would be nice to have access to the dict. Would also be nice to merge the reader and writer into one class.

jaimergp:
Finally, I have some free time! How should we proceed with this API merge? Would it make sense to first refactor the CJSON dumper for direct access to the dict and then think about the reader/writer merge?

berquist:
@jaimergp, do you want to work on merging readers and writers, or should we do that?

jaimergp:
@berquist I can handle that if you want! I mean, I don't want to interfere but I am happy to help :) That said, I'd need some guidelines to see how you want to abstract the interface for more general cases.

langner:
I think that's not strictly necessary for this particular change, namely returning the dict, but it would be great as well. Also, please use `as_dict` as Adam suggested or similar.

berquist:
Yes, I think we would appreciate the help. My strategy was going to start with making an abstract `FileFormat` class or something similar.

PR #578, opened 2018-08-01
Refactor CJSON writer so the Python dict is directly available

jaimergp:
Following discussion in issue  #484.

I named the method `as_dict()`, as suggested by @ATenderholt, but maybe a property is better.

### DIFF
--- a/cclib/io/cjsonwriter.py
+++ b/cclib/io/cjsonwriter.py
@@ -38,9 +38,8 @@ class CJSON(filewriter.Writer):
         name = os.path.basename(os.path.splitext(path)[0])
         return name
 
-    def generate_repr(self):
-        """Generate the CJSON representation of the logfile data."""
-
+    def as_dict(self):
+        """ Build a Python dict with the CJSON data"""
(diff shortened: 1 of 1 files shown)


berquist:
I think making properties is only possible once we combine readers and writers, no?

ATenderholt:
Documentation says that `property()` used as a decorator "...makes it possible to create read-only properties." So I think it's reasonable without combining readers and writers.

Question is what would a property bring us? I'm not sure there is much benefit to being able to do `writer.dict` (or whatever) instead of `writer.as_dict()` since we presumably wouldn't be using the setter and deleter.

It might make sense to have something like `cjsondict` as property of `ccData` (lazily computed) that could then be used by a reader/writer, but that's a bit of an undertaking for the reader and corresponding setter.

berquist:
A possible setter could try multiple ways to see if the argument could be converted into the internal representation. It could be a file name or URL, a `ccData` instance, a string or "rendered" representation, and so on, depending on how flexible you want it to be.

PR #583, opened 2018-08-04
docs: automatically update website on merge commit to master

berquist:
This is the final part of https://github.com/cclib/cclib.github.io/issues/32.

travis-sphinx can't be used to build the docs, only deploy, because our Makefile also clones and tests release and development versions of cclib. After some communication with the author on other topics, and having written working deploy code without it, maybe travis-sphinx should be dropped.


langner:
> maybe travis-sphinx should be dropped

Sounds good, and looks like deploy uses travis-sphinx but build does not... is there something I'm missing related to this "maybe"?

Have you tested this? The script seems to be returning a bad exit code. From Travis:
```
The command "bash travis/build_docs.sh" exited with 1.
$ if [[ "${TRAVIS_BRANCH}" == master && "${TRAVIS_PULL_REQUEST}" == false ]];
then
    bash travis/deploy_docs_travis-sphinx.sh
fi
```

langner:
Would be great to get this in before v1.5.4 - @berquist do you have time to take another look? If not, I can poke around here later in the week.

berquist:
Yes, I'll get to it.

berquist:
I think it will work now...?

Issue #577, opened 2018-07-28
Drop Python 3.2 and bump Travis to 3.4


berquist:
I've run into problems in https://github.com/cclib/cclib/issues/455 and https://github.com/cclib/cclib/pull/568 with Python 3.2 not being supported properly. Even Ubuntu 14.04.5 LTS, released in 2014 and supported until 2019, ships 3.4.3. I propose that we switch to 3.4 as the minimum version for 3.x.

ATenderholt:
Seems like a good idea to me.

langner:
+1 but let's please update the carious files/docs mentioning the requirements

berquist:
Well, now it's much easier to update the docs and the config simultaneously.

langner:
> Well, now it's much easier to update the docs and the config simultaneously.

And kudos for that!

langner:
Done.

PR #584, opened 2018-08-04
Update minimum required 3.x version from 3.2 to 3.4

berquist:
This handles https://github.com/cclib/cclib/issues/577.

### DIFF
--- a/doc/sphinx/how_to_install.rst
+++ b/doc/sphinx/how_to_install.rst
@@ -7,7 +7,7 @@ Requirements
 ------------
 
 Before you install cclib, you need to make sure that you have the following:
- * Python (at least version 3.0 is recommended, although 2.7 is still tested)
+ * Python (at least version 3.4 is recommended, although 2.7 is still tested)
  * NumPy (at least version 1.5 is recommended)
 
 Python is an open-source programming language available from http://www.python.org. It is available for Windows as well as being included in most Linux distributions. In Debian/Ubuntu it is installed as follows (as root):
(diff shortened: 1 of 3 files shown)


doc/sphinx/how_to_install.rst
```diff
@@ -7,7 +7,7 @@ Requirements
 ------------
 
 Before you install cclib, you need to make sure that you have the following:
- * Python (at least version 3.0 is recommended, although 2.7 is still tested)
+ * Python (at least version 3.4 is recommended, although 2.7 is still tested)
```
langner: Is "recommended" the right word here? Will something break with version <3.4?

berquist: It means that 3.0, 3.1, 3.2, and 3.3 are not supported, but 2.7 is, and we prefer >= 3.4 over 2.7.

How about:

> Before you install cclib, you need to make sure that you have the following:
> * Python (at least version 3.4, although 2.7 is still tested)
> * NumPy (at least version 1.12)

That's the new minimum version of NumPy.

langner: Well, I'm trying to understand if cclib will still work for uers with Python < 3.4. If so, then "make sure that you have" is not the right wording. IIUC we want >= 3.4 for our own convenience.

berquist: If we limit `periodictable` to the `Nuclear` method, then only that part would require 3.4. If the built-in periodic table gets replaced with `periodictable`, then all of our code will require 3.4.

langner: OK, I guess I'd prefer to make this recommended, not required. I like that cclib will still work with older Python version, even if some parts won't and we don't test older version (although we could test and exclude some modules for those versions).

Since we do have a number of "soft" dependencies, it'd be good to standardize the way we try to import them if we don't already. There are some options here, as discussed https://stackoverflow.com/questions/14050281/how-to-check-if-a-python-module-exists-without-importing-it/25045228.

berquist: I've made a separate issue for the imports.

Even if it's possible, I'd rather not include language that implies we can run on a version of Python that hasn't received a security update in 4 years.

langner: OK

PR #591, opened 2018-08-08
Added the geometry optimization files for Turbomole 7.2

kunalsharma05:
Please have a look at my comments and questions about this calculation.


berquist:
These folders were accidentally added as submodules because they have `.git` folders in them. If you delete the `.git` folders and re-add, it should be fine.

berquist:
Actually, I also need to check the energies; one of these is correct for the unit tests and the other should go in regressions.

berquist:
There is still a problem with one of the directories being added as a submodule because it contains a `.git` folder.

I also think the one with "Gaussian" in the name should be moved to regressions.

langner:
> I also think the one with "Gaussian" in the name should be moved to regressions.

+1 let's leave just one of each calculation type in the unit test suite

langner:
@kunalsharma05 Any updates here?

kunalsharma05:
@langner @berquist Okay, will move the gaussian file to regressions.

langner:
Since Kunal is no longer active, I'm going to break this up into separate PRs. I'm going to download the files and re-add them, so unfortunately we'll lose the original authorship, because everything was added in one commit (along with the submodule thingy).

First one is the regression file suffixed with Gaussian: cclib/cclib-data#94

langner:
Hmmm... I don't actually see any files in the non-Gaussian directory when I check out the source branch of this PR (@kunalsharma05 please correct me if I'm wrong). Which means I can close this PR. We'll need to get the Turbomole gopt output for unit tests from somewhere else.

Issue #533, opened 2018-06-11
Ensure metadata contains most detailed package version


berquist:
As mentioned in https://github.com/cclib/cclib/pull/515#discussion_r194088771, each parser should set `self.metadata['package_version']` to be the most specific version possible.

berquist:
The "most specific version possible" will be within reason. For example, packages that print a value that is very clearly the version control commit number or hash (DALTON) should read that separately. Some are less clear (GAMESS-UK, Jaguar) and those can be added to the primary package version.

berquist:
There is a line in the ORCA parser that reads
```python
# Clashes with Orca 2.6 (and presumably before) TDDFT absorption spectrum printing
elif line == 'ABSORPTION SPECTRUM' and float(self.metadata['package_version']) > 2.6:
```
which gives a better definition of what `package_version` should contain: even if it cannot be cast to a number, it should be a string with a clear ordering so comparisons can easily be performed. [Dictionary order](https://en.wikipedia.org/wiki/Lexicographical_order) is probably sufficient. This clearly requires things like Git revision hashes to be separate fields, but does not exclude things like build numbers, SVN revision numbers, or counters that start from 0 or 1.

However, I believe this should be simplified so that `float(package_version)` always works, otherwise it will be inconsistent for people modifying parser code. _All_ more specific version information that isn't like 1.2.3 should go in some other metadata.

Thoughts, @dgasmith? I don't recall the current schema spec containing these details.

dgasmith:
I would go with something where you can use `pkg_resources`:

```
>>> v1 = pkg_resources.parse_version("0.2.2+12.g41eebf2")
>>> v2 = pkg_resources.parse_version("0.2.2")
>>> v2 > v1
False
```

berquist:
The way I would set it up is that the parser would read the version and save it right away after calling `parse_version` on it, so that it can be used for version comparisons in `extract()` without much work:
```python
self.metadata["package_version"] = pkg_resources.parse_version(package_version)
```
and `testtypes` in https://github.com/cclib/cclib/pull/599 would be changed to check for `packaging.version.Version`:
```python
In [10]: s2 = "1.2.1"

In [11]: v2 = pkg_resources.parse_version(s2)

In [12]: type(v2)
Out[12]: packaging.version.Version
```
Is this sufficient?

langner:
[...]

PR #601, opened 2018-08-25
Psi4: add unit test files for 1.2.1

berquist:


langner:
Thanks, looks good! We now have two Psi4 versions, so we don't need to move anything to regressions.

berquist:
The tests will be enabled once the parser is updated and https://github.com/cclib/cclib/issues/533 is completely handled, since comparing package versions is necessary.

PR #605, opened 2018-08-27
Add method to calculate electric multipole moments

### LINKED ISSUE #311: Calculate dipole (or other moments) from atomic charges & positions?
Should there be a calculation method to estimate the dipole moment from atomic coordinates and charges when available? I notice, for example that NWChem (and possibly other programs) don't provide moments for every output file.

It's simple code + using the unit converter.

xymaxim:
This adds a new `Moments` method and closes #311. It allows to calculate dipole and traceless quadrupole moments using parsed atomic charges and coordinates. Using input parameters, the choice of origin, the type of population analysis, and the necessity of updating parsed data object with calculated values can be specified.

### DIFF
--- a/cclib/method/moments.py
+++ b/cclib/method/moments.py
@@ -72,12 +72,12 @@ class Moments(Method):
         Keyword Args:
             masses - if None, then use default atomic masses. Otherwise,
                 the user-provided will be used.
-            decimals - a number of decimal places that atomic charges are
-                to be truncated to without rounding.
 
         Returns:
             A list where the first element is the origin of
-            coordinates, while other elements are multipole moments.
+            coordinates, while other elements are dipole and
+            quadrupole moments expressed in terms of Debye and
+            Buckingham units respectively.
 
         """
         coords = self.data.atomcoords[-1]
--- a/cclib/method/moments.py
+++ b/cclib/method/moments.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2018, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
+
+"""Calculation of electric multipole moments based on data parsed by cclib."""
+
+import numpy
+
+from cclib.parser.utils import convertor
+from cclib.method.calculationmethod import Method
+
+
+class Moments(Method):
+    def __init__(self, data):
+        self.required_attrs = ('atomcoords', 'atomcharges')
+
+        super(Moments, self).__init__(data)
+
+    def __str__(self):
+        """Returns a string representation of the object."""
+        return "Multipole moments of %s" % (self.data)
+
+    def __repr__(self):
+        """Returns a representation of the object."""
+        return 'Moments("%s")' % (self.data)
+
+    def calculate(self, gauge='nuccharge', population='mulliken', **kwargs):
+        """Calculate electric dipole moment using parsed partial atomic
+        charges.
+        
+        Args:
+            gauge - a choice of gauge origin. Can be either a string or
+                a three-element iterable. If iterable, then it
+                explicitly defines the origin. If string, then the
+                value can be any one of the following and it describes
+                what is used as the origin:
+                    * 'nuccharge' -- center of positive nuclear charge
+                    * 'mass' -- center of mass
+
+        Keyword Args:
+            masses - if None, then use default atomic masses. Otherwise,
+                the user-provided will be used.
+            decimals - a number of decimal places that atomic charges are
+                to be truncated to without rounding.
+
+        Returns:
+            A list where the first element is the origin of coordinates,
+            while other elements are multipole moments expressed in Debye
+            units.
+        """
+        coords = self.data.atomcoords[-1]
+        try:
+            charges = self.data.atomcharges[population]
+        except KeyError:
+            raise ValueError
+
+        if hasattr(gauge, '__iter__') and not isinstance(gauge, str):
+            origin = numpy.asarray(gauge)
+        elif gauge == 'nuccharge':
+            origin = numpy.average(coords, weights=self.data.atomnos, axis=0)
+        elif gauge == 'mass':
+            if 'masses' in kwargs:
+                atommasses = numpy.asarray(kwargs.get('masses'))
+            else:
+                atommasses = self.data.atommasses
+            origin = numpy.average(coords, weights=atommasses, axis=0)
+        else:
+            raise ValueError
+
+        one_debye = convertor(1, 'Angstrom', 'bohr') * convertor(1, 'ebohr', 'Debye')
+        if self.data.charge == 0:
+            dipole = numpy.dot(charges, coords) * one_debye
+        else:
+            dipole = numpy.dot(charges, coords - origin) * one_debye
+
+        moments = [origin, dipole]
+        return moments
--- a/cclib/method/moments.py
+++ b/cclib/method/moments.py
@@ -27,16 +27,45 @@ class Moments(Method):
         """Returns a representation of the object."""
         return 'Moments("%s")' % (self.data)
 
+    def _calculate_dipole(self, charges, coords, origin):
+        coords_au = convertor(coords, 'Angstrom', 'bohr')
+        
+        if self.data.charge == 0:
+            dipole = numpy.dot(charges, coords_au)
+        else:
+            origin_au = convertor(origin, 'Angstrom', 'bohr')
+            dipole = numpy.dot(charges, coords_au - origin_au)
+            
+        return convertor(dipole, 'ebohr', 'Debye')
+
+    def _calculate_quadrupole(self, charges, coords, origin):
+        transl_coords_au = convertor(coords - origin, 'Angstrom', 'bohr')
+
+        delta = numpy.eye(3)
+        Q = numpy.zeros([3,3])
+        for i in range(3):
+            for j in range(3):
+                for q, r in zip(charges, transl_coords_au):
+                    Q[i,j] += q * (3 * r[i] * r[j] - \
+                              numpy.linalg.norm(r)**2 * delta[i][j])
+
+        triu_idxs = numpy.triu_indices_from(Q)
+        raveled_idxs = numpy.ravel_multi_index(triu_idxs, Q.shape)
+        quadrupole = numpy.take(Q.flatten(), raveled_idxs)
+        
+        return convertor(quadrupole, 'ebohr2', 'Buckingham')
+    
     def calculate(self, gauge='nuccharge', population='mulliken', **kwargs):
-        """Calculate electric dipole moment using parsed partial atomic
-        charges.
+        """Calculate electric dipole and quadrupole moments using parsed
+        partial atomic charges.
         
         Args:
-            gauge - a choice of gauge origin. Can be either a string or
-                a three-element iterable. If iterable, then it
-                explicitly defines the origin. If string, then the
-                value can be any one of the following and it describes
-                what is used as the origin:
+            gauge - a choice of gauge origin. Can be either a string
+                or a three-element iterable. If iterable, then it
+                explicitly defines the coordinate system origin (in
+                Angstrom).  If string, then the value can be any one
+                of the following and it describes what is used as the
+                origin:
                     * 'nuccharge' -- center of positive nuclear charge
                     * 'mass' -- center of mass
 
@@ -47,9 +76,9 @@ class Moments(Method):
                 to be truncated to without rounding.
 
         Returns:
-            A list where the first element is the origin of coordinates,
-            while other elements are multipole moments expressed in Debye
-            units.
+            A list where the first element is the origin of
+            coordinates, while other elements are multipole moments.
+
         """
         coords = self.data.atomcoords[-1]
         try:
@@ -70,11 +99,8 @@ class Moments(Method):
         else:
             raise ValueError
 
-        one_debye = convertor(1, 'Angstrom', 'bohr') * convertor(1, 'ebohr', 'Debye')
-        if self.data.charge == 0:
-            dipole = numpy.dot(charges, coords) * one_debye
-        else:
-            dipole = numpy.dot(charges, coords - origin) * one_debye
+        dipole = self._calculate_dipole(charges, coords, origin)
+        quadrupole = self._calculate_quadrupole(charges, coords, origin)
 
-        moments = [origin, dipole]
+        moments = [origin, dipole, quadrupole]
         return moments
--- a/cclib/method/moments.py
+++ b/cclib/method/moments.py
@@ -72,6 +72,9 @@ class Moments(Method):
         Keyword Args:
             masses - if None, then use default atomic masses. Otherwise,
                 the user-provided will be used.
+            overwrite - it defines whether to modify the data object or
+                just return results. If `moments` attribute already exists,
+                the old value will be fully replaced. 
 
         Returns:
             A list where the first element is the origin of
@@ -101,6 +104,13 @@ class Moments(Method):
 
         dipole = self._calculate_dipole(charges, coords, origin)
         quadrupole = self._calculate_quadrupole(charges, coords, origin)
-
         moments = [origin, dipole, quadrupole]
+        
+        if 'overwrite' in kwargs:
+            if hasattr(self.data, 'moments'):
+                msg = ("Overwriting existing moments attribute values "
+                       "with newly calculated ones.")
+                self.logger.info(msg)
+            self.data.moments = moments
+
         return moments
--- a/test/method/testmoments.py
+++ b/test/method/testmoments.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2018, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
+
+"""Test the Moments method in cclib"""
+
+from __future__ import print_function
+
+import sys
+import os
+import logging
+import unittest
+
+import numpy
+from numpy.testing import assert_almost_equal
+
+from cclib.method import Moments, MPA
+from cclib.parser import Gaussian
+
+sys.path.insert(1, "..")
+
+from ..test_data import getdatafile
+
+
+class MomentsTest(unittest.TestCase):
+    def test_results(self):
+        data, _ = getdatafile(Gaussian, "basicGaussian16", ["water_mp2.log"])
+        x = Moments(data).calculate()
+        assert_almost_equal(x, [[0,0,0], [0,0, -0.91543]], 5)
+
+    def test_origin_displacement(self):
+        data, _ = getdatafile(Gaussian, "basicGaussian16", ["water_mp2.log"])
+        x = Moments(data).calculate()
+        y = Moments(data).calculate(gauge=(2e7,18,28))
+        assert_almost_equal(x[1], y[1])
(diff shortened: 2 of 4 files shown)


cclib/method/moments.py
```diff
@@ -0,0 +1,116 @@
...
+                or a three-element iterable. If iterable, then it
+                explicitly defines the coordinate system origin (in
+                Angstrom).  If string, then the value can be any one
+                of the following and it describes what is used as the
+                origin:
+                    * 'nuccharge' -- center of positive nuclear charge
+                    * 'mass' -- center of mass
+
+        Keyword Args:
+            masses - if None, then use default atomic masses. Otherwise,
```
langner: I would prefer to make these normal arguments with default values. Is there a benefit from stuffing them into `kwargs`?

cclib/method/moments.py
```diff
@@ -0,0 +1,116 @@
...
+        
+        Args:
+            gauge - a choice of gauge origin. Can be either a string
+                or a three-element iterable. If iterable, then it
+                explicitly defines the coordinate system origin (in
+                Angstrom).  If string, then the value can be any one
+                of the following and it describes what is used as the
+                origin:
+                    * 'nuccharge' -- center of positive nuclear charge
+                    * 'mass' -- center of mass
```
langner: Missing the `population` argument here.

cclib/method/moments.py
```diff
@@ -0,0 +1,116 @@
...
+        quadrupole = numpy.take(Q.flatten(), raveled_idxs)
+        
+        return convertor(quadrupole, 'ebohr2', 'Buckingham')
+    
+    def calculate(self, gauge='nuccharge', population='mulliken', **kwargs):
+        """Calculate electric dipole and quadrupole moments using parsed
+        partial atomic charges.
+        
+        Args:
+            gauge - a choice of gauge origin. Can be either a string
```
langner: Is `gauge` really the right word here? Is `com` (center of mass) sufficient?

cclib/method/moments.py
```diff
@@ -0,0 +1,116 @@
...
+
+"""Calculation of electric multipole moments based on data parsed by cclib."""
+
+import numpy
+
+from cclib.parser.utils import convertor
+from cclib.method.calculationmethod import Method
+
+
+class Moments(Method):
```
langner: So, we parse moments already from logfiles, when they are present. The method is useful for situations where moments are not printed (and calculating them at different COMs). Do you have any suggestions about how to expose these two things in a consistent way? I realize you may no have context for this - @berquist @ATenderholt any thoughts?

cclib/method/moments.py
```diff
@@ -0,0 +1,116 @@
...
+                              numpy.linalg.norm(r)**2 * delta[i][j])
+
+        triu_idxs = numpy.triu_indices_from(Q)
+        raveled_idxs = numpy.ravel_multi_index(triu_idxs, Q.shape)
+        quadrupole = numpy.take(Q.flatten(), raveled_idxs)
+        
+        return convertor(quadrupole, 'ebohr2', 'Buckingham')
+    
+    def calculate(self, origin='nuccharge', population='mulliken',
+                  masses=None, overwrite=False):
```
xymaxim: `overwrite` is also not a good name here since the output file might not contain multipole moments. Maybe `update` (a data object) or `save` (results) is much better.

cclib/method/moments.py
xymaxim: Yes, this could potentially cause confusions during further data inspection or comparison. Personally, I see a few practical weaknesses with the idea of updating the data object: (1) does it make any sense to compare parsed results and calculated ones coming from available partial atomic charges; (2) how to determine which type of atomic charges is used to evaluate moments; (3) there is no way to check is it parsed or calculated values.

cclib/method/moments.py
berquist: I like `save`, the reason being that other methods (like MPA) store their results as attributes in the `Method` instance, not the `ccData` instance.

cclib/method/moments.py
```diff
@@ -0,0 +1,116 @@
...
+        dipole = self._calculate_dipole(charges, coords, origin_pos)
+        quadrupole = self._calculate_quadrupole(charges, coords, origin_pos)
+        moments = [origin_pos, dipole, quadrupole]
+        
+        if overwrite:
+            if hasattr(self.data, 'moments'):
+                msg = ("Overwriting existing moments attribute values "
+                       "with newly calculated ones.")
+                self.logger.info(msg)
+            self.data.moments = moments
```
berquist: As implied by my other comment, I'd rather this be
```python
selfattr(self, 'moments', moments)
```
or, more explicitly,
```python
setattr(self, 'origin', origin)
setattr(self, 'dipole', dipole)
setattr(self, 'quadrupole', quadrupole)
```

cclib/method/moments.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __str__(self):
+        """Returns a string representation of the object."""
+        return "Multipole moments of %s" % (self.data)
+
+    def __repr__(self):
+        """Returns a representation of the object."""
+        return 'Moments("%s")' % (self.data)
+
+    def _calculate_dipole(self, charges, coords, origin):
```
berquist: Even though this is private, please add a docstring.

cclib/method/moments.py
```diff
@@ -0,0 +1,116 @@
...
+        
+        if self.data.charge == 0:
+            dipole = numpy.dot(charges, coords_au)
+        else:
+            origin_au = convertor(origin, 'Angstrom', 'bohr')
+            dipole = numpy.dot(charges, coords_au - origin_au)
+            
+        return convertor(dipole, 'ebohr', 'Debye')
+
+    def _calculate_quadrupole(self, charges, coords, origin):
```
berquist: Even though this is private, please add a docstring. It should include a comment that this is the _traceless_ quadrupole tensor.

cclib/method/moments.py
berquist: 1. Sure, why not? They might be a decent approximation if the partial charge scheme is good. However, you're correct that they're fundamentally different.

Because they're fundamentally different, plus your point 3, plus you might want to compare different partial charge schemes (Mulliken, Lowdin, GAPT, ChElPG, ...), it doesn't make sense to overwrite the parsed moments in`ccData`, but at most store them in the `Method` object.

cclib/method/moments.py
xymaxim: In this case, it seems there is no need for `save` argument.

cclib/method/moments.py
```diff
@@ -0,0 +1,125 @@
...
+        atomic charges and their coordinates with respect to the origin.
+        """
+        transl_coords_au = convertor(coords - origin, 'Angstrom', 'bohr')
+
+        delta = numpy.eye(3)
+        Q = numpy.zeros([3, 3])
+        for i in range(3):
+            for j in range(3):
+                for q, r in zip(charges, transl_coords_au):
+                    Q[i,j] += 1/2 * q * (3 * r[i] * r[j] - \
```
xymaxim: Please notice that I changed the way to calculate the (traceless) quadrupole moment tensor Q to make it consistent with the Buckingham definition. There is another definition (it differs by multiplication by 2), e.g., given on p. 99 of section 41 in the reference below:

L.D. Landau and E.M. Lifshitz, The Classical Theory of Fields (4th ed.), Pergamon Press, Oxford (1994).

berquist: Could you state that as a short comment in the code?

Also, would you be able to find the DOI for the Buckingham reference and add it?

cclib/method/moments.py
```diff
@@ -0,0 +1,125 @@
...
+        if hasattr(origin, '__iter__') and not isinstance(origin, str):
+            origin_pos = numpy.asarray(origin)
+        elif origin == 'nuccharge':
+            origin_pos = numpy.average(coords, weights=self.data.atomnos, axis=0)
+        elif origin == 'mass':
+            if masses:
+                atommasses = numpy.asarray(masses)
+            else:
+                atommasses = self.data.atommasses
+            origin_pos = numpy.average(coords, weights=atommasses, axis=0)
```
berquist: I didn't know you could do this in a one-liner, but it makes sense. +1

test/method/testmoments.py
```diff
@@ -0,0 +1,70 @@
...
+        data, _ = getdatafile(Gaussian, "basicGaussian16", ["water_mp2.log"])
+        x = Moments(data).calculate()
+        assert_almost_equal(x[0], [0, 0, 0], 5)
+        assert_almost_equal(x[1], [0, 0, -0.91543], 5)
+        assert_almost_equal(x[2][0] + x[2][3] + x[2][5], 0)
+
+    def test_origin_displacement(self):
+        data, _ = getdatafile(Gaussian, "basicGaussian16", ["water_mp2.log"])
+        x = Moments(data).calculate()
+        y = Moments(data).calculate(origin=(2e7,18,28))
```
berquist: Can you make sure all of the lines in the test file follow PEP 8? The method itself looks ok.

xymaxim:
While working on it, I'm faced with a problem of managing the truncated values printed in the output files. This affects the obtained origin position and multipole moment values. 

For example, let's take the water molecule and[ the output](https://github.com/cclib/cclib/blob/master/data/Gaussian/basicGaussian16/water_mp2.log) from Gaussian 16. The printed atomic charges and coordinates have 6 decimal places accuracy. The center of nuclear charge calculated by us is located at `(0, 0, 2e-07)`.  Since the sum of charges (`[0.319889, 0.159944, 0.15994]`) is not zero, the origin displacement to (0.0, 0.1, 0.1) leads to the expected and nonsense changing of the dipole moment (if we don't consider using [if-else clause](https://github.com/xymaxim/cclib/blob/6b9e924f894d5e7251e75d307f9f7a263577496c/cclib/method/moments.py#L41) from this pull request; charged systems are also prone to this problem).

I created another (sandbox) [branch](https://github.com/xymaxim/cclib/tree/handle-trunc-val) to illustrate the problem and proposed solution. I looked at [the largest remainder method](https://stackoverflow.com/a/13483710/5896272) (how to deal with the symmetry?) at first, but finally stops on [the another way](https://github.com/xymaxim/cclib/blob/1c5c7e168070b084420491ef3dcfedb96f70ee34/cclib/method/moments.py#L35) to compensate the remainder equally between all charges.

The tests passed successfully for small and large (1e+07 <= x <= 1e+09) (see comments in [`testmoments.py`](https://github.com/xymaxim/cclib/blob/1c5c7e168070b084420491ef3dcfedb96f70ee34/test/method/testmoments.py)) origin displacements. If we consider different-sized systems of randomly sampled charges (I tried to make it carefully but it needs another pair of eyes) and the case of super large (> 1e+09) translation, it fails periodically while testing up to 6 decimals.

cclib/method/moments.py
langner: Sure.

cclib/method/moments.py
langner: It would be ideal to have both versions.

xymaxim:
Sorry, the minus sign was accidentally omitted—it should read as [-0.319889, 0.159944, 0.15994]. Due to round-off error, the molecule have a non-zero net charge.

berquist:
This is very interesting. At the risk of adding complexity, I'd like the option to turn `ensure_charges` (the largest remainder method) off, but I'm ok with having it on by default. As a "pro user", I've always known that round-off causes strange errors and just took it for granted (the error in the total charge should be less than 1e-5 or something).

In terms of final results, do you think the error introduced by the non-zero total charge is anywhere near the magnitude the error of moments calculated from partial charge schemes, rather than _ab initio_?

langner:
I'm OK with the state of this PR now, but please do address Eric's questions. They're interesting, although some, I think, go outside the scope of this PR.

xymaxim:
> I would suggest testing the method with idealized inputs first. We can also test using logfiles, and in that case you'll unfortunately need to check for "almost equals". If you think there's some interesting edge case with extreme values caused by precision, feel to add that as a test as well.

Intuitively, I'm all for it. So I made some changes according to your suggestions. I'd ask you to check it additionally. The method so far is covered with tests without handling potential round-off errors.

For me, `assert_almost_equal()` is not enough option if we parse actual data. The accuracy for "non-zero total charge" cases is also connected with the possibility to provide custom position of the origin. We know that usually it's located inside the system of charges and my initial intention to accept iterable as `origin` argument value was for setting (0, 0, 0) point if needed. The extreme values very rarely, if ever, be the case. We can enlarge the predefined values of `origin` with zero point and don't accept custom values. How about this?

cclib/method/moments.py
xymaxim: @berquist Done it. I added another reference mentioned by Buckingham, 1959 in his article.

@langner There's two options from me: one, add optional argument, say, `quad` with 'buckingham' and 'hirschfelder' acceptable values. Or two, leave it as is. After reading docstring the end-user can decide for himself—to multiply or not to multiply.

cclib/method/moments.py
```diff
@@ -0,0 +1,131 @@
...
+"""Calculation of electric multipole moments based on data parsed by cclib."""
+
+import numpy
+
+from cclib.parser.utils import convertor
+from cclib.method.calculationmethod import Method
+
+
+class Moments(Method):
+    """This class calculates multipole moments and stores them in
```
langner: Could you change this into a one-line summary and a separate paragraph for everything else? Please see PEP 257 (https://www.python.org/dev/peps/pep-0257/) for generally accepted docstring style conventions.

cclib/method/moments.py
langner: That's fine.

cclib/method/moments.py
```diff
@@ -0,0 +1,131 @@
...
+            Quarterly Reviews, Chemical Society, 13(3), 183.
+            https://doi.org:10.1039/qr9591300183.
+         .. [2] Hirschfelder J. O., Curtiss C. F. and Bird R. B. (1954).
+            The Molecular Theory of Gases and Liquids. New York: Wiley.
+        """
+        coords = self.data.atomcoords[-1]
+        try:
+            charges = self.data.atomcharges[population]
+        except KeyError:
+            raise ValueError
```
langner: Why is it useful to catch this error and recast it as a different type? If we want to go in this direction, please make a custom exception type and use that here for clarity.

Also, consider doing `except KeyError as e` and pass `e` along to the new exception.

Also, please document this in a Raises section of the docstring.

cclib/method/moments.py
```diff
@@ -0,0 +1,131 @@
...
+            else:
+                try:
+                    atommasses = self.data.atommasses
+                except AttributeError:
+                    raise ValueError(
+                        "atomic masses were not parsed, consider provide "
+                        "'masses' argument instead")
+            origin_pos = numpy.average(coords, weights=atommasses, axis=0)
+        else:
+            raise ValueError
```
langner: Please also the same custom exception here, maybe call it `CclibMomentError`?

cclib/method/moments.py
```diff
@@ -0,0 +1,131 @@
...
+         .. [2] Hirschfelder J. O., Curtiss C. F. and Bird R. B. (1954).
+            The Molecular Theory of Gases and Liquids. New York: Wiley.
+        """
+        coords = self.data.atomcoords[-1]
+        try:
+            charges = self.data.atomcharges[population]
+        except KeyError:
+            raise ValueError
+
+        if hasattr(origin, '__iter__') and not isinstance(origin, str):
```
langner: Please use isinstnace(origin, collections.Iterable) for detecting the iterable.

test/method/testmoments.py
```diff
@@ -0,0 +1,133 @@
...
+import numpy as np
+from numpy.testing import assert_equal, assert_almost_equal
+
+from cclib.method import Moments
+
+
+class TestIdealizedInputs(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.linear_dipole_attrs = {
```
langner: Can't you just make this is a class or module level constant? Is there a need for `setUpClass`?

test/method/testmoments.py
```diff
@@ -0,0 +1,133 @@
...
+            'atomcoords': np.array([[[-1, 0, 0], [ 1,  0, 0]]]),
+            'atomcharges': {'mulliken': [-1, 1]},
+            'atomnos': [1, 1],
+        }
+    
+    @mock.patch('cclib.parser.ccData', spec=True)
```
langner: Do we really require a mock for this? Isn't it easy to just set up a real ccData object?

test/method/testmoments.py
```diff
@@ -0,0 +1,133 @@
...
+    def test_nonzero_quadrupole_moment(self, mock):
+        mock.atomcoords = np.array([[
+            [-1, 0, 0],
+            [0, 0, 0],
+            [1, 0, 0]]])
+        mock.atomcharges = {'mulliken': [1/2, -1, 1/2]}
+        mock.atomnos = np.ones(mock.atomcoords.shape[1])
+
+        x = Moments(mock).calculate()
+        assert np.count_nonzero(x[1]) == 0
```
langner: Please use one of the `self.assert*` method instead of raw asserts.

cclib/method/moments.py
xymaxim: Wouldn't `cclib.method.moments.CclibMomentError` name be somewhat stuttering? The `ValueError` from its [description](https://docs.python.org/3/library/exceptions.html#ValueError) doesn't fully suite our error case. Maybe `cclib.method.moments.ArgumentError`?

cclib/method/moments.py
xymaxim: > Why is it useful to catch this error and recast it as a different type? 

Since `KeyError`and `AttributeError` don't semantically describe the true cause of the error connected with the inappropriate inputs to our method.

berquist:
> We can enlarge the predefined values of origin with zero point and don't accept custom values.

I would always like the ability to pass custom values, even if we were to eventually add code that automatically minimizes the dipole (or other) moment for systems with non-zero charge.

My personal preference for scientific software is to have sensible defaults but as much flexibility as possible for those who want it.

berquist:
> although some, I think, go outside the scope of this PR.

Yes, this is nothing that should hold up the PR, but it's being brought up as a serious issue that I always took for granted, so I'd like to know if I should be more concerned scientifically than I have been.All the other parts of the PR I'm happy with.

xymaxim:
Eric earlier asked me:
> In terms of final results, do you think the error introduced by the non-zero total charge is anywhere near the magnitude the error of moments calculated from partial charge schemes, rather than ab initio?

I tried to analyze the worst-case accuracy of dot product between charges and their positions before, but stuck with the error cancellation due to numbers with opposite signs which minimizes the propagated error. I'll try to get a chance to finish and answer on your question directly somehow later.

Anyway, for now the user can be notified with warning message about the non-integer total charge.

xymaxim:
Oops, Travis can't run the tests on Python 2.7 due to missing mock library. I can make the following changes to fix it, Also `self.assertRaiseRegexp()` was renamed to `self.assertRaisesRegexp()` in version 3.2.

xymaxim:
Because of `ci skip` in commit message. Not a good idea, even I wanted fix that later after pushing all my planned changes not in one batch.

xymaxim:
Here are the changes formatted as a patch to resolve this problem because I can't figure out the proper way to keep commit history clean.

[0001-Adapt-files-to-run-Travis-tests-on-version-2.7.patch.zip](https://github.com/cclib/cclib/files/2386225/0001-Adapt-files-to-run-Travis-tests-on-version-2.7.patch.zip)

PR #609, opened 2018-09-01
docs: fix missing make target from #583

berquist:


berquist:
Not quite ready...

berquist:
Ok, I think it's ready.

langner:
OK, let's see...

PR #617, opened 2018-09-02
Start cleanup of CJSON-related methods

berquist:
The primary change is moving from camelCase to snake_case.


cclib/io/ccio.py
```diff
@@ -148,7 +148,7 @@ def guess_filetype(inputfile):
     return filetype
 
 
-def ccread(source, *args, **kargs):
+def ccread(source, *args, **kwargs):
```
langner: I'd prefer not to use args and kwargs at all, since there's no clear need (we know the max number of arguments, I think). But we can deal with that at a later time.

PR #619, opened 2018-09-02
Alpha & beta electrons

jaimergp:
As discussed in PR #295, this implements `alpha` and `beta` electrons in `cclib.method.electrons`. Tests are included. By the way, writing those tests I realized that `testelectrons` were not being imported in `test_method`. Since those weren't being tested, there was an error in the `getdatafile` calls that I have also fixed with this PR.

I have also tried to add convenience properties to `ccData`, but there is this issue with the newlines. Git still marks some lines as changed, even though I did not trim anything with the editor. Anyway, the code is trivial:

In `cclib.parser.data.ccData`:

```
    @property
    def alpha_electrons(self):
        return Electrons(self).alpha()

    @property
    def beta_electrons(self):
        return Electrons(self).beta()
```

At the end of `tests.method.testelectrons.ElectronsTest.test_alpha_beta`:

```
        # Test ccData properties
        self.assertEqual(data.alpha_electrons, 35)
        self.assertEqual(data.beta_electrons, 34)

```

### DIFF
--- a/cclib/method/electrons.py
+++ b/cclib/method/electrons.py
@@ -19,7 +19,7 @@ class Electrons(Method):
 
     def __init__(self, data, progress=None, loglevel=logging.INFO, logname="Log"):
 
-        self.required_attrs = ('atomnos','charge','coreelectrons')
+        self.required_attrs = ('atomnos','charge','coreelectrons','homos')
 
         super(Electrons, self).__init__(data, progress, loglevel, logname)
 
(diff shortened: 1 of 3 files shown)


cclib/method/electrons.py
```diff
@@ -31,6 +31,20 @@ def __repr__(self):
         """Returns a representation of the object."""
         return "Electrons"
 
+    def alpha(self):
+        """Number of alpha electrons"""
+        homos = self.data.homos
+        if len(homos) == 1:
```
langner: I don't think you need this condition, since it's returning the same thing as below.

cclib/method/electrons.py
```diff
@@ -31,6 +31,20 @@ def __repr__(self):
...
+            return homos[0] + 1
+        return homos[0] + 1
+
+    def beta(self):
+        """Number of beta electrons"""
+        homos = self.data.homos
+        if len(homos) == 1:
```
langner: Since there's a maximum of two elements, is this whole thing equivalent to return self.data.homos[-1] + 1?

cclib/method/electrons.py
jaimergp: Doh ^^' Painfully true.

cclib/method/electrons.py
jaimergp: Oh, that's right. Thanks for the suggestion!

Issue #629, opened 2018-10-04
"ValueError: invalid literal for float()" for ORCA 4.0.1.2 logfile


dudektria:
Hi everyone,

I may have found a bug in the ORCA parser:

    $ ccget scfenergies atomcoords 4.out
    Attempting to read 4.out
    Traceback (most recent call last):
      File "/usr/local/bin/ccget", line 11, in <module>
        load_entry_point('cclib==1.6', 'console_scripts', 'ccget')()
      File "/usr/local/lib/python2.7/dist-packages/cclib/scripts/ccget.py", line 165, in ccget
        data = ccread(filename, **kwargs)
      File "/usr/local/lib/python2.7/dist-packages/cclib/io/ccio.py", line 175, in ccread
        return log.parse()
      File "/usr/local/lib/python2.7/dist-packages/cclib/parser/logfileparser.py", line 318, in parse
        self.extract(inputfile, line)
      File "/usr/local/lib/python2.7/dist-packages/cclib/parser/orcaparser.py", line 628, in extract
        temp.append(float(vals[k:k+10]))
    ValueError: invalid literal for float(): 1 74.46258

The only place in the output file where the string "1 74.46258" appears is:

    ------------------
    MOLECULAR ORBITALS
    ------------------
                          0         1         2         3         4         5
                     -101.65359 -101.64676 -101.64160 -19.22880 -19.19334 -10.29501
                       2.00000   2.00000   2.00000   2.00000   2.00000   2.00000
                      --------  --------  --------  --------  --------  --------
      0Pd  1s         0.000024 -0.000025 -0.000024 -0.000010  0.000004 -0.000016
    (...)
      4C   2dx2y2    -0.005845  0.030497  0.003297  0.012404 -0.015548 -0.039709
      4C   2dxy      -0.002457 -0.027637 -0.010772 -0.024476  0.017044  0.039060
      4C   6s        29.895635100.086670-23.048342-79.143850-155.230721 74.462587
      4C   4pz        9.010492 14.676334  0.370009 -1.961820-11.737009  8.031800
    (...)

(Last column, second line from bottom to top. I know this is a very strange coefficient, but it's one ORCA actually generated.)

langner:
Thanks for the report, Felipe. Looks like an easy fix. Would you be comfortable with sharing the logfile causing this problem, so that we can use it as a regression?

dudektria:
Sorry for the late reply. The particular logfile above I would rather not share (right now), sorry.

[...]

Issue #628, opened 2018-09-07
Rethink output of methods


langner:
So we "export" some method output to ccData objects as kinda-attributes via properties. I'm leaning towards reversing that trend for 2.x. I think #605 shows an example of where problems can occur.

I also think there's a conceptual divide between ccData and method objects - the first contains just parsed data, the second computed stuff. It's probably better to keep it that way.

ATenderholt:
Fine with me. Is the divide a strict rule so that only parsed attributes are in `ccData` and anything calculated from those attributes, even if trivial (e.g. bond lengths), should be a method object? Methods were originally for non-trivial calculations like the various population analysis.

berquist:
> I'm leaning towards a strict separation, just because it's less confusing and simpler for developing.

+1. I would rather not have methods on one object instance change the state of another object of a different class; each class should only mutate instances of itself.

langner:
Need to revisit if something needs to happen here, but glancing at LPA seems to say we have separate objects for methods.

PR #630, opened 2018-10-10
[ORCA] Fix for mocoeffs that are glued together (cclib/cclib#629)

langner:
This is a fix for #628, along with a regression test.

Needs to be submitted after cclib/cclib-data#92, should fail in Travis until that happens.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -581,17 +581,40 @@ class ORCA(logfileparser.Logfile):
                     broken = line.split()
                     self.aooverlaps[j, i:i+size] = list(map(float, broken[1:size+1]))
 
-        # Molecular orbital coefficients.
-        # This is also where atombasis is parsed.
+        # Molecular orbital coefficients are parsed here, but also related things
+        #like atombasis and aonames if possible.
+        #
+        # Normally the output is easy to parse like this:
+        # ------------------
+        # MOLECULAR ORBITALS
+        # ------------------
+        #                       0         1         2         3         4         5   
+        #                  -19.28527 -19.26828 -19.26356 -19.25801 -19.25765 -19.21471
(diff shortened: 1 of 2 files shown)


ATenderholt:
What about a regex?  Something like `(-{0,1}\d+\.\d{6})` should work, assuming there is always 6 decimal places. Consume the string until no more matches are available.

langner:
> What about a regex?  Something like `(-{0,1}\d+\.\d{6})` should work, assuming there is always 6 decimal places. Consume the string until no more matches are available.

Yes, that'll work. Done, and it's so short... genius!

ATenderholt:
Looks like there's a problem with the conversion from regex `findall` results and converting to float: https://travis-ci.org/cclib/cclib/jobs/440163390#L682

Also, looks like there's an unrelated problem with the build for 2.7 (https://travis-ci.org/cclib/cclib/jobs/440163389#L583). Is `mock` part of `unittest` in 2.x?

langner:
Thanks for pointing these out. I kinda assumed it would be OK because it was failing already...

> Looks like there's a problem with the conversion from regex `findall` results and converting to float: https://travis-ci.org/cclib/cclib/jobs/440163390#L682

Fixed by using a comprehension.

> Also, looks like there's an unrelated problem with the build for 2.7 (https://travis-ci.org/cclib/cclib/jobs/440163389#L583). Is `mock` part of `unittest` in 2.x?

I attempted to fix these thing in #631. So that should clear Travis, after which this PR should also be clean when I rebase it.

langner:
... and cclib/cclib-data#92 also needs to be merged in first.

berquist:
```
  File "/home/travis/build/cclib/cclib/cclib/parser/orcaparser.py", line 651, in extract
    mocoeffs[spin][i:i+len(coeffs), j] = map(float, coeffs)
TypeError: float() argument must be a string or a number, not 'map'
```

Issue #518, opened 2018-05-28
Sunsetting Python2


jevandezande:
I just wrote some code, and got bitten by the fact that it wasn't supported by python2 (along the lines of `a, *b = [1, 2, 3]`). The advanced unpacking features in python3 are incredibly useful for code like this, but can't be used due to the current need to maintain python2 support. Would it be possible to get rid of python2 support soon?

I'd love to force everyone to use python3.6 (released December 2016), but I realize that may be a bit overzealous. However, forcing at least python3.5 (released September 2015) would allow [additional unpacking generalizations](https://docs.python.org/3.5/whatsnew/3.5.html#whatsnew-pep-448) (e.g. `a, *b, c = [1, 2, 3, 4]`). Wikipedia has a [list of release dates](https://en.wikipedia.org/wiki/History_of_Python#Version_release_dates) for those interested.

Thoughts?

ghutchis:
My $0.02 - I think it's an okay idea, but cclib should really set out a specific timeframe (e.g., end of 2018) and substantially change the version number (e.g., cclib-2.x will only work with Py3).

While there's a strong change towards Python3 in the entire science "stack," the deadline is usually more like 2019-2020:
http://python3statement.org

langner:
Hopefully GSOC will accelerate our work towards 2.x, but I wouldn't be surprised if it went into 2019. I don't think we should drop py2 in any 1.x version.

What we could do, and which might motivate us further to finish 2.x, is to start the 2.x branch right away, and at some point only add bugfixes to the 1.x branch. But I'd worry that would stall new features that are backwards compatible.

langner:
(I am supportive of being forward looking and dropping py2 in the 2.x branch)

langner:
Python2 is officially not supported anymore, so we'll want to accelerate this process unless we formulate a plan for a version 2.x this year. It's probably OK to go ahead with the drop at this point.

langner:
As per dev discussion, we will be dropping Py2 for v1.7 - separately from QCElemental and other changes that could be related.

PR #631, opened 2018-10-12
Get travis passing again in Python2

langner:
I tried to fix all of the things not working in Python2 here:
- mock not being in the standard lib
- getting division to work in test_moments that was written for Python3 (and skipping some cases I can't figure out)
- skipping some tests that rely on periodictable which is not available in Python2

... yet another reason to move off of Python2 (#518)


langner:
I still get some errors in test_io locally due to some weirdness in pybel for Python3, but I think that's specific to Debian at this time since Travis looks fine.

berquist:
The mock stuff definitely existed at one point; not sure what happened to it.

PR #634, opened 2018-10-14
Molcas: fix for test suite job stevenv/001 which was tripping the parser

langner:
The parser was crashing, I think due to there being multiple MO output blocks. This just skips that code, which is probably the right thing to do until we think about support for multiconfiguration jobs.

This depends on cclib/cclib-data#93


cclib/parser/molcasparser.py
```diff
@@ -636,10 +675,19 @@ def extract(self, inputfile, line):
         #         60 H20   1s      0.1835
         #  --
         if '++    Molecular orbitals:' in line:
+
             self.skip_lines(inputfile, ['d', 'b'])
             line = next(inputfile)
-            if 'Natural orbitals' not in line:
-                self.skip_lines(inputfile, ['b', 'symm'])
+
+            # We don't currently support parinsg natural orbitals or active space orbitals.
```
berquist: `parinsg -> parsing`

cclib/parser/molcasparser.py
```diff
@@ -98,7 +110,7 @@ def extract(self, inputfile, line):
             self.atomcoords.append(atomcoords)
 
             if self.atomnos == []:
-                self.atomnos = [utils.PeriodicTable().number[atomelement] for atomelement in atomelements]
+                self.atomnos = [utils.PeriodicTable().number[ae.title()] for ae in atomelements]
```
berquist: The Logfile also has a PeriodicTable stored:
```suggestion
                self.atomnos = [self.table.number[ae.title()] for ae in atomelements]
```

cclib/parser/molcasparser.py
```diff
@@ -40,17 +40,29 @@ def normalisesym(self, label):
...
         # atom label in the "Molecular structure info" block.
-        self.re_atomelement = re.compile('([a-zA-Z]+)\d+')
+        self.re_atomelement = re.compile('([a-zA-Z]+)\d?')
 
         # Compile the dashes-and-or-spaces-only regex.
         self.re_dashes_and_spaces = re.compile('^[\s-]+$')
 
+        # Molcas can do multiple calculations in one job, and each one
+        # starts from the gateway module. Onle parse the first.
+        # TODO: It would be best to parse each calcualtion as a separate
```
berquist: `calcualtion -> calculation`

langner: Fixed in a0671265b22f7a1e8afb76d929405398f3fc1499

cclib/parser/molcasparser.py
langner: Fixed in a0671265b22f7a1e8afb76d929405398f3fc1499

PR #636, opened 2018-10-20
Molcas core

langner:
This is a revival of #593. I only addressed the hanging comments there, so we can merge it in.


cclib/parser/logfileparser.py
```diff
@@ -426,6 +426,21 @@ def append_attribute(self, name, value):
             self.set_attribute(name, [])
         getattr(self, name).append(value)
 
+    def _assign_coreelectrons_to_element(self, element, ncore, divide_by_count=False):
```
ATenderholt: Is there a clearer name for `divide_by_count`? Without looking how this function is used elsewhere, I think it's supposed to indicate whether `ncore` is a per-atom value instead of total # of core electrons across all atoms of given element. Perhaps `ncore_is_total_count` or `ncore_is_per_atom`?

cclib/parser/molcasparser.py
```diff
@@ -40,13 +40,14 @@ def normalisesym(self, label):
     def before_parsing(self):
         # Compile the regex for extracting the element symbol from the
         # atom label in the "Molecular structure info" block.
-        self.re_atomelement = re.compile('([a-zA-Z]+)\d+')
+        self.re_atomelement = re.compile('([a-zA-Z]+)|d+')
```
ATenderholt: Was this supposed to be changed? Or are you missing a `\` for the a digit class?

cclib/parser/logfileparser.py
berquist: That was the original intent, yes.

langner: Improved this in a2b2e6458f5cf26de1743ab8077ff6fa9476904d

langner: Improved this in a2b2e6458f5cf26de1743ab8077ff6fa9476904d

langner: Improved this in a2b2e6458f5cf26de1743ab8077ff6fa9476904d

langner: Improved in 3a3f6edbb4dcda098f464408f63d7496006d4b6f

langner: Improved in 3a3f6edbb4dcda098f464408f63d7496006d4b6f

langner:
...

cclib/parser/molcasparser.py
langner: So it should really be, I think `([a-zA-Z]{1,3})\d*` which I check and does work. But, I really don't think we need regexes for this. Rewrote this Pythonically in e0c65498759d0f5955e19904cd6202ae1b6c4302.

langner:
> Only a few minor things.

Rebased yesterday when GH was having problems. Should be good to go.

PR #637, opened 2018-10-23
Minor psi4 parser fixes

alesgenova:
I tried running the parser against a psi4 1.2 output and ran into a couple of problems.

I believe the origin changed between 1.1 and 1.2, while the orbital coefficients were already broken in 1.1


alesgenova:
Sure I will. Where should I send it to?

I also noticed that the normal modes parser needs to be updated, I'll try to fix it quickly in the next commit

langner:
> Sure I will. Where should I send it to?

You can attach it here and we'll take care of the rest. Or, if you're up for it, you can submit it to the regression repo: https://github.com/cclib/cclib-data. If the latter, just plop the file in the right directory, and add the path to `regressionfiles.txt`.

alesgenova:
Added the output file to the `cclib-data` repo!
https://github.com/cclib/cclib-data/pull/95

cclib/parser/psi4parser.py
```diff
@@ -718,9 +729,13 @@ def get_normalization_factor(exp, lx, ly, lz):
...
         # Properties will be evaluated at   0.000000,   0.000000,   0.000000 Bohr
         #
+        # OR
+        #
+        # Properties will be evaluated at   0.000000,   0.000000,   0.000000 [a0]
+        #
         if "Properties will be evaluated at" in line.strip():
             self.origin = numpy.array([float(x.strip(',')) for x in line.split()[-4:-1]])
-            assert line.split()[-1] == "Bohr"
+            assert line.split()[-1] in ["Bohr", "[a0]"]
```
langner: So if the units are angstroms here, do you need to do the conversion on the line below?

Is the regression file you added in cclib/cclib-data#95 represents this case, you can consider adding an appropriate regression test in `regression.py` (in the main repo) and check that the parsed value is in the right units. The tricky thing is to name the regression test function correctly, and the file has gotten really big, so I can take care of that if needed.

alesgenova: I assumed that `a0` would still be Bohr, while I would expect something like `ang` to be Angstrom

https://en.wikipedia.org/wiki/Bohr_radius

alesgenova:
Thanks for the speedy review! :)

Issue #309, opened 2016-08-17
Clean up reading input


langner:
As a follow up to #306 it would be good to put a little work into dealing with various inputs. Ideally, we'd like something with the following properties:
* given a filename, opens it and reads as usual (this is like 99% of use cases)
* given an open file, uses that
* given a list of filenames or open files, uses then consecutively
* given an URL, open it and read from there
* given any other stream-like object, use that
* in particular, make sure to handle standard input nicely (Avogadro integration relies on this)
* bonus: a mechanism to read ahead on streams, but not necessarily read in the whole thing

langner:
Will review before v1.5.3.

langner:
This is long overdue, and I really had a hard time fixing the code for #638. Before doing this cleanup (or rather refactoring) it'd be good to make sure test coverage is sufficient to trust the refactoring, including #220.

The bevarios described at the top here should probably be implemented in one class, to keep IO central. `FileWrapper` is perhaps a good starting point.

berquist:
Some of this, at least refactoring, was done in #1197.

PR #638, opened 2018-10-24
Quick fixes to get multiple files working with wildcard for Turbomole regressions

langner:
Specifically this is for getting the file added in cclib/cclib-data#94 parsed correctly (added a small test func here).

Obviously we need to refactor our code in `ccio.py` and `regression.py`, really badly. Also parts of `logfileparser.py` are related to this mess. But this is just about getting things working for Turbomole regressions. Further work will continue in #309, when there's the will :)

### DIFF
--- a/cclib/parser/logfileparser.py
+++ b/cclib/parser/logfileparser.py
@@ -223,7 +223,7 @@ class Logfile(object):
             self.isstream = True
             self.stream = source
         else:
-            raise ValueError, "Unexpected source type."
+            raise ValueError("Unexpected source type.")
 
         # Set up the logger.
         # Note that calling logging.getLogger() with one name always returns the same instance.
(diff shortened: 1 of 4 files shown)


cclib/parser/logfileparser.py
```diff
@@ -207,18 +207,23 @@ def __init__(self, source, loglevel=logging.INFO, logname="Log",
...
+            self.filename = source
+            self.isstream = False
+            self.isfileinput = True
         elif hasattr(source, "read"):
             self.filename = "stream %s" % str(type(source))
             self.isstream = True
             self.stream = source
         else:
-            raise ValueError
+            raise ValueError, "Unexpected source type."
```
berquist: This is causing the Travis failure.

langner: Should be good to go now. I started using six, which I think will be useful in many other places in the code as we work towards 2.x

PR #642, opened 2018-10-29
Add unit tests for missing required attributes in population methods

langner:
This is a paraphrase and extension of #480.

### DIFF
--- a/cclib/method/population.py
+++ b/cclib/method/population.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2017, the cclib development team
+# Copyright (c) 2018, the cclib development team
 #
 # This file is part of cclib (http://cclib.github.io) and is distributed under
 # the terms of the BSD 3-Clause License.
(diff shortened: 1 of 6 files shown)


cclib/method/population.py
```diff
@@ -32,6 +34,13 @@ def __repr__(self):
         """Return a representation of the object."""
         return "Population"
 
+    def _check_required_attributes(self):
+        super(Population, self)._check_required_attributes()
+        
+        if not any(hasattr(self.data, a) for a in self.overlap_attributes):
+            raise MissingAttributeError(
+                    'Need overlap matrix (aooverlaps or fooverlaps attribute) for Population methods')
```
ATenderholt: The overlap matrix isn't required for CSPA.

cclib/method/population.py
```diff
@@ -11,17 +11,19 @@
...
 
-from cclib.method.calculationmethod import Method
+from cclib.method.calculationmethod import Method, MissingAttributeError
 
 
 class Population(Method):
     """An abstract base class for population-type methods."""
 
+    required_attrs = ('homos', 'mocoeffs', 'nbasis')
+    overlap_attributes = ("aooverlaps", "fooverlaps")
```
ATenderholt: Any reason you switched from single quotes to double quotes? (No real preference, just was surprised.)

cclib/method/population.py
langner: Indeed! Done in 83c0738ef18535f16e807a1f5235329e54cf6a3b.

cclib/method/population.py
langner: Done in 83c0738ef18535f16e807a1f5235329e54cf6a3b.

PR #644, opened 2018-11-01
CJSON bond atom index should start from 0

alesgenova:
In all the `cjson` examples I have seen so far, the atom index in the bonds connection array start from 0

### DIFF
--- a/cclib/io/cjsonwriter.py
+++ b/cclib/io/cjsonwriter.py
@@ -133,8 +133,8 @@ class CJSON(filewriter.Writer):
             cjson_dict['bonds']['connections'] = dict()
             cjson_dict['bonds']['connections']['index'] = []
             for bond in self.bond_connectivities:
-                cjson_dict['bonds']['connections']['index'].append(bond[0] + 1)
-                cjson_dict['bonds']['connections']['index'].append(bond[1] + 1)
+                cjson_dict['bonds']['connections']['index'].append(bond[0])
+                cjson_dict['bonds']['connections']['index'].append(bond[1])
             cjson_dict['bonds']['order'] = [bond[2] for bond in self.bond_connectivities]
 
         if _has_openbabel:
--- a/test/io/testcjsonwriter.py
+++ b/test/io/testcjsonwriter.py
@@ -49,6 +49,13 @@ class CJSONTest(unittest.TestCase):
             sqrt(sum(data.moments[1] ** 2))
         )
 
+        # Ensure the bond connectivity index starts from 0
+        bonds = json_data.get('bonds', None)
+        self.assertIsNotNone(bonds)
+        indices = bonds['connections']['index']
+        self.assertEqual(min(indices), 0)
+        self.assertTrue(max(indices) < number_of_atoms)
+
     def test_zero_dipole_moment(self):
         """Does the CJSON writer handle zero dipole moment correctly?"""
         fpath = os.path.join(__datadir__, "data/GAMESS/basicGAMESS-US2017/C_bigbasis.out")


cclib/io/cjsonwriter.py
```diff
@@ -131,8 +131,8 @@ def as_dict(self):
             cjson_dict['bonds']['connections'] = dict()
             cjson_dict['bonds']['connections']['index'] = []
             for bond in self.bond_connectivities:
-                cjson_dict['bonds']['connections']['index'].append(bond[0] + 1)
-                cjson_dict['bonds']['connections']['index'].append(bond[1] + 1)
+                cjson_dict['bonds']['connections']['index'].append(bond[0])
+                cjson_dict['bonds']['connections']['index'].append(bond[1])
```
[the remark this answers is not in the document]
langner: Since I think `len(bond) == 2`, this can just be `.extend(bond)`.

langner:
> In all the `cjson` examples I have seen so far, the atom index in the bonds connection array start from 0

Thanks, I think you're right. I'll give @berquist or @ATenderholt a day or two to respond in case they have a different opinion.

cclib/io/cjsonwriter.py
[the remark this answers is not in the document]
alesgenova: By looking at the code below, it looks like the length of bond is 3, with `bond[2]` being the bond order

berquist:
I will dig up some examples and check.

berquist:
Here is the CJSON writer code from Avogadro 2:
``` c++
  // Bonds are optional, but if present should be loaded.
  json bonds = jsonRoot["bonds"];
  if (bonds.is_object() && isNumericArray(bonds["connections"]["index"])) {
    json connections = bonds["connections"]["index"];
    for (unsigned int i = 0; i < connections.size() / 2; ++i) {
      molecule.addBond(static_cast<Index>(connections[2 * i]),
                       static_cast<Index>(connections[2 * i + 1]), 1);
    }
    json order = bonds["order"];
    if (isNumericArray(order)) {
      for (unsigned int i = 0; i < molecule.bondCount() && i < order.size();
           ++i) {
        molecule.bond(i).setOrder(static_cast<int>(order[i]));
      }
    }
  }
```
so yes, connectivities start from 0.

For the bond order storage, it is calculated [here](https://github.com/cclib/cclib/blob/5802269656379fcb04977a1824c8e11d88568014/cclib/io/filewriter.py#L109). It just seemed like a simple thing to do at the time, so if there is a better way to do it, please do change it.

langner:
I think this is OK to merge in (extending `[:2]` is OK, but IMO not necessary). But, would you be able to add a unit test check for this index starting from zero?

test/io/testcjsonwriter.py
```diff
@@ -43,5 +43,11 @@ def test_cjson_generation(self):
         number_of_atoms = json_data['properties']['number of atoms']
         self.assertEqual(number_of_atoms, data.natom)
 
+        # Ensure the bond connectivity index starts from 0
+        bonds = json_data.get('bonds', None)
+        self.assertIsNotNone(bonds)
```
berquist: This doesn't check that the index starts at zero. Consider what some example CJSON should look like:
``` json
  "bonds": {
    "connections": {
      "index": [
        0,
        1,
        0,
        2,
        0,
        3,
```

alesgenova: Generally, I don't think the atom with index 0 is guaranteed to be involved in any bond.
That's why my all my test does is to make sure that the larges index used in bond connectivity doesn't exceed the number of atoms.

berquist: While that is true, in the case of our test molecules, 0 will exist in the list, and should be checked, otherwise we aren't actually testing the change in the code. We need to know that it _can_ occur for some example.

alesgenova: @berquist 
Ok, I also added a check to ensure that 0 is indeed part of the connections.

langner:
@berquist is this ready to go?

PR #647, opened 2018-11-08
Switch from unittest to pytest

berquist:
Closes https://github.com/cclib/cclib/issues/455.

There are still small annoyances, like not displaying the individual tests from `test_all`, and `test_regressions` showing as one test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,6 @@ build
 dist
 *.pyc
 .cache
+.coverage
+.pytest_cache
 cclib.egg-info
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ __pycache__
 MANIFEST
 build
 dist
+htmlcov
 *.pyc
 .cache
 .coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_script:
   - install -dm755 "${THEME_DIR}"
 
 script:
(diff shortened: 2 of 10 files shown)


langner:
Two questions:
1. Could you perhaps create separate issues for the bits you think should still be done?
2. Would it be feasible to publish the HTML output for coverage somewhere like cclib.github.io/coverage?

berquist:
> Would it be feasible to publish the HTML output for coverage somewhere like cclib.github.io/coverage?

Yes, it just needs to get copied into the right place, and a link should be added in the docs. Should I do it for this PR?

langner:
> > Would it be feasible to publish the HTML output for coverage somewhere like cclib.github.io/coverage?
> 
> Yes, it just needs to get copied into the right place, and a link should be added in the docs. Should I do it for this PR?

Sure, if it's not too much trouble.

dgasmith:
Huge fan of `pytest`, things like parametrize and fixture really cut down on the amount of required code.

langner:
> > > Would it be feasible to publish the HTML output for coverage somewhere like cclib.github.io/coverage?
> > 
> > 
> > Yes, it just needs to get copied into the right place, and a link should be added in the docs. Should I do it for this PR?
> 
> Sure, if it's not too much trouble.

I guess the doc generation process/script will take care of this, so I suppose we'll need to udpate it once this is submitted.

PR #656, opened 2018-11-20
Fixes for Gaussian98 G3 job from NIST CCCBDB

langner:
These are several minor quick fixes to get this file parsing, with a basic regression test.

The major issue is that G3 is actually a sequence of jobs. So until we have better support for that I just wanted to get the file parsing with useful data. Deleting only some attributes (`mpenergies` here) from previous steps is maybe not the best thing to do, but it seemed like a better option than deleting all attributes.


langner:
The data for this is in cclib/cclib-data#99.

PR #658, opened 2018-11-21
Log error and break on StopIteration in extract

langner:


langner:
Logging, breaking and keeping what was already parsed seems better than just raising the exception

langner:
The corresponding regression logfile is in cclib/cclib-data#101

PR #662, opened 2018-12-01
Test failures should always return a non-zero error code

### LINKED ISSUE #661: Travis not catching errors during tests redux
I discovered while testing https://github.com/cclib/cclib/pull/659 locally that the pytest setup is broken, and Travis reports a pass. Even rebasing the PR on top of latest master doesn't fix things. This is because `run_pytest.sh` is calling a slightly different function than `python -m test.test_data`.

berquist:
https://github.com/cclib/cclib/pull/647 removed the inability for failing regression tests to not return a zero error code, but didn't do the same for the data tests.

Rather than just call the test driver with `--status`, or have `status=True` be the default, I think it should be removed completely, like for regression tests. It's dangerous if you forget it, and I don't see why you'd ever want the option of not returning an error code, so it's safer to just remove the logic completely.

The other tests (bridge, IO, etc.) shouldn't need modification, since they're plain `unittest` tests that return 1 on failures.

Closes #661.


langner:
I don't quite remember why `status` was added, but I agree it doesn't seem necessary right now. Hopefully this makes Travis more consistent.

PR #666, opened 2018-12-07
Add oscillator strengths (etoscs) from Dalton response calculation

peter-reinholdt:
Adds parsing of etoscs from Dalton linear response calculation.


peter-reinholdt:
Yes, the test should be enabled now

berquist:
Great!

Issue #667, opened 2018-12-07
Error in parsing versions information of newer Dalton releases


peter-reinholdt:
When the `cclib/parser/daltonparser.py` is trying to  match the Dalton version
```python
        if line[4:30] == "This is output from DALTON":
            rs = r"([0-9]{4})\D\s?\w*\s?([0-9]{1})"
            match = re.search(rs, line)
            package_version = ".".join(match.groups())
            self.metadata["package_version"] = package_version
```
it is unable to match version info from new/development version of Dalton. 

For example, the following while the following version is not matched
```
    This is output from DALTON release Dalton2019.alpha (2018)
```
which leads to
```
     86             match = re.search(rs, line)
---> 87             package_version = ".".join(match.groups())
     88             self.metadata["package_version"] = package_version
     89         # Don't add revision information to the main package version for now.

AttributeError: 'NoneType' object has no attribute 'groups'
```
since the `re.search` returns `None`

berquist:
Correct, this is bad. Here are all the unique version strings I grepped from the current test:
```
    This is output from DALTON 2013.2
    This is output from DALTON 2013.4
    This is output from DALTON 2014.0
    This is output from DALTON 2014.alpha
    This is output from DALTON 2015.0
    This is output from DALTON 2016.alpha
               This is output from DALTON (Release 1.1, September 2000)
    This is output from DALTON Release 2011 (DEVELOPMENT VERSION)
    This is output from DALTON Release 2011 (Rev. 0, Dec. 2010)
    This is output from DALTON Release 2011 (Rev. 0, Mar. 2011)
    This is output from DALTON (Release 2.0 rev. 0, Mar. 2005)
    This is output from DALTON (Release Dalton2013 patch 0)
    This is output from DALTON release Dalton2017.alpha (2016)
    This is output from DALTON release Dalton2017.alpha (2017)
    This is output from DALTON release Dalton2018.0 (2018)
    This is output from DALTON release Dalton2018.0-rc (2018)
    This is output from DALTON release Dalton2018.alpha (2018)
    This is output from DALTON release Dalton2019.alpha (2018)
```
I can rewrite the regex, but what detail do you want to see in the version field (if it matters to you)?

berquist:
Here is a start:
```
from DALTON \(?(?:Release|release)?\s?(?:Dalton)?(\d+\.?[\w\d\-]*)[\s,]\(?
```
[...]

PR #668, opened 2018-12-08
DALTON: fix version parsing of never development versions

berquist:
The file for https://github.com/cclib/cclib/issues/667 and https://github.com/cclib/cclib-data/pull/103.

This needs to go in **before** https://github.com/cclib/cclib-data/pull/103.

### DIFF
--- a/cclib/parser/daltonparser.py
+++ b/cclib/parser/daltonparser.py
@@ -81,6 +81,27 @@ class DALTON(logfileparser.Logfile):
 
         # Extract the version number and optionally the Git revision
         # number.
+        #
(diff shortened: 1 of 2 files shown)


cclib/parser/daltonparser.py
```diff
@@ -82,10 +82,12 @@ def extract(self, inputfile, line):
         # Extract the version number and optionally the Git revision
         # number.
         if line[4:30] == "This is output from DALTON":
```
ATenderholt: Can you add some examples that the regex is trying to match?

berquist: Let me know if the comment block is enough.

PR #659, opened 2018-11-21
Natural spin orbitals

shivupa:
Natural spin orbitals can now be parsed and a regression test has been added.

PR #671, opened 2018-12-23
Molcas Parser fix

shivupa:
Small molcas parser fix. Mocoeffs were not parsed in regression tests because
of an extra comment line and extra blank line. This is fixed.

This implements the change discussed in #659

### DIFF
--- a/test/regression.py
+++ b/test/regression.py
@@ -627,7 +627,7 @@ def testMolcas_Molcas18_test_standard_001_out(logfile):
     assert logfile.data.nbasis == 30
     assert logfile.data.nmo == 30
 
-def testMolcas_Molcas18_test_stadard_003_out(logfile):
+def testMolcas_Molcas18_test_standard_003_out(logfile):
     """This logfile has extra charged monopoles (not part of the molecule)."""
     assert logfile.data.charge == 0
 
--- a/cclib/parser/molcasparser.py
+++ b/cclib/parser/molcasparser.py
@@ -685,7 +685,10 @@ class Molcas(logfileparser.Logfile):
                 self.skip_line(inputfile, 'b')
 
                 # Symmetry is not currently supported, so this line can have one form.
-                line = next(inputfile)
+                while 'Molecular orbitals for symmetry species 1: a' not in line.strip():
+                    line = next(inputfile)
+
+                # Symmetry is not currently supported, so this line can have one form.
                 if line.strip() != 'Molecular orbitals for symmetry species 1: a':
                     return
                 


langner:
Thanks for this. Is there a change to `regression.py` that would be appropriate here?

berquist:
Maybe take the lines that are in `old_unittests` for the SP and SPun tests:
```python
    "Molcas/Molcas8.0/dvb_sp.out":      MolcasSPTest,
    "Molcas/Molcas8.0/C_bigbasis.out":  MolcasBigBasisTest,
```
and turn them into regressions that assert they have `mocoeffs`?

shivupa:
Good idea thanks. Done!

test/regression.py
```diff
@@ -614,6 +614,16 @@ def testJaguar_Jaguar8_3_stopiter_jaguar_hf_out(logfile):
 
 # Molcas #
 
+def testMolcas_Molcas8_0_dvb_sp_out(MolcasSPTest):
+ """Test to confirm that MOs are parsed."""
+ def test_for_mo_parsing(self):
```
langner: This defines a nested function, but doesn't actually do anything. Just do the assert in `testMolcas*`, like in other regression test functions.

test/regression.py
```diff
@@ -614,6 +614,16 @@ def testJaguar_Jaguar8_3_stopiter_jaguar_hf_out(logfile):
...
 # Molcas #
 
+def testMolcas_Molcas8_0_dvb_sp_out(MolcasSPTest):
+ """Test to confirm that MOs are parsed."""
+ def test_for_mo_parsing(self):
+     assert hasattr(self.data, "mocoeffs")
+
+def testMolcas_Molcas8_0_dvb_sp_un_out(GenericSPunTest):
+ """Test to confirm that MOs are parsed."""
+ def test_for_mo_parsing(self):
```
langner: Ditto.

shivupa: Fixed

test/regression.py
```diff
@@ -627,7 +635,7 @@ def testMolcas_Molcas18_test_standard_001_out(logfile):
     assert logfile.data.nbasis == 30
     assert logfile.data.nmo == 30
 
-def testMolcas_Molcas18_test_stadard_003_out(logfile):
+def testMolcas_Molcas18_test_standard_003_out(logfile):
```
langner: Thanks for catching this!

langner:
> Maybe take the lines that are in `old_unittests` for the SP and SPun tests:
> 
> ```python
>     "Molcas/Molcas8.0/dvb_sp.out":      MolcasSPTest,
>     "Molcas/Molcas8.0/C_bigbasis.out":  MolcasBigBasisTest,
> ```
> 
> and turn them into regressions that assert they have `mocoeffs`?

I believe `mocoeffs` are already checked, at least in `MolcasSPTest`, so there should be no need for an explicit regression function here. Was this not working before the fix?

shivupa:
You're right, I reverted the changes in `regression.py`. The checks  in `MolcasSPTest` and `GenericSPunTest` were silently failing, but after the change they pass and actually test for `mocoeffs`.  This PR doesn't touch or address `Molcas/Molcas8.0/C_bigbasis.out":  MolcasBigBasisTest`

langner:
Thanks for this!

Issue #677, opened 2018-12-31
DALTON: dominant excitation contributions not printed for Hartree-Fock


berquist:
The section used for parsing oscillator strengths:
```
>                             PBHT MO Overlap Diagnostic
>                             --------------------------
>
>   Reference: MJG Peach, P Benfield, T Helgaker, and DJ Tozer.
>              J Chem Phys 128, 044118 (2008)
>
>
>   The dominant contributions:
>
>       I    A    K_IA      K_AI   <|I|*|A|> <I^2*A^2>    Weight   Contrib
>
>      27   28  0.564533  0.007792  0.790146  0.644560  0.309960  0.244913
>      58   59 -0.422308  0.013798  0.784974  0.651925  0.190188  0.149293
>
> @ Overlap diagnostic LAMBDA =    0.7882
>
```
is only printed for DFT, not Hartree-Fock-based wavefunctions.

langner:
What is the action item here?

berquist:
I just looked at the source code and there is no way to get this outside of DFT. I am trying to find the equivalent part of the output that exists in Hartree-Fock output, even if it means turning up the print level. If I can't find anything, it might be a simple calculation to implement. Then I'll just stick the output in regressions.

berquist:
I remember the problem, I have a patch that works for 2015 and up, but some output is missing from 2013 (`Eigenvector for state no.`) and was working on getting a copy of 2013 compiled.

langner:
An upstream patch to Dalton?

berquist:
Nope, there is a configurable print level that works, so I'll update the unit tests.

On Mon, Jan 14, 2019 at 12:20 AM Karol M. Langner <notifications@github.com<mailto:notifications@github.com>> wrote:

An upstream patch to Dalton?

langner:
One thing I do not understand: why can't we parse the DFT-specific version?

langner:
Eric, I guess we're done here. But, before closing out, could you explain (or re-explain if I missed it) why we can't parse the DFT-specific version?

berquist:
Sure. The coefficient it prints in the DFT section for the excitation is not the one that other programs print; it isn't normalized properly.

I ran some RPA calculations with symmetry disabled to remove any inconsistencies due to DFT, and checked the first excitation transition coefficients that belonged to configurations common among all calculations:
```python
#!/usr/bin/env python

from functools import reduce
from pathlib import Path
from pprint import pprint
from cclib.io import ccread

[...]

PR #680, opened 2019-01-21
DALTON: fix parsing electronic transition coefficients for non-DFT wavefunctions

berquist:
Closes https://github.com/cclib/cclib/issues/677. With this fix, more tests can be enabled.

This PR should go in **before** https://github.com/cclib/cclib-data/pull/106.


cclib/parser/daltonparser.py
```diff
@@ -1148,8 +1151,9 @@ def extract(self, inputfile, line):
```
langner: Consider adding a comment with an example above this block.

test/data/testTD.py
```diff
@@ -24,7 +24,6 @@ class GenericTDTest(unittest.TestCase):
     number = 5
     expected_l_max = 41000
 
-    @skipForParser('DALTON', 'etoscs are not parsed')
```
langner: Was this failing before?

test/regression.py
```diff
@@ -165,6 +165,16 @@ def testADF_ADF2016_fa2_adf_out(logfile):
 
 # DALTON #
 
+def testDALTON_DALTON_2013_dvb_td_normalprint_out(logfile):
+    """This original unit test prints a DFT-specific version of the excitation
```
langner: Could you reiterate why we do not parse the DFT-specific version?

berquist: The field that was parsed (https://github.com/cclib/cclib/pull/680/files#diff-88f10a78215b8cf004128ef198e5762bL1197) is incorrect, and the correct field (the scaled one) that matches what other programs print isn't available in the DFT-specific output.

langner: Could you make this clear in the docstring here, and/or in the parser code? I think these types of things are worth documenting, so we don't trip over them again in the future.

langner:
Redirecting this to the `release-1.6.1` branch as well.

test/regression.py
berquist: Done in https://github.com/cclib/cclib/pull/680/commits/bfd9a8d8bbaa1b34e2af4d57e30174fd7e34d964

langner: Thanks!

Issue #673, opened 2018-12-24
Silently failing old unit tests


shivupa:
I think a change might be necessary in the `old_unitesttests`.

Starting from a clean `master` branch, If you change the `devnull` output of the old unit tests like below:
```python
def make_regression_from_old_unittest(test_class):
    """Return a regression test function from an old unit test logfile."""

    def old_unit_test(logfile):
        test_class.logfile = logfile
        test_class.data = logfile.data
        #devnull = open(os.devnull, 'w')
        devnull = sys.stdout
        return unittest.TextTestRunner(stream=devnull).run(unittest.makeSuite(test_class))

    return old_unit_test
```  
then the output of `python -m test.regression --traceback Molcas` changes from:
```
Are the Molcas files ccopened and parsed correctly?
  /home/shiv/Documents/real/cclib/test/../data/regression/Molcas/Molcas18/test_standard_000.out ... parsed and tested
  /home/shiv/Documents/real/cclib/test/../data/regression/Molcas/Molcas18/test_standard_001.out ... parsed and tested
  /home/shiv/Documents/real/cclib/test/../data/regression/Molcas/Molcas18/test_standard_003.out ... parsed
  /home/shiv/Documents/real/cclib/test/../data/regression/Molcas/Molcas18/test_stevenv_001.out ... parsed and tested
  /home/shiv/Documents/real/cclib/test/../data/regression/Molcas/Molcas18/test_stevenv_desym.out ... parsed and tested
  /home/shiv/Documents/real/cclib/test/../data/regression/Molcas/Molcas8.0/C_bigbasis.out ... parsed and tested
  /home/shiv/Documents/real/cclib/test/../data/regression/Molcas/Molcas8.0/dvb_sp.out ... parsed and tested
  /home/shiv/Documents/real/cclib/test/../data/regression/Molcas/Molcas8.0/dvb_sp_un.out ... parsed and tested

Total: 8   Failed: 0  Errors: 0

WARNING: You are missing 1 regression file(s).
Run regression_download.sh in the ../data directory to update.
Missing files:
Turbomole/Turbomole7.2/dvb_gopt_b3-lyp_Gaussian

WARNING: There are 1 orphaned regression test functions.
Please make sure these function names correspond to regression files:
testMolcas_Molcas18_test_stadard_003_out
```
to:
```
Are the Molcas files ccopened and parsed correctly?
[...]

Issue #523, opened 2018-05-29
Future of testing metadata


berquist:
As mentioned in https://github.com/cclib/cclib/pull/506, testing metadata will become difficult as more keys become added, since little is standardized and nothing is documented yet.

These are the keys currently used for metadata that are tested in some way:

* `basis_set`
* `input_file_contents`
* `input_file_name`
* `methods`
* `package`
* `package_version`

These are untested keys:

* `coords`
* `functional`
* `info`
* `keywords`
* `success`
* `unrestricted`
* `warnings`

All mentions of `metadata` in the test directory:

```
test/regression.py:        assert self.data.metadata["symmetry_full"] == "c1"
test/regression.py:        assert self.data.metadata["symmetry_abelian"] == "c1"
test/regression.py:        assert self.data.metadata["symmetry_full"] == "c2h"
test/regression.py:        assert self.data.metadata["symmetry_abelian"] == "c1"
test/data/testGeoOpt.py:        self.assertTrue(self.data.metadata['success'])
test/data/testSP.py:    @skipForParser('ADF', 'Does not support metadata yet')
test/data/testSP.py:    @skipForParser('GAMESSUK', 'Does not support metadata yet')
test/data/testSP.py:    def testmetadata(self):
test/data/testSP.py:        """Does metadata have expected keys and values?"""
test/data/testSP.py:        self.assertTrue(hasattr(self.data, "metadata"))
test/data/testSP.py:            self.assertIn("basis_set", self.data.metadata)
test/data/testSP.py:            self.assertIn("input_file_name", self.data.metadata)
test/data/testSP.py:            self.assertIn("input_file_contents", self.data.metadata)
test/data/testSP.py:        self.assertIn("methods", self.data.metadata)
test/data/testSP.py:        self.assertIn("package", self.data.metadata)
test/data/testSP.py:        self.assertIn("package_version", self.data.metadata)
```

(I'm not sure how the ones for symmetry made it in to regressions.)

This issue is to collect thoughts about how to structure tests for `metadata`, which will make documenting it easier in the future, and also what additional things should go in `metadata` in the near future. I can poke in https://github.com/cclib/cclib/issues/451, which is directly related.

berquist:
https://github.com/cclib/cclib/pull/599 tried this by making metadata its own calculation type, only for the purposes of testing. What we really need is to be able to test metadata across different calculation types, or having something like #737.

PR #685, opened 2019-01-30
Removing unnecessary asserts

shivupa:
In the regression tests, the asserts testing the `self.data.metadata["symmetry_*"]` cause key errors. 

For example:
```
Error for testsymmetry (__main__.ADFSPTest_nosyms_valence)
   File "/home/shiv/Documents/cclib/test/regression.py", line 1658, in testsymmetry
     assert self.data.metadata["symmetry_full"] == "c1"
KeyError: 'symmetry_full
```

**I removed these because as far as I can tell, neither `metadata["symmetry_full"]` nor `metadata["symmetry_abelian"]` are ever even set.**

This was somewhat mentioned in #523.

This fixes 6 errors in the regression tests (relevant issue: #673):
Before:
```
Total: 418   Failed: 0  Errors: 50
```
After:
```
Total: 418   Failed: 0  Errors: 44
```

### DIFF
--- a/test/regression.py
+++ b/test/regression.py
@@ -746,7 +746,7 @@ def testGaussian_Gaussian09_relaxed_PES_testCO2_log(logfile):
     # The next new point is after a convergence.
     assert optstatus[new_points[3]-1] == ccData.OPT_DONE
     assert all(optstatus[i] == ccData.OPT_UNKNOWN for i in range(new_points[2]+1, new_points[3]-1))
-    
+
     # All subsequent point are both new and converged, since they seem
     # to have converged in a single step.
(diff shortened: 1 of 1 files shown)


langner:
Hmmm... @berquist added something like that, I remember. Eric, did we end up removing it?

In any case, let's do this for the 1.6.2 release. I'd like to start freezing things for 1.6.1. I'll do that in a branch, though, so feel free to merge this in is it's good to go.

berquist:
I remember seeing them in the file and being confused, because the symmetry functionality was never merged. But here it is: https://github.com/cclib/cclib-data/pull/56

langner:
Well, in that case I guess we can merge this in. @shivupa let me know if you have trouble resolving the conflicts.

shivupa:
Thanks and sorry about the delay!

PR #689, opened 2019-02-07
Set default logging level in test_data to ERROR

### LINKED ISSUE #687: Print less debugging by default when running unit tests
Currently we get a lot of logging messages during parsing when running unit tests. Before these were printed at the point where test methods were run, but now for some reason they all show up at the bottom, under the summary, which makes thing hard to read.

AI: add a verbose flag somewhere and make it False by default.

langner:
Fixes #687.

### DIFF
--- a/test/test_data.py
+++ b/test/test_data.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2017, the cclib development team
+# Copyright (c) 2019, the cclib development team
 #
 # This file is part of cclib (http://cclib.github.io) and is distributed under
 # the terms of the BSD 3-Clause License.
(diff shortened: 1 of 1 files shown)


test/test_data.py
```diff
@@ -275,6 +276,11 @@ def test_all(parsers=None, modules=None, terse=False, silent=True, summary=True,
 
 if __name__ == "__main__":
 
+    if "--debug" in sys.argv:
+        logging.getLogger().setLevel(logging.DEBUG)
+    else:
+        logging.getLogger().setLevel(logging.FATAL)
```
ATenderholt: Any reason for `FATAL` instead of `ERROR`? Just to suppress *all* output from the parsers/tests?

Also, went looking for description of `FATAL` in the docs, and was surprised to find that it's not mentioned even though it's identical to `CRITICAL`. See https://www.python.org/dev/peps/pep-0282/#levels.

langner: You're right. Changed to ERROR.

Issue #237, opened 2016-01-26
Use of logging utilities


berquist:
Should we more liberal with the use of the logger? It seems like a very nice thing to have that isn't being taken advantage of. This is all we have right now:

```
./molproparser.py:467:                            self.logger.warn("setting mocoeff element to zero: %s" % detail)
./molproparser.py:736:            self.logger.info("Creating attribute hessian")
./gaussianparser.py:458:                    self.logger.debug(line)
./gaussianparser.py:621:                self.logger.debug(line)
./gaussianparser.py:626:                    self.logger.error("Problem parsing the value for geometry optimisation: %s is not a number." % parts[2])
./gaussianparser.py:1028:                self.logger.debug(parts)
./gaussianparser.py:1122:                    self.logger.warning("Number of basis functions (nbasis) has changed from %i to %i" % (self.nbasis, nbasis))
./gaussianparser.py:1213:                    self.logger.warning("Molecular coefficients header found but no coefficients.")
./orcaparser.py:190:                self.logger.warning('The geotargets attribute should not exist yet. There is a problem in the parser.')
./orcaparser.py:544:                self.logger.warning(msg)
./orcaparser.py:707:                    self.logger.warning('Overwriting previous multipole moments with new values')
./nwchemparser.py:552:                            self.logger.warning("Symmetry of MO %i has changed" % (index+1))
./nwchemparser.py:905:                self.logger.warning("Found quadrupole moments but no previous dipole")
./nwchemparser.py:960:                self.logger.warning("Found octupole moments but no previous dipole or quadrupole moments")
./nwchemparser.py:963:                self.logger.warning("Found octupole moments but no previous quadrupole moments")
./nwchemparser.py:1021:            self.logger.warning("Found a non-standard aoname representation type.")
./nwchemparser.py:1034:                self.logger.warning(msg)
./gamessparser.py:653:                    self.logger.warning(msg)
./gamessparser.py:658:                    self.logger.warning("\n   " + "\n   ".join((line1, line2, line3)))
./gamessparser.py:886:                    self.logger.warning('MO section found but could not be parsed!')
./gamessparser.py:1076:                            self.logger.warning("Number of occupied orbitals not consistent. This is normal for ECP and FMO jobs.")
./gamessparser.py:1118:                self.logger.info("Reading additional aooverlaps...")
./gamessparser.py:1248:                    self.logger.warning('Overwriting previous multipole moments with new values')
[...]

PR #694, opened 2019-02-10
Add value checking of data, and add check for negative etenergies

langner:
This implements a pattern for the data checks described in #237, and adds a check for negative values in `etenergies`. Initially I added `check_values` to the parser class, but I think checks like negative `etenergies` are not specific to any one parser and so belong in the data class. And, they are probably best done after parsing is done, since things may change during parsing. That being said, there may be things in #237 that clearly belong on the parser side (we have some already). Let's deal with that on a case by case basis.

If this looks OK I'll add more checks.

### DIFF
--- a/cclib/parser/data.py
+++ b/cclib/parser/data.py
@@ -1,15 +1,16 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2017, the cclib development team
+# Copyright (c) 2019, the cclib development team
 #
 # This file is part of cclib (http://cclib.github.io) and is distributed under
 # the terms of the BSD 3-Clause License.
 
 """Classes and tools for storing and handling parsed data"""
 
+import logging
+from collections import namedtuple
 
 import numpy
-from collections import namedtuple
 
 from cclib.method import Electrons
 from cclib.method import orbitals
@@ -289,6 +290,15 @@ class ccData(object):
             except ValueError:
                 args = (attr, type(val), self._attributes[attr].type)
                 raise TypeError("attribute %s is %s instead of %s and could not be converted" % args)
+
+    def check_values(self, logger=logging):
+        """Perform custom checks on the values of attributes."""
+        if hasattr(self, "etenergies") and any(e < 0 for e in self.etenergies):
+            negative_values = [e for e in self.etenergies if e < 0]
+            msg = ("At least one excitation energy is negative. "
+                   "\nNegative values: %s\nFull etenergies: %s"
+                   % (negative_values, self.etenergies))
+            logger.error(msg)
 
     def write(self, filename=None, indices=None, *args, **kwargs):
         """Write parsed attributes to a file.
--- a/cclib/parser/logfileparser.py
+++ b/cclib/parser/logfileparser.py
@@ -372,6 +372,9 @@ class Logfile(object):
             if not attr in _nodelete:
                 self.__delattr__(attr)
 
+        # Perform final checks on values of attributes.
+        data.check_values(logger=self.logger)
+
         # Update self.progress as done.
         if hasattr(self, "progress"):
             self.progress.update(inputfile.size, "Done")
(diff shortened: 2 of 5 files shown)


test/parser/testdata.py
```diff
@@ -23,11 +29,18 @@ class ccDataTest(unittest.TestCase):
...
 
     def _set_attributes(self, names, val):
         for n in names:
             setattr(self.data, n, val)
 
+    def test_valuecheck_negative_etenergies(self):
+        self.data.etenergies = [1, -1]
+        self.data.check_values(logger=self.mock_logger)
+        self.mock_logger.error.assert_called_once_with(
+            AnyStringWith("At least one excitation energy is negative"))
```
berquist: So this is similar to possible behavior using https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex, where the check is that part of the message contains `At least one excitation energy is negative`?

test/parser/testdata.py
```diff
@@ -1,19 +1,25 @@
...
 
+import mock
 import unittest
 
 import numpy
 
 import cclib
 
 
+class AnyStringWith(str):
```
berquist: I understand exactly what this does from reading `__eq__`, but not the class name. Can you change it to something containing `Contains` or similar?

cclib/parser/data.py
```diff
@@ -290,6 +291,15 @@ def typecheck(self):
...
                 raise TypeError("attribute %s is %s instead of %s and could not be converted" % args)
 
+    def check_values(self, logger=logging):
+        """Perform custom checks on the values of attributes."""
+        if hasattr(self, "etenergies") and any(e < 0 for e in self.etenergies):
+            negative_values = [e for e in self.etenergies if e < 0]
+            msg = ("At least one excitation energy is negative. "
+                   "\nNegative values: %s\nFull etenergies: %s"
+                   % (negative_values, self.etenergies))
+            logger.error(msg)
```
berquist: Using `ERROR` doesn't halt parsing, right?

cclib/parser/logfileparser.py
```diff
@@ -434,19 +437,28 @@ def new_internal_job(self):
...
+        
+        Parameters
+        ----------
+        name: str
+            The name of the attribute.
+        value: str
+            The value for the attribute.
+        check_change: bool
+            By default we want to check that the value does not change
+            if the attribute already exists.
```
berquist: Are we going to stick to NumPy-style docstrings? If we pick something, it should go in the developer docs.

test/parser/testdata.py
```diff
@@ -1,19 +1,25 @@
...
 #
-# Copyright (c) 2017, the cclib development team
+# Copyright (c) 2019, the cclib development team
 #
 # This file is part of cclib (http://cclib.github.io) and is distributed under
 # the terms of the BSD 3-Clause License.
 
 """Unit tests for parser data module."""
 
+import mock
```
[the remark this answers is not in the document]
langner: I'd prefer to use six for this kind of thing - done in 47454d3 for both modules. We could consolidate that in one place, but I don't think it's worth the effort for a few lines.

test/parser/testdata.py
langner: Right. I didn't want to put in the whole string verbatim, and I don't think there's anything built in like this in mock.

test/parser/testdata.py
langner: Changed to `StringContains`, hopefully that's good.

cclib/parser/data.py
langner: Not by default, no.

cclib/parser/logfileparser.py
langner: Added to #555. I'm not convinced by any particular format, but I'm OK with NumPy.

berquist: I've written each of those forms, I have no preference.

PR #695, opened 2019-02-13
Use ERROR loglevel for regressions by default, and add --debug option

langner:


langner:
Oh, it looks like I messed something up here...

langner:
Should be good now.

PR #675, opened 2018-12-24
Old regression tests errors are now counted.

shivupa:
Old regression tests were sometimes throwing errors that were being ignored (see #673). This lets us see the errors.
Before:
```
Total: 415   Failed: 0  Errors: 0
```
After:
```
Total: 415   Failed: 0  Errors: 57
```
This doesn't fix any of the errors; it just makes them visible.

PR #698, opened 2019-02-17
Regressions: skip test methods requiring scfvalues when they are not present for one old ADF unit test

langner:


ATenderholt:
Seems like this can go in before #675?

langner:
> Seems like this can go in before #675?

Yes, and needs to.

Issue #665, opened 2018-12-06
Add missing MOPAC examples


berquist:
As mentioned in https://github.com/cclib/cclib/issues/664, we are missing many MOPAC examples for unit tests. This issue is to help figure out what attributes are  missing coverage, and check if new attributes are needed.

langner:
> Also,
> 
> > It might be useful for some people to separate out the calculated heat-of-formation from the actual SCF total energy. We’ve found the latter to be more useful for machine learning, for example.

Is this specific to MOPAC?

berquist:
Yes, I think so.

langner:
Eric, can we close this for v1.6.3 or is there something actionable left here?

langner:
> Eric, can we close this for v1.6.3 or is there something actionable left here?

I'm going to assume we can close this out.

PR #700, opened 2019-02-21
MOPAC: enable GeoOpt unit test and move logfile to basic* subdir as is the convention

langner:
This doesn't really address #665, but gets the ball rolling for MOPAC unit tests.


langner:
Any issues here blocking the merge?

ATenderholt:
Nope, looks good.

PR #702, opened 2019-02-21
Skipping Psi4 atommasses absent in old tests.

shivupa:
Similar to #698, we skip atommasses for psi4 which weren't present in the older files. This covers all of the psi4 errors in #675


langner:
Thanks!

PR #703, opened 2019-02-22
Remove redundant test

shivupa:
The file [regression C_bigbasis](https://github.com/cclib/cclib-data/blob/master/Molcas/Molcas8.0/C_bigbasis.out) has strong overlap with [C_bigbasis](https://github.com/cclib/cclib/blob/master/data/Molcas/basicOpenMolcas18.0/C_bigbasis.out). However this test wasn't run with the appropriate print level so `gbasis` can't be parsed.

Ignoring the triggered errors for a basis test would result in nothing being tested (all the tests require `gbasis`). Since it has strong similarities to the other test that does contain `gbasis`, I decided it would be proper to just remove it. 

Relevant PR #675

### DIFF
--- a/test/regression.py
+++ b/test/regression.py
@@ -1888,6 +1888,23 @@ class JaguarGeoOptTest_6_31gss(GenericGeoOptTest):
     nbasisdict = {1: 5, 6: 15}
     b3lyp_energy = -10530
 
+class MolcasBigBasisTest_nogbasis(MolcasBigBasisTest):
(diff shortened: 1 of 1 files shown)


langner:
I would prefer to not remove regression files, even if they are not add any new checks. Please, could you make it so it is just parsed as part of the regression test?

shivupa:
@langner How does this look?

langner:
> @langner How does this look?

That's good, thanks!

Issue #705, opened 2019-03-04
Error parsing vibrational frequencies in ORCA 4.1.1


schneiderfelipe:
Hi all,

`orcaparser.py` throws the following error when reading an output file from ORCA 4.1.1 (released December 2018):

```
In [1]: import cclib

In [2]: cclib.ccopen("methylpheophorbideB.out")
Out[2]: ORCA("methylpheophorbideB.out")

In [3]: of = cclib.ccopen("methylpheophorbideB.out")

In [4]: of
Out[4]: ORCA("methylpheophorbideB.out")

In [5]: of.parse()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-5-464ffa4a8ac3> in <module>()
----> 1 of.parse()

/usr/local/lib/python3.6/dist-packages/cclib/parser/logfileparser.py in parse(self, progress, fupdate, cupdate)
    316             # Any attributes can be freely set and used across calls, however only those
    317             #   in data._attrlist will be moved to final data object that is returned.
--> 318             self.extract(inputfile, line)
    319 
    320         # Close input file object.

/usr/local/lib/python3.6/dist-packages/cclib/parser/orcaparser.py in extract(self, inputfile, line)
    939             vibfreqs = numpy.zeros(3 * self.natom)
    940             for i, line in zip(range(3 * self.natom), inputfile):
--> 941                 vibfreqs[i] = float(line.split()[1])
    942 
    943             nonzero = numpy.nonzero(vibfreqs)[0]

ValueError: could not convert string to float: 'factor'
```

[(The outfile is attached.)](https://github.com/cclib/cclib/files/2926679/methylpheophorbideB.zip)

Relevant lines in `orcaparser.py` (936--941):

    (...)
    if line[:23] == "VIBRATIONAL FREQUENCIES":
        self.skip_lines(inputfile, ['d', 'b'])

        vibfreqs = numpy.zeros(3 * self.natom)
        for i, line in zip(range(3 * self.natom), inputfile):
            vibfreqs[i] = float(line.split()[1])
    (...)

It looks like the new output is as follows:

```
(...)
-----------------------
VIBRATIONAL FREQUENCIES
-----------------------

Scaling factor for frequencies =  1.000000000  (already applied!)

   0:         0.00 cm**-1
   1:         0.00 cm**-1
   2:         0.00 cm**-1
   3:         0.00 cm**-1
   4:         0.00 cm**-1
   5:         0.00 cm**-1
[...]

PR #706, opened 2019-03-04
ORCA 4.1.x adds a new line in vibrational frequencies output (test data update)

schneiderfelipe:
This solves #705. The new line about the scaling factor breaks the present ORCA parser:

```
(...)
-----------------------
VIBRATIONAL FREQUENCIES
-----------------------

Scaling factor for frequencies =  1.000000000  (already applied!)

   0:         0.00 cm**-1
   1:         0.00 cm**-1
   2:         0.00 cm**-1
   3:         0.00 cm**-1
   4:         0.00 cm**-1
   5:         0.00 cm**-1
   6:       -97.00 cm**-1 ***imaginary mode***
(...)
```

This PR also adds a single test instance for ORCA 4.1.x. If it's OK, I can rerun and add all tests from ORCA 4.0 to 4.1.


schneiderfelipe:
I updated it so that all test data is now updated to 4.1.x (work was based on PR #575).

ATenderholt:
Thanks! I'm happy to merge it as is, even though the new test files don't follow the 'copy from old test files --> commit --> update for new version --> commit' steps.

What do @langner and @berquist think? Is it ok to not have history of the new test files this time?

langner:
I'd say it's not too important, but @schneiderfelipe if you can manage to keep the history that would be preferable. I can also do it for you if it's too much trouble. Let me know.

Like Adam said, it's basically three steps:
1. `git gopy` the latest checked in logfiles to the new location.
2. Update the copied files with the new outputs.
3. (optional) move oldest test files over to regressions

schneiderfelipe:
I moved regressions to cclib-data ([PR here](https://github.com/cclib/cclib-data/pull/108)).

langner:
Looks great, thanks!

PR #710, opened 2019-03-07
 New CODE_OF_CONDUCT.md

Chokerino:
Need to check the email address provided in the enforcement section.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
+level of experience, education, socio-economic status, nationality, personal
+appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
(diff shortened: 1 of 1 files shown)


langner:
I don't think we need this as a top level file, and I don't think we need a verbatim code of conduct. It's OK is we want to link to something like this in our guidelines, but IMO that should be sufficient.

@berquist @ATenderholt thoughts?

berquist:
What are other open-source computational chemistry projects doing?

Chokerino:
For GSoC, I'm not working on any other open source project but I intend to
work with my professor on some biology/chemistry related computational
projects this summer.

shivupa:
I think the question is "What are other open source chemistry programs using as a code of conduct?"

There are some projects that include the code of conduct at top level:
https://github.com/MolSSI/QCEngine/blob/master/CODE_OF_CONDUCT.md
https://github.com/MDAnalysis/mdanalysis/blob/develop/CODE_OF_CONDUCT.md
https://github.com/deepchem/deepchem/blob/master/CODE_OF_CONDUCT.md

I didn't check if some projects kept the code of conduct in the documentation or in some other place. (I just skimmed through my starred projects and the majority of them didn't have a code of conduct that I could see)

Chokerino:
Yes, many other GSoC orgs do not have any code of conduct at the top level.

ATenderholt:
I don't really have a preference. It sounds like it's used by major projects (https://www.contributor-covenant.org/), so there's at least precedent for including it. Since it's the "contributor" code of conduct, it makes more sense to be with docs about contributing to cclib (i.e. not top level) unless we modify it to also include general users.

langner:
OK then, it sounds like the consensus is somewhere around putting this in the `docs` directory. @Chokerino can you do that? (and maybe integrate that into the menu)

Chokerino:
Yes, and also will be adding some of the things u wanted to be included.

Chokerino:
I have created a pull request to add the conduct in the docs directory. Also about adding it to the menu, do you want me to add it onto the website or the GitHub page( which I have proposed in this pull request).

langner:
Closing this for similar reasons to #713

PR #712, opened 2019-03-08
Replace sys.argv/getopt/logic with argparse

berquist:
And some fixes for propagating the command-line logging level. Related: https://github.com/cclib/cclib/pull/695

### DIFF
--- a/cclib/scripts/ccget.py
+++ b/cclib/scripts/ccget.py
@@ -14,9 +14,9 @@ from __future__ import print_function
 import glob
 import logging
 import os.path
+import sys
 from functools import partial
 from pprint import pprint
-from sys import version_info
 
 import numpy
 
@@ -37,7 +37,7 @@ from cclib.io import ccread, URL_PATTERN
 # """
 
 # Set up options for pretty-printing output.
-if version_info < (3, 4):
+if sys.version_info < (3, 4):
     pprint = partial(pprint, width=120)
 else:
     pprint = partial(pprint, width=120, compact=True)
--- a/cclib/scripts/ccget.py
+++ b/cclib/scripts/ccget.py
@@ -24,18 +24,6 @@ from cclib.parser import ccData
 from cclib.io import ccread, URL_PATTERN
 
 
-# MSG_USAGE_LONG = """\
-# Usage:  ccget <attribute> [<attribute>] <compchemlogfile> [<compchemlogfile>]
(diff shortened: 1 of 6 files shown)


cclib/scripts/ccget.py
```diff
@@ -11,48 +11,33 @@
...
+# Usage:  ccget <attribute> [<attribute>] <compchemlogfile> [<compchemlogfile>]
+#     where <attribute> is one of the attributes to be parsed by cclib
+#     from each of the compchemlogfiles.
+# For a list of attributes available in a file, use --list (or -l):
+#     ccget --list <compchemlogfile>
+# To extract attributes available in a CJSON file, use --json (or -j):
+#     ccget --json  <attr> [<attr>]  <compchemlogfile>
+# To parse multiple files as one input stream, use --multi (or -m):
+#     ccget --multi <attr> [<attr>]  <cclogfile> <cclogfile> [<cclogfile>]
+# """
```
ATenderholt: Is there a reason to keep this commented out here?

berquist:
Everything should be fixed now. I don't think this solution is quite as customizable as using `getopt`, and most of the logic for `ccget` in particular is still there. I guess the advantage is it's easier to maintain? I couldn't figure out how to generate the nice long message we had before either.

langner:
> I couldn't figure out how to generate the nice long message we had before either.

What message?

berquist:
The one from `MSG_USAGE_LONG`, I don't think what's in this PR is as clear.

berquist:
It looks like
```
$ ccget --help
usage: ccget [-h] [--list | --json | --multi] [--verbose] [--future] [--full]
             attribute_or_compchemlogfile [attribute_or_compchemlogfile ...]

positional arguments:
  attribute_or_compchemlogfile
                        one or more attributes to be parsed from one ore more
                        logfiles

optional arguments:
  -h, --help            show this help message and exit
  --list, -l            print a list of attributes available in each file
  --json, -j            the given logfile is in CJSON format
  --multi, -m           parse multiple input files as one input stream
  --verbose, -v         more verbose parsing output (only errors by default)
  --future, -u          use experimental features (currently optdone_as_list)
  --full, -f            toggle full print behaviour for attributes
```

langner:
@ATenderholt are you OK with this? I think this is an improvement, and we can figure out a better long help message later on.

Issue #620, opened 2018-09-02
Improve cclib's community profile


langner:
Look at https://github.com/cclib/cclib/community, we might want to:
1. Add a code of conduct
2. Add contributing guidelines, which is essentially #555, but we might want to make sure GitHub is aware of it so we get all the benefits.
3. Consider making issue and PR tempaltes.

ghutchis:
The issue templates in particular are a big win - people do a much better job writing bug reports with them IMHO. Feel free to steal from Avogadro if you like.
https://github.com/OpenChemistry/avogadrolibs/tree/master/.github

You might also want to check out the Probot apps - several very nice ones for automation:
https://probot.github.io/apps/

langner:
I'm sad I didn't get around to this, pushing back to v.1.6.2.

langner:
Note to self: finish this before GSOC starts this year.

Chokerino:
Do the contributing guidelines need to be written by self or can I edit some other repo's?

langner:
> Do the contributing guidelines need to be written by self or can I edit some other repo's?

We have a bunch of very specific things we want to go into the guidelines - please take a look at #555. Although there will, of course, be some generic things in there, I think the most valuable for cclib itself will be those specific items we enumerated in #555.

That being said, I would have nothing against you starting up the guidelines based on some generic text, which we can modify to our needs. Do you have something in mind already?

langner:
Ah, we can now add some of things mentioned here with GitHub actions.

PR #713, opened 2019-03-08
Added Code of Conduct to Docs

Chokerino:


Chokerino:
IF a link needs to be attached in the website please let me know.

langner:
I don't mean to berate your effort, but I don't think this text in its current form is useful for the project. I wholeheartedly agree with the sentiments expressed, and believe we have fostered an inclusive and welcoming environment here. But just copying a snippet from another community's code of conduct doesn't achieve much. More importantly, for us this specific formulation is not actionable: IMO it is not appropriate to send or investigate complaints on a public user mailing list.

So, if you are passionate about this topic, please give it more thought first. I think there is more immediate value in adding a general document about contributor guidelines as outlined in #555 and #620, and putting something like this (maybe part of it) within that context.

langner:
P. S. @berquist @ATenderholt and any other devs please let me know if you disagree!

Issue #598, opened 2018-08-14
Consistent import of soft dependencies


berquist:
As mentioned in https://github.com/cclib/cclib/pull/584#discussion_r209109117, we might want a safer import mechanism for optional dependencies. How to do it is outlined in https://stackoverflow.com/questions/14050281/how-to-check-if-a-python-module-exists-without-importing-it/25045228.

PR #716, opened 2019-03-10
Check for soft dependencies without importing them

berquist:
Closes https://github.com/cclib/cclib/issues/598.


test/io/testscripts.py
```diff
@@ -90,42 +91,36 @@ def test_ccwrite_call(self, mock_ccwrite):
...
-        """is ccframe called with the given parameters?"""
-        self.main()
+        """Is ccframe called with the given parameters?"""
+        cclib.scripts.ccframe.main()
 
         self.assertEqual(mock_ccframe.call_count, 1)
         ccframe_call_args, ccframe_call_kwargs = mock_ccframe.call_args
         self.assertEqual(ccframe_call_args[0][0].filename, INPUT_FILE)
 
+    @unittest.skip("Can't figure out why this is failing")
```
berquist: I need help from someone with mock experience to fix this.

I needed to get rid of `self.main` so I could test the function call inside of `ccframe()` by reimporting it each time, but `scripts/ccframe` doesn't seem to be triggering the system exit in `io/ccio/ccframe`.

langner: I don't quite follow what you need help with here, but it looks like you figure it out? :)

berquist: So you're ok with the skipped test? I needed to do that because I don't know how to use `mock`.

langner: Hmm... see berquist/cclib#9 for one way to go about this. I think here it's better to test what actually happens and not the mocks, which in this case means testing what's printed.

Ideally we would have pandas as a requirement and make sure that works end to end as well.

berquist: So then would you be ok with adding pandas as a first-class dependency?

ATenderholt: Isn't pandas a bit heavy to be a first-class dependency? Unless it's used in other places besides `ccframe.py`, it seems like it should be an optional dependency.

berquist: That was our initial conclusion. It is only required for ccframe and it should fail I just made a fresh virtualenv, and here is what installing pandas requires:
```
$ pip install pandas
Collecting pandas
  Downloading https://files.pythonhosted.org/packages/22/e6/2d47835f91eb010036be207581fa113fb4e3822ec1b4bafb0d3d105fede6/pandas-0.24.2-cp37-cp37m-manylinux1_x86_64.whl (10.1MB)
     |████████████████████████████████| 10.1MB 1.3MB/s
Collecting python-dateutil>=2.5.0 (from pandas)
  Using cached https://files.pythonhosted.org/packages/41/17/c62faccbfbd163c7f57f3844689e3a78bae1f403648a6afb1d0866d87fbb/python_dateutil-2.8.0-py2.py3-none-any.whl
Collecting pytz>=2011k (from pandas)
  Using cached https://files.pythonhosted.org/packages/3d/73/fe30c2daaaa0713420d0382b16fbb761409f532c56bdcc514bf7b6262bb6/pytz-2019.1-py2.py3-none-any.whl
Collecting numpy>=1.12.0 (from pandas)
  Downloading https://files.pythonhosted.org/packages/bb/76/24e9f32c78e6f6fb26cf2596b428f393bf015b63459468119f282f70a7fd/numpy-1.16.3-cp37-cp37m-manylinux1_x86_64.whl (17.3MB)
     |████████████████████████████████| 17.3MB 69.3MB/s
Collecting six>=1.5 (from python-dateutil>=2.5.0->pandas)
  Using cached https://files.pythonhosted.org/packages/73/fb/00a976f728d0d1fecfe898238ce23f502a721c0ac0ecfedb80e0d88c64e9/six-1.12.0-py2.py3-none-any.whl
Installing collected packages: six, python-dateutil, pytz, numpy, pandas
Successfully installed numpy-1.16.3 pandas-0.24.2 python-dateutil-2.8.0 pytz-2019.1 six-1.12.0
```
With 1.6.1, here is what happens when pandas isn't installed:
```
$ ccframe -O c4h4_freq.csv c4h4_freq.out
  File "/home/eric/data/virtualenvs/pandas/bin/ccframe", line 6, in <module>
    from cclib.scripts.ccframe import main
  File "/home/eric/data/virtualenvs/pandas/lib/python3.7/site-packages/cclib/__init__.py", line 28, in <module>
    from cclib import parser
  File "/home/eric/data/virtualenvs/pandas/lib/python3.7/site-packages/cclib/parser/__init__.py", line 37, in <module>
    from cclib.io.ccio import ccopen
  File "/home/eric/data/virtualenvs/pandas/lib/python3.7/site-packages/cclib/io/__init__.py", line 24, in <module>
    from cclib.io.ccio import ccframe
  File "/home/eric/data/virtualenvs/pandas/lib/python3.7/site-packages/cclib/io/ccio.py", line 21, in <module>
    from six import string_types
ModuleNotFoundError: No module named 'six'
```
And then after installing six:
```
$ ccframe -O c4h4_freq.csv c4h4_freq.out
You need to have pandas installed
$ echo $?
1
```
With this PR, the issue with six is still present, but then after installing that here is what happens:
```
$ ccframe -O c4h4_freq.csv c4h4_freq.out
[NWChem c4h4_freq.out INFO] Creating attribute metadata: {}
$ echo $?
1
```
with no CSV file generated.

There is a bug somewhere...

berquist:
Ok, should be all set for review again.

test/io/testscripts.py
```diff
@@ -103,14 +103,17 @@ def test_main_empty_argv(self):
         with self.assertRaises(SystemExit):
             cclib.scripts.ccframe.main()
 
+    @unittest.skipIf(sys.version_info < (3, 2), "assertRaisesRegex is Python >= 3.2")
```
ATenderholt: I'm a bit concerned that this won't ever get tested with Python 2.7+. I'd be fine with letting it slide if there wasn't an elegant solution, but it looks like  just plain `assertRaises` could be used and the exception message checked afterwards. See the example in the documentation https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises.

cclib/bridge/__init__.py
```diff
@@ -7,23 +7,13 @@
 
 """Facilities for moving parsed data to other cheminformatic libraries."""
 
-try:
-    import openbabel
-except ImportError:
-    pass
-else:
+from cclib.parser.utils import find_package as _find_package
```
langner: Leading underscores are typically used for names internal to the imported module - which is not the case here.

https://www.python.org/dev/peps/pep-0008/#id36

test/io/testscripts.py
berquist: Ok, I misread the documentation and didn't know that `assertRaisesRegexp` works on all versions.

PR #718, opened 2019-03-18
Log to stderr by default

langner:


ATenderholt:
As pointed out by @langner:

> Typically logging and other debugging messages are printed to stderr (https://stackoverflow.com/questions/4919093/should-i-log-messages-to-stderr-or-stdout).

PR #719, opened 2019-03-29
fixing Gamess and Gaussian regression errors 

amandadumi:
This corrects all remaining errors in the regression tests for issue #673, specifically for Gamess and Gaussian.  The ADF errors will still be present on this branch, but I didn't want to alter that here since those changes are already on master.

My text editor removed some superfluous tabs on blank lines, would you like me to add those back in?

PR #715, opened 2019-03-09
Addressing ADF regression tests so values not present in old files are not parsed

amandadumi:
Fixing old tests for remaining ADF regression test errors in #673

PR #714, opened 2019-03-09
Fixing Jaguar tests to not test values that are not present in old regression tests.

shivupa:
Addresses all the jaguar regression test errors in #673

PR #720, opened 2019-03-30
Fix silent regression failures

berquist:
Closes https://github.com/cclib/cclib/issues/673.

All of the PRs merged into this branch have already been reviewed. Thank you @shivupa and @amandadumi for doing all the hard work.

Related: #675, #714, #715, #719


berquist:
The `push` build is failing because of the ADF tests that are already in master; `push` is the test of the branch as-is, and `pr` is the test of PR branch merged into the base branch, so we know that this works.

I'll let @langner decide if this should be added with a merge commit or squashed.

PR #721, opened 2019-04-03
Making test/regressions.py PEP8 compliant

amandadumi:
The spacing between top level functions classes and their methods are PEP8 compliant. A few lines have some spacing discrepancies with PEP8 I can also fix, but I wanted to check in before making those few additional changes.


langner:
Thanks!

Issue #723, opened 2019-04-10
Comment line for XYZ files not stored


vingman:
When parsing XYZ files, the comment line is not stored anywhere, nor can the default comment be overwritten when writing XYZ files.

berquist:
Good to know that someone is using the XYZ code. For reading, the comment can be placed as an entry into `metadata`, with the empty string in the case that one wasn't present.

Right now, when writing out, the code looks like
```python
        # Create a comment derived from the filename and the index.
        if index == -1:
            geometry_num = len(self.ccdata.atomcoords)
        else:
            geometry_num = index + 1
        if self.jobfilename is not None:
            comment = "{}: Geometry {}".format(self.jobfilename, geometry_num)
        else:
            comment = "Geometry {}".format(geometry_num)
```
so we could replace part of this with looking up `self.ccdata.metadata["comment"]`. I'd like information to stay about the geometry number if more than one geometry has been requested, otherwise grepping for how many frames there are is more difficult.

Is that enough, or were you thinking of something else?

langner:
I think we could also make this more convenient by adding a `comment` argument into the `XYZ` class somewhere that does the writing. Internally using `metadata` to store the comment is fine, of course.

PR #724, opened 2019-04-19
Handle existing comments for XYZ readers and writers

berquist:
Closes https://github.com/cclib/cclib/issues/723.


cclib/io/xyzwriter.py
```diff
@@ -91,6 +93,12 @@ def _xyz_from_ccdata(self, index):
             comment = "{}: Geometry {}".format(self.jobfilename, geometry_num)
         else:
             comment = "Geometry {}".format(geometry_num)
+        # Wrap the geometry number part of the comment in square brackets,
+        # prefixing it with one previously parsed if it existed.
+        if existing_comment:
+            comment = "{} [{}]".format(existing_comment, comment)
+        else:
+            comment = "[{}]".format(comment)
```
berquist: For
> I think we could also make this more convenient by adding a comment argument into the XYZ class somewhere that does the writing. Internally using metadata to store the comment is fine, of course.

we could take this and make a new method out of it. Having a property on the class wouldn't make as much sense until we unify the readers and writers.

PR #730, opened 2019-04-26
Base file reader and writer classes should be abstract

berquist:
Per https://github.com/cclib/cclib/pull/454#discussion_r166019944 and https://github.com/cclib/cclib/pull/517#discussion_r193953499, `io.filereader.Reader` and `io.filewriter.Writer` should be abstract base classes. I didn't do this in those PRs because...I didn't understand what an ABC was then.

We need `__metaclass__` rather than `class Reader(ABC)`, etc., because of Python 2.

Separately, it looks like `parser.logfileparser.Logfile` should be abstract too, with `before_parsing()` and `after_parsing()` as abstract methods?


test/io/testcjsonreader.py
```diff
@@ -24,6 +24,10 @@
 class CJSONReaderTest(unittest.TestCase):
     """Unit tests for the CJSON reader."""
 
+    def test_subclass(self):
+        """Is the reader a subclass of the abstract file reader?"""
+        self.assertTrue(issubclass(cclib.io.cjsonreader.CJSON, cclib.io.filereader.Reader))
```
ATenderholt: What's your take on using `assertIsInstance` instead? https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIsInstance

test/io/testcjsonwriter.py
```diff
@@ -31,6 +31,15 @@ def test_init(self):
         # The object should keep the ccData instance passed to its constructor.
         self.assertEqual(cjson.ccdata, data)
 
+    def test_subclass(self):
+        """Is the writer a subclass of the abstract file writer?"""
+        fpath = os.path.join(__datadir__, "data/ADF/basicADF2007.01/dvb_gopt.adfout")
+        self.assertTrue(os.path.exists(fpath))
+        data = cclib.io.ccread(fpath)
+        writer = cclib.io.cjsonwriter.CJSON(data)
```
ATenderholt: Can you just use an empty `ccData` instance instead of reading from disk?

ATenderholt:
Added a couple of comments that may be applicable elsewhere.

cclib/io/filewriter.py
```diff
@@ -8,8 +8,15 @@
 """Generic file writer and related tools"""
 
 import logging
+import sys
+from abc import ABCMeta, abstractmethod
+from six import add_metaclass
 
-from collections import Iterable
```
langner: Doesn't this still work in Python3?

test/io/testcjsonreader.py
langner: I don't think that's applicable here, since CJSON is a class, not an instance of it.

However, I'm not sure we really need to make this assertion at all, since it's mostly testing Python's implementation and doesn't say anything about the behavior of cclib per se. Instead, is there an assertion we can make about CJSON or an instance of it that relies on it being an implementation of the abstract Reader?

cclib/io/filewriter.py
berquist: It is deprecated.

test/io/testcjsonwriter.py
berquist: Ideally, but
```
======================================================================
ERROR: test_init (test.io.testcjsonwriter.CJSONWriterTest)
Does the class initialize correctly?
----------------------------------------------------------------------
  File "/home/eric/development/cclib_berquist/test/io/testcjsonwriter.py", line 28, in test_init
    cjson = cclib.io.cjsonwriter.CJSON(data)
  File "/home/eric/development/cclib_berquist/cclib/io/cjsonwriter.py", line 34, in __init__
    super(CJSON, self).__init__(ccdata, terse=terse, *args, **kwargs)
  File "/home/eric/development/cclib_berquist/cclib/io/filewriter.py", line 70, in __init__
    self.obmol, self.pbmol = self._make_openbabel_from_ccdata()
  File "/home/eric/development/cclib_berquist/cclib/io/filewriter.py", line 107, in _make_openbabel_from_ccdata
    obmol = makeopenbabel(self.ccdata.atomcoords,
AttributeError: 'ccData' object has no attribute 'atomcoords'
```

berquist:
There are no hard feelings. I don't think it provides any features we need to work correctly. I like it because it enforces a certain API and helps with development and maintenance. For example, `generate_repr` is something all readers and writers should currently have, but if we ever switch to a streaming API, this will help us make that transition by complaining until each method is implemented (like [this](https://stackoverflow.com/a/30328931)). I'm not sure we need the more advanced features from [here](https://stackoverflow.com/a/19328146) yet.

We are probably better served by thinking about what design we'd like. There was talk at some point about merging the `Reader` and `Logfile` classes.

langner:
> There are no hard feelings. I don't think it provides any features we need to work correctly. I like it because it enforces a certain API and helps with development and maintenance. For example, `generate_repr` is something all readers and writers should currently have, but if we ever switch to a streaming API, this will help us make that transition by complaining until each method is implemented (like [this](https://stackoverflow.com/a/30328931)). I'm not sure we need the more advanced features from [here](https://stackoverflow.com/a/19328146) yet.

I agree on all these points. It definitely provides a certain space for development. Using the abc doesn't introduce much overhead, as long as everyone's comfortable with it. I think we can merge this after another round of reviews.

> We are probably better served by thinking about what design we'd like. There was talk at some point about merging the `Reader` and `Logfile` classes.

My impression was that we should merge the readers and writers into one class. Those are seriously sharing code, at least when I last looked at the CJSON stuff.

test/io/testcjsonwriter.py
TDI-tenderholt: What about an empty `ccData` instance where you add required attributes? Something like  `data.atomcoords = numpy.zeros((1,3))`)?

TDI-tenderholt:
I'm all for `abc` when it makes sense, which I think is true for the base `Reader` and `Writer` classes.

test/io/testcjsonwriter.py
langner: I agree, it'd be better to have some kind of fixture here instead of relying on the parser working correctly.

test/io/testfilereader.py
```diff
@@ -0,0 +1,37 @@
...
+__filepath__ = os.path.realpath(__filedir__)
+__datadir__ = os.path.join(__filepath__, "..", "..")
+
+
+class FileReaderTest(unittest.TestCase):
+
+    def test_type(self):
+        """Is the class properly a metaclass?"""
+
+        self.assertTrue(inspect.isabstract(cclib.io.filereader.Reader))
```
langner: We can leave this, but I don't see much value in testing the implementation of Python. Is there something meaningful about the base class that we can test?

berquist:
I've removed tests that just check the abstract class hierarchy.

langner:
I think this has been lingering long enough... we can improve more going forward.

PR #732, opened 2019-05-08
Metadata: full package versions

### LINKED ISSUE #533: Ensure metadata contains most detailed package version

berquist:
Closes #533.

Needs tests to make sure that parsing the version doesn't produce a `LegacyVersion`, which is something that would violate PEP 440.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+packaging>=19.0
 numpy
 # We need this as long as we need support both Python 2 and 3.
 six
--- a/test/regression.py
+++ b/test/regression.py
@@ -99,10 +99,14 @@ def testADF_ADF2004_01_Fe_ox3_final_out(logfile):
     """Make sure HOMOS are correct."""
     assert logfile.data.homos[0] == 59 and logfile.data.homos[1] == 54
 
+    assert logfile.data.metadata["package_version"] == "2004.01+200410211341"
+
 
 def testADF_ADF2013_01_dvb_gopt_b_unconverged_adfout(logfile):
     """An unconverged geometry optimization to test for empty optdone (see #103 for details)."""
     assert hasattr(logfile.data, 'optdone') and not logfile.data.optdone
+
+    assert logfile.data.metadata["package_version"] == "2013.01+201309012319"
 
 
 def testADF_ADF2013_01_stopiter_dvb_sp_adfout(logfile):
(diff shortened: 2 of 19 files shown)


langner:
> Needs tests

Do you mean you want to add tests here, or later?

berquist:
I'd like to add tests here. Thinking:
- Add to https://github.com/cclib/cclib/blob/4fe80270e40b9150319b3dc309b028f22b894f13/test/data/testSP.py#L301 checking that it parses properly
- For each test in regressions, add a line checking that it parses properly.

By "parse to `Version` and not `LegacyVersion`", I mean
```python
In [53]: from packaging.version import parse as parse_version

In [54]: v2 = parse_version("5.1.0dev+branches_libresponse-27553")

In [55]: v2
Out[55]: <Version('5.1.0.dev0+branches.libresponse.27553')>
```
https://stackoverflow.com/a/11887885/ is a good explanation of why this is important.

test/regression.py
```diff
@@ -749,43 +1018,59 @@ def testGaussian_Gaussian09_issue_460_log(logfile):
     assert logfile.data.scfvalues[0][0, 0] == 3.37e-03
     assert numpy.isnan(logfile.data.scfvalues[0][0, 2])
 
+    assert logfile.data.metadata["package_version"] == "2009+D.01"
```
langner: Do you think it's a good idea to change the version formatting programs use natively? IIRC this is originally something like '09 Rev D01'.

requirements.txt
```diff
@@ -1,3 +1,4 @@
+packaging>=19.0
```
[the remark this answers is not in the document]
berquist: No, which is confusing to me. On my system, the `setuptools` package requires `packaging`, but working in a clean virtualenv, without requesting the `packaging` install:
```
/home/eric/development/cclib_berquist (metadata-full-package-versions|✚2…) (env: test_cclib_install 3.7.3)
[eric@osmium]$ pip list
Package    Version
---------- -------
cclib      1.6.1
numpy      1.16.3
pip        19.1.1
setuptools 41.0.1
six        1.12.0
wheel      0.33.4
/home/eric/development/cclib_berquist (metadata-full-package-versions|✚2…) (env: test_cclib_install 3.7.3)
[eric@osmium]$ python
Python 3.7.3 (default, Mar 26 2019, 21:43:19) 
[GCC 8.2.1 20181127] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from packaging.version import parse
ModuleNotFoundError: No module named 'packaging'
>>> 
```
There is a PEP 440 parser in `setuptools`, but is much less feature complete than the one in `packaging`.

test/regression.py
berquist: I don't think it's too confusing, and most native formatting is not PEP 440 compliant and will parse to `LegacyVersion`, so it won't sort properly, major/minor versions can't be parsed out of it, etc., which is important if we want to use the version numbers in the parsers, and QCSchema will require it.

requirements.txt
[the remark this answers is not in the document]
langner: Got it.

test/regression.py
langner: OK, sounds good. Perhaps it would be prudent to save the legacy version as well as the PEP 440 compliant version we generate from it?

berquist: Yes, will do. How does `package_legacy_version` sound?

langner: Might be better to leave the original name in there, so maybe `legacy_package_version` or `package_version_legacy`?

berquist: `legacy_package_version` it is.

berquist:
I think this is ready for re-review.

langner:
I think some of the parsers currently skipped in `testSP` can be turned on now, no? But this is good enough to go in, we can continue with the tests downstream.

Issue #602, opened 2018-08-26
Re-use test methods or reorganize by attribute


langner:
So the issue of unit test code re-use and redundancy has come up several times. For example:
* #599 where `metadata` has its own unit test module because its contents can depend on the calc type.
* cclib/cclib-data#19 where an `atomcoords` test was added to IR regressions - this should just go in the IR test module in the current setup.

I think there is something of a consensus that testing attributes repeatedly for different calc types would not be a bad thing, something we have avoided up till now. I think it makes sense to just restructure our unit test modules to be organized by attributes, not by calculation type. This will make testing attributes for multiple calc types easier, and the test suite probably more sparse. I don't think we necessarily want one module per attribute, but maybe that is the way to go.

However, it'd be good to retain the ability to run tests just for a specific calc type which I think is possible today, and that might require some extra test code. So marking this for the next release after v1.5.4.

langner:
Hasn't happened yet, so closing. We can create a new issue if we want this in the future.

PR #737, opened 2019-05-18
Implement per-attribute data unit tests parameterized with decorators

langner:
**(This is a proposal, please don't submit)**

So I had a go at #602 , at it wasn't too hard to implement. The result is kind of similar to what we have now in `test/test_data.py` just in decorators and encapsulated in a test case class. I like the idea of decorator syntax and relying more on the standard test framework.

Beyond the framework, I migrated the `BOMD` unit tests here, because it's only 4 files and 3 test method right now. This translates into three new test files, which can coexist with the current test system. Which means the migration could be incremental. Travis is currently failing due to relative imports I used.

Some cons I now observe:
- The logic in the decorator implementation is brittle, so it might be harder to maintain.
- Organizing tests by attributes seemed natural, until I actually tried it. There are certain bits that tie together the test methods in our current test modules. For `BOMD`, this is the number of energies/steps. Organizing by attribute disperses this across files for all the attributes being tested.
- This organization doesn't play well with our regression suite... which are all job-type focused.

The second and third points are most important IMO. I wonder if there is some middle ground where we could get the best of both worlds. Perhaps something that leverages attribute classes like we thought about for 2.x. Or maybe we should leave the current organization and just migrate to a decorator-based framework if you think it's an improvement.

Thoughts? I have some more ideas about how to change this, but want to first get your guys' general input.

### DIFF
--- a/test/data/common.py
+++ b/test/data/common.py
@@ -5,7 +5,7 @@
 # This file is part of cclib (http://cclib.github.io) and is distributed under
 # the terms of the BSD 3-Clause License.
 
-"""Functions used across multiple data tests."""
+"""Helpers used across multiple data tests."""
 
 import itertools
 
--- a/test/data/paths.py
+++ b/test/data/paths.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2019, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
+
+UNIT_TEST_PATHS = {
+    'BOMD': {
+        'Gaussian': {
+            'basicGaussian09': ('dvb_bomd.out', ),
+        },
+        'NWChem': {
+            'basicNWChem6.6': ('dvb_bomd_ks.out',),
+        },
+        'QChem': {
+            'basicQChem4.2': ('dvb_bomd.out',),
+            'basicQChem5.1': ('dvb_bomd.out',),
+        },
+    },
+}
--- a/test/data/testBOMD.py
+++ b/test/data/testBOMD.py
@@ -23,32 +23,6 @@ class GenericBOMDTest(unittest.TestCase):
     nsteps = 35
     nenergies = 36
 
-    def testdimscfenergies(self):
-        """Are the number of parsed energies consistent with the number of MD
-        steps?
-        """
-        self.assertEqual(self.data.scfenergies.shape, (self.nenergies, ))
-
-    def testdimatomcoords(self):
-        """Are the number of parsed geometries consistent with the number of
-        MD steps?
-        """
-        self.assertEqual(self.data.atomcoords.shape, (self.nenergies, 20, 3))
-
-    def testdimtime(self):
-        """Are the number of time points consistent with the number of MD
-        steps?
-        """
-        self.assertEqual(self.data.time.shape, (self.nsteps, ))
-
-
-class GaussianBOMDTest(GenericBOMDTest):
-    """Customized Born-Oppenheimer molecular dynamics unittest"""
-
-    # This may have something to do with our corrections for
-    # extrapolation step rejection.
-    nenergies = 35
-
 
 if __name__=="__main__":
 
--- a/test/data/test_atomcoords.py
+++ b/test/data/test_atomcoords.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2019, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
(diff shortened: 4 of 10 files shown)


langner:
And here is an example of the output when running with the new framework. Notice I haven't any customization via `TestCase` yet, which is definitely possible to make it more similar to what we have today.
```
$ python -m test.test_data_new -v
test_shape_BOMD_basicGaussian09 (test.data.test_atomcoords.AtomCoordsTest) ... ok
test_shape_BOMD_basicNWChem6.6 (test.data.test_atomcoords.AtomCoordsTest) ... ok
test_shape_BOMD_basicQChem4.2 (test.data.test_atomcoords.AtomCoordsTest) ... ok
test_shape_BOMD_basicQChem5.1 (test.data.test_atomcoords.AtomCoordsTest) ... ok
test_shape_BOMD_basicGaussian09 (test.data.test_scfenergies.ScfEnergiesTest) ... ok
test_shape_BOMD_basicNWChem6.6 (test.data.test_scfenergies.ScfEnergiesTest) ... ok
test_shape_BOMD_basicQChem4.2 (test.data.test_scfenergies.ScfEnergiesTest) ... ok
test_shape_BOMD_basicQChem5.1 (test.data.test_scfenergies.ScfEnergiesTest) ... ok
test_shape_BOMD_basicGaussian09 (test.data.test_time.TimeTest) ... ok
test_shape_BOMD_basicNWChem6.6 (test.data.test_time.TimeTest) ... ok
test_shape_BOMD_basicQChem4.2 (test.data.test_time.TimeTest) ... ok
test_shape_BOMD_basicQChem5.1 (test.data.test_time.TimeTest) ... ok

----------------------------------------------------------------------
Ran 12 tests in 0.306s

OK
```

langner:
And here is the output when run through pytest:
```
$ python -m pytest test/test_data_new.py
======================================================== test session starts ========================================================
platform linux2 -- Python 2.7.16, pytest-4.5.0, py-1.8.0, pluggy-0.11.0 -- /usr/local/google/home/langner/.virtualenvs/cclib2/bin/python
cachedir: .pytest_cache
rootdir: /usr/local/google/home/langner/tmp/cclib/test, inifile: pytest.ini
collected 12 items                                                                                                                  

test/test_data_new.py::AtomCoordsTest::test_shape_BOMD_basicGaussian09 PASSED                                                 [  8%]
test/test_data_new.py::AtomCoordsTest::test_shape_BOMD_basicNWChem6.6 PASSED                                                  [ 16%]
test/test_data_new.py::AtomCoordsTest::test_shape_BOMD_basicQChem4.2 PASSED                                                   [ 25%]
test/test_data_new.py::AtomCoordsTest::test_shape_BOMD_basicQChem5.1 PASSED                                                   [ 33%]
test/test_data_new.py::ScfEnergiesTest::test_shape_BOMD_basicGaussian09 PASSED                                                [ 41%]
test/test_data_new.py::ScfEnergiesTest::test_shape_BOMD_basicNWChem6.6 PASSED                                                 [ 50%]
test/test_data_new.py::ScfEnergiesTest::test_shape_BOMD_basicQChem4.2 PASSED                                                  [ 58%]
test/test_data_new.py::ScfEnergiesTest::test_shape_BOMD_basicQChem5.1 PASSED                                                  [ 66%]
test/test_data_new.py::TimeTest::test_shape_BOMD_basicGaussian09 PASSED                                                       [ 75%]
test/test_data_new.py::TimeTest::test_shape_BOMD_basicNWChem6.6 PASSED                                                        [ 83%]
test/test_data_new.py::TimeTest::test_shape_BOMD_basicQChem4.2 PASSED                                                         [ 91%]
test/test_data_new.py::TimeTest::test_shape_BOMD_basicQChem5.1 PASSED                                                         [100%]

===================================================== 12 passed in 0.50 seconds =====================================================
```

langner:
I'm now thinking organizing by attribute is not the best idea. But, I'm now also thinking that some of this code could be adapted to make the data unit tests easier by eliminating our custom test code in `test-data.py` - see #742 also.

berquist:
On a purely cosmetic level, it looks nice, and I like that it automatically generates individual tests which pytest can detect. I also like that the paths are now specified in Python; I was going to suggest JSON but this is even simpler.

Agreed on all your points, and I think if we can get individual regression tests to show up under pytest, the rest will be clearer.

langner:
> On a purely cosmetic level, it looks nice, and I like that it automatically generates individual tests which pytest can detect. I also like that the paths are now specified in Python; I was going to suggest JSON but this is even simpler.
> 
> Agreed on all your points, and I think if we can get individual regression tests to show up under pytest, the rest will be clearer.

Yeah, I agree that's the goal. I'll close this out once I pull out the code I want to re-use.

langner:
This is not working now anyway, closing out.

Issue #736, opened 2019-05-15
ORCA parser parse_scf_condensed_format: parsing SCF ITERATIONS table fails


schmalgo:
Hi, 

we noticed that you use split to separate the different columns in the SCF ITERATIONS. However, this won't work for this ORCA output (Version 4.1.0 - here Delta-E & Max-DP are not seperated by a whitespace). I guess one needs to extract it using sclicing :-(

```
--------------
SCF ITERATIONS
--------------
ITER       Energy         Delta-E        Max-DP      RMS-DP      [F,P]     Damp
               ***  Starting incremental Fock matrix formation  ***
  0   -257.0554667435   0.000000000000537.42184135  4.76025534  0.4401076 0.8500
  1   -258.2342173315  -1.178750588020456.80721455  4.04621310  0.3742558 0.8500
                               ***Turning on DIIS***
  2   -259.2290511175  -0.994833785987388.28592863  3.43928100  0.3183503 0.8500
  3   -260.0731915048  -0.844140387303330.04293195  2.92338837  0.2708737 0.8500
  4   -260.7921237931  -0.718932288310280.53628083  2.48487742  0.2306843 0.8500
  5   -261.4039721141  -0.611848321021238.45577989  2.11214389  0.1962770 0.8500
  6   -261.9253123193  -0.521340205231202.68746987  1.79532193  0.1668506 0.8500
  7   -262.3695926902  -0.444280370885172.28449836  1.52602429  0.1418080 0.8500
  8   -262.7481012818  -0.378508591610146.44192909  1.29712147  0.1205453 0.8500
  9   -263.0703054054  -0.322204123613124.47572947  1.10255402  0.1025067 0.8500
 10   -263.3444922447  -0.274186839222105.80446614  0.93717161  0.0871763 0.8500
 11   -263.5777979156  -0.23330567096289.93388568  0.79659644  0.0741253 0.8500
 12   -263.7763095489  -0.19851163326776.44387048  0.67710741  0.0630152 0.8500
 13   -263.9451947557  -0.16888520678364.97733415  0.57554163  0.0535673 0.8500
 14   -264.0888505553  -0.14365579965555.23076249  0.48921064  0.0455408 0.8500
 15   -264.2110301008  -0.12217954543946.94616995  0.41582923  0.0387211 0.8500
 16   -264.3149353252  -0.10390522442839.90426339  0.35345498  0.0329225 0.8500
[...]

PR #738, opened 2019-05-18
ORCA: add regression test

langner:
This is for #736. Needs to go in after cclib/cclib-data#113.


langner:
> Can you move this test and the one above it into sorted order with the other ORCA tests?

Done.

PR #743, opened 2019-05-25
Fix pytest not reporting test_data failures

berquist:
Closes #742

PR #744, opened 2019-05-26
ORCA: fix ROCIS parsing that changed from 4.0 to 4.1

berquist:
Fixes for the ORCA ROCIS issues exposed by https://github.com/cclib/cclib/pull/743. The fix for the second problem (there is a missing excitation section) is not ideal.

### DIFF
--- a/test/data/testTD.py
+++ b/test/data/testTD.py
@@ -160,11 +160,7 @@ class OrcaROCISTest(GenericTDTest):
     """Customized test for ROCIS"""
     number = 57
     expected_l_max = 2316970.8
-    # In ORCA 4.0, an additional spectrum ("COMBINED ELECTRIC DIPOLE +
-    # MAGNETIC DIPOLE + ELECTRIC QUADRUPOLE SPECTRUM (Origin Independent,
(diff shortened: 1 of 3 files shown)


test/data/testTD.py
```diff
@@ -160,7 +160,7 @@ class OrcaROCISTest(GenericTDTest):
     """Customized test for ROCIS"""
     number = 57
     expected_l_max = 2316970.8
-    n_spectra = 9
+    n_spectra = (8, 9)
```
langner: Perhaps add a comment why it's a 2-tuple with these values.

langner:
> not ideal

But it seems to work, which is the most important thing.

test/data/testTD.py
langner: Oh! This seems a little error prone, especially to one-off errors. It might be more prudent to just override the value for one ORCA version and list that in `testdata`. It won't be a lot of boilerplate, and will also be more clear.

berquist:
@langner This should be ready now.

berquist:
So master no longer appears as broken, I'm merging this.

PR #747, opened 2019-07-05
Ignore comments in empty lines in Orca parser

felixplasser:
Hi, I would just like to commit a simple fix for Orca in the case that the user puts comments or extra lines into the coordinate section.

The input and output files I used can be found here: [fafa_td.zip](https://github.com/cclib/cclib/files/3362451/fafa_td.zip). Should I commit those as well?

Cheers
-Felix

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -171,6 +171,8 @@ class ORCA(logfileparser.Logfile):
                         for line in lines_iter:
                             if not line:
                                 continue
+                            if line[0] == '#' or line.strip(' ') == '\n':
+                                continue
                             if line[0] == '*' or line.strip() == "end":
                                 break
                             # Strip basis specification that can appear after coordinates
--- a/test/regression.py
+++ b/test/regression.py
@@ -1133,6 +1133,10 @@ def testORCA_ORCA4_0_IrCl6_sp_out(logfile):
     numpy.testing.assert_almost_equal(logfile.data.scfvalues[0][0], vals_first)
     numpy.testing.assert_almost_equal(logfile.data.scfvalues[0][-1], vals_last)
 
+def testORCA_ORCA4_0_comment_or_blank_line_out(logfile):
+    """Coordinates with blank lines or comments weren't parsed correctly (#747)."""
+    assert hasattr(logfile.data,"atomcoords")
+    assert logfile.data.atomcoords.shape == (1, 8, 3)
 
 def testORCA_ORCA4_1_725_out(logfile):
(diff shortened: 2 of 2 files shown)


shivupa:
This fix works. I recommend adding the input and output files to [cclib-data](https://github.com/cclib/cclib-data) as a regression test. We could rename the files to `comment_or_blank_line_input.inp` and `comment_or_blank_line_input.out`

Here's how to do this:
1. Add the files to the `Orca/Orca4.0` folder of cclib-data
2. Add a line in `regressionfiles.txt` that corresponds to this test
3. PR

Here's how to add the regression test:
1. Add the test to `regression.py` something like this should work:
```
def testORCA_ORCA4_0_comment_or_blank_line_out(logfile):
    """Coordinates with blank lines or comments weren't parsed correctly (#747)."""
    assert hasattr(logfile.data,"atomcoords")
    assert logfile.data.atomcoords.shape == (1, 8, 3)
```
2. To test that this works locally, you can run `regression_download.sh` in the data folder but change the script to point at your fork of `cclib-data`. Then run `python -m test.regression Orca/Orca4.0/testORCA_ORCA4_0_comment_or_blank_line.out` in your cclib folder.
3. Push changes to update this PR

Let me know if you have questions!

felixplasser:
Ok, thanks. I think I will be able to do that. And while I am at it, it might make sense to add a test for my other commit to cclib: 0f8e6db4991 . I'll just add a a test that checks both things (atomcoords and gbasis) at the same time, ok?

shivupa:
That sounds reasonable. It seems the ORCA gbasis is already tested. In `test/testdata` the files `ORCA/basicORCA4.0/dvb_sp.out` and `ORCA/basicORCA4.1/dvb_sp.out` is tested as a test of type `GenericBasisTest`. This test type is defined in `test/data/testBasis.py` and makes sure that the basis is parsed correctly, but the extra check in the regression test you are adding doesn't hurt!

felixplasser:
Hi, I just did what you said. The new regression file is in [PR 115 for cclib-data](https://github.com/cclib/cclib-data/pull/115). I hope this was correct.

berquist:
Thanks Felix and Shiv, everything looks good. I want to get https://github.com/cclib/cclib/pull/744/ in first so master doesn't look broken any longer.

berquist:
@felixplasser Can you rebase this PR on top of the latest master?

felixplasser:
Ok! I pulled the newest master from cclib/cclib and pushed the merged version back to my fork. I hope that is what you meant ...

-Felix

berquist:
Yep, thanks!

PR #749, opened 2019-07-17
Molden writer can now support ghost atoms.

shivupa:
Molden writer can support ghost atoms.
A flag to ccwrite is given with which to label the ghost atoms.
Fails gracefully when ghost atoms are present, but no flag is given.
The capability can easily be extended to other writers.

This is a [WIP] because I still need to write a test, but I'd like some feedback to this approach.


cclib/io/filewriter.py
```diff
@@ -50,6 +50,8 @@ def __init__(self, ccdata, jobfilename=None, indices=None, terse=False,
         self.jobfilename = jobfilename
         self.indices = indices
         self.terse = terse
+        if "ghost" in kwargs:
+            self.ghost = kwargs["ghost"]
```
berquist: A one-liner:
```python
self.ghost = kwargs.get("ghost")
```
which has the advantage of always making sure `hasattr(my_writer, "ghost")` is true, and only `None` needs to be checked for.

cclib/io/moldenwriter.py
```diff
@@ -43,6 +43,10 @@ def _title(self, path):
     def _coords_from_ccdata(self, index):
         """Create [Atoms] section using geometry at the given index."""
         elements = [self.pt.element[Z] for Z in self.ccdata.atomnos]
+        if hasattr(self,'ghost'):
+            elements = [self.ghost if e is None else e for e in elements]
+        elif None in elements:
+            raise ValueError("It seems that there is at least one ghost atom in these elements. Please use the ghost flag to specify a label for the ghost atoms.")
```
berquist: Can you split this string onto at least two lines?

cclib/io/moldenwriter.py
```diff
@@ -43,6 +43,10 @@ def _title(self, path):
     def _coords_from_ccdata(self, index):
         """Create [Atoms] section using geometry at the given index."""
         elements = [self.pt.element[Z] for Z in self.ccdata.atomnos]
+        if hasattr(self,'ghost'):
```
berquist:
```suggestion
        if hasattr(self, 'ghost'):
```
or, if you do the one-liner, `if self.ghost is not None`.

shivupa:
Good points, thanks! Resolved. Working on a test now.

langner:
This is a great contribution, looking forward to seeing it merged in!

shivupa:
This is rebased on #752 and now contains a test. This should be good after that one goes in.

langner:
Why are there so many GAMESS-2018 logfiles in this PR?

shivupa:
Sorry it was rebased against the already merged #752

berquist:
Can you rebase again?

PR #758, opened 2019-08-23
Adding a bridge module to Psi4

schneiderfelipe:
This adds a simple bridge to the Python API of [Psi4](http://psicode.org/) [[1]](http://www.psicode.org/psi4manual/1.2/psiapi.html) [[2]](https://github.com/psi4/psi4numpy). I tried to follow the [PyQuante bridge](https://github.com/cclib/cclib/blob/428b4cf07dfd91a6b28c49c63d93ddc0175b5cf2/cclib/bridge/cclib2pyquante.py) as closely as possible.

This allows one to do things like:

```python3
import psi4
from cclib import ccopen
from cclib.bridge import makepsi4

data = ccopen("acetic_acid.out").parse()
psi4mol = makepsi4(data.atomcoords[-1], data.atomnos, data.charge, data.mult)
energy = psi4.energy("mp2/cc-pvdz", molecule=psi4mol)
print(energy)  # -228.48044415212507
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,33 @@ addons:
         - swig
 cache:
   pip: true
+  directories:
+    - $HOME/miniconda
 
 before_install:
   - sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu cosmic universe"
   - sudo apt update
   - sudo apt install libopenbabel-dev
-  - pip install -r requirements.txt
+
+  - |
+    if [[ $(echo "$TRAVIS_PYTHON_VERSION >= 3.6" | bc -l) == 1 ]];
+    then
+        wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
+        bash miniconda.sh -b -p $HOME/miniconda -u
+        export PATH="$HOME/miniconda/bin:$PATH"
+
+        hash -r
+        conda config --set always_yes yes --set changeps1 no
+        conda update -q conda
+        conda info -a
+        conda create -q -n p4env python=$TRAVIS_PYTHON_VERSION ci-psi4 psi4 numpy matplotlib jupyter scipy -c psi4/label/dev -c defaults -c conda-forge
+
+        source activate p4env
+        conda install pytest pytest-cov codecov -c conda-forge
+        pip install pytest-shutil
+        conda list
+    fi
+    pip install -r requirements.txt
 
 install:
   - pip install .
@@ -36,7 +57,7 @@ script:
 
 after_success:
   - |
-    if [[ "${TRAVIS_BRANCH}" == master && "${TRAVIS_PULL_REQUEST}" == false && $TRAVIS_PYTHON_VERSION == 3.7 ]];
+    if [[ "${TRAVIS_BRANCH}" == master && "${TRAVIS_PULL_REQUEST}" == false && $(echo "$TRAVIS_PYTHON_VERSION == 3.7" | bc -l) == 1 ]];
     then
         # Commits to master that are not pull requests, that is, only actual
         # addition of code to master, should deploy the documentation.
--- a/test/test_bridge.py
+++ b/test/test_bridge.py
@@ -12,9 +12,13 @@ import unittest
 
 sys.path.insert(1, "bridge")
 
-if sys.version_info[0] == 3 and sys.version_info[1] >= 4:
-    from .bridge.testbiopython import *
+if sys.version_info[0] == 3:
+    if sys.version_info[1] >= 6:
+        from .bridge.testpsi4 import *
+    if sys.version_info[1] >= 4:
+        from .bridge.testbiopython import *
 from .bridge.testopenbabel import *
+
 if sys.version_info[0] == 2:
     from .bridge.testpyquante import *
 
--- a/cclib/bridge/__init__.py
+++ b/cclib/bridge/__init__.py
@@ -18,5 +18,8 @@ if find_package("openbabel"):
 if find_package("PyQuante"):
     from cclib.bridge.cclib2pyquante import makepyquante
 
+if find_package("psi4"):
+    from cclib.bridge.cclib2psi4 import makepsi4
+
 
 del find_package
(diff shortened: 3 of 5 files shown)


shivupa:
Could you rebase this on master please.

schneiderfelipe:
> Could you rebase this on master please.

Let me know if it's OK now.

langner:
This is pretty cool! @berquist any thoughts about this?

However, before merging in I would prefer to get Travis cleaned up.

schneiderfelipe:
> This is pretty cool! @berquist any thoughts about this?
> 
> However, before merging in I would prefer to get Travis cleaned up.

I agree. I think I might need help on this since this would mean compiling psi4 on Travis. Or am I missing something?

test/test_bridge.py
```diff
@@ -12,9 +12,13 @@
 
 sys.path.insert(1, "bridge")
 
-if sys.version_info[0] == 3 and sys.version_info[1] >= 4:
-    from .bridge.testbiopython import *
+if sys.version_info[0] == 3:
+    if sys.version_info[1] >= 6:
+        from .bridge.testpsi4 import *
+    elif sys.version_info[1] >= 4:
```
berquist:
```suggestion
    if sys.version_info[1] >= 4:
```

berquist:
And yes, it's great to have this as a bridge.

test/test_bridge.py
schneiderfelipe: Well, I missed that, thanks! :)

.travis.yml
```diff
@@ -14,7 +14,25 @@ before_install:
...
   - sudo apt install libopenbabel-dev
+
+  - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda
+  - conda info -a
+  - if [[ "$PYTHON_VER" == "3.6" ]]; then
```
berquist: Can you try removing this branch (which just pins a specific version of NumPy)? It's actually the other one that runs for 2.7, which seems to work fine, so it's probably unnecessary.

.travis.yml
```diff
@@ -14,7 +14,25 @@ before_install:
...
+        conda create -q -n p4env python=$PYTHON_VER ci-psi4 psi4 numpy=1.13 matplotlib jupyter scipy -c psi4;
+    else
+        conda create -q -n p4env python=$PYTHON_VER ci-psi4 psi4 numpy matplotlib jupyter scipy -c psi4/label/dev -c defaults -c conda-forge;
```
berquist:
```suggestion
        conda create -q -n p4env python=$TRAVIS_PYTHON_VERSION ci-psi4 psi4 numpy matplotlib jupyter scipy -c psi4/label/dev -c defaults -c conda-forge;
```
should be the environment variable.

berquist:
Also, I think we want to add the miniconda directory to the Travis cache.

langner:
> Psi4 is printing the test calculation to standard out (grep for `Loading Basis Set` in https://api.travis-ci.org/v3/job/583138897/log.txt). Normally this is fine, but I've disabled having pytest capturing stdout, so printing small stuff to the log is ok, but this is really big. We should either
> 
> 1. Keep printing Psi4 to stdout, but allow pytest to capture stdout by default
> 2. Have Psi4 print to stderrr or a file, and keep disallowing pytest to capture stdout so we can print to it for easier debugging.
> 
> I don't have an opinion right now. @langner what do you think?

Yeah, it is a little much when I look at Travis. I'd like to eliminate more of what's in there, so it's easier to debugf failures.

I don't have a big preference. If we go with the momentum, which seems like a reasonable default, we should redirect Psi4's output. I'm also OK with merging this and cleaning up that bit later.

.travis.yml
schneiderfelipe: Done.

.travis.yml
schneiderfelipe: Done.

schneiderfelipe:
> Also, I think we want to add the miniconda directory to the Travis cache.

Sure. I added

```
...
cache:
  directories:
    - $HOME/miniconda
...
```

Is this OK?

berquist:
Conda is failing to solve for dependencies on Python 3.4. I think this is fundamentally due to Psi4 not working with such an old Python version, and it's debatable if it should work for 2.7.

I think the solution is to put all conda things in `.travis.yml` back behind a conditional block that looks for `TRAVIS_PYTHON_VERSION -gt 3.6`, which looks like [their minimum tested version](https://github.com/psi4/psi4/blob/aa2bb2859175ca9b0c1264834e164a588d55074b/.travis.yml).

berquist:
The noisy output is coming from the energy calculation, not creating the Molecule, which means it can just be silenced in the test using `psi4.core.set_output_file("output.dat", False)`.

schneiderfelipe:
It looks like everything is working fine, except in Python 3.7. Psi4 is not being built there and I don't know why. Can somebody help me with this?

berquist:
```
/home/travis/.travis/functions: line 109: [[: 3.7: syntax error: invalid arithmetic operator (error token is ".7")
```
because `bash` can't do floating-point arithmetic, apparently. This should work:
```bash
[[ $(echo "$TRAVIS_PYTHON_VERSION >= 3.6" | bc -l) == 1 ]]
```

.travis.yml
```diff
@@ -36,7 +57,7 @@ script:
 
 after_success:
   - |
-    if [[ "${TRAVIS_BRANCH}" == master && "${TRAVIS_PULL_REQUEST}" == false && $TRAVIS_PYTHON_VERSION == 3.7 ]];
+    if [[ "${TRAVIS_BRANCH}" == master && "${TRAVIS_PULL_REQUEST}" == false && $(echo "$TRAVIS_PYTHON_VERSION >= 3.6" | bc -l) == 1 ]];
```
berquist: I think we want to keep this a check specifically for 3.7, otherwise once we have 3.6 and 3.8 (for example) each CI run will try and push the docs. It isn't that we need 3.7 specifically, just that we only want one version to push.

PR #759, opened 2019-08-27
Make makecclib the inverse of makeopenbabel

schneiderfelipe:
It's currently impossible to call `makecclib` from the result of `makeopenbabel`, since `makeopenbabel` demands the dimension of `atomcoords` to be three, while the return object of `makecclib` is always two dimensional.

This PR ensures `np.asarray(atomcoords).ndim == 3` for the return object of ` cclib.bridge.cclib2openbabel.makecclib`.


shivupa:
Could you rebase this on master please.

schneiderfelipe:
> Could you rebase this on master please.

Let me know if it's OK now.

langner:
> > Could you rebase this on master please.
> 
> Let me know if it's OK now.

Look like the history is somehow mangled (you rebasing modified the commits in the history).

shivupa:
You should be able to fix this by checking out your branch and running `git rebase -i master`, all that should show up is [this commit](https://github.com/cclib/cclib/commit/9daee8f2e4ef8872297becf5fe0200c4aab6dc3f)

cclib/bridge/cclib2openbabel.py
```diff
@@ -38,6 +38,7 @@ def makecclib(mol):
         attributes['atomcoords'].append([atom.GetX(), atom.GetY(), atom.GetZ()])
         attributes['atommasses'].append(atom.GetAtomicMass())
         attributes['atomnos'].append(atom.GetAtomicNum())
+        attributes['atomcoords'] = [attributes['atomcoords']]
```
berquist: This should actually be indented to the same level as the return statement, otherwise you'll create an extra level of nesting every time an atom is read.

berquist:
Poke for rebase. I think this is otherwise good, but can't tell because of all the other changes Git sees as present.

schneiderfelipe:
I'm closing this in favor of #811 (I lost the repository this was based on :sob:).

PR #760, opened 2019-08-31
Update travisyml to install newer openbabel from apt

shivupa:


shivupa:
Feels good to finally see the green check. Shout out to @berquist for the help.

langner:
Thanks for this!

PR #762, opened 2019-09-01
Fixes for moved GAMESS-US 2014 unit tests

shivupa:
The molden tests needed to be rewritten to include the Gamess 2018 data since 2014 was removed.


cclib/io/moldenwriter.py
```diff
@@ -244,6 +244,8 @@ def reformat(self):
 
             # Check for [Title] section.
             if '[title]' in line.lower():
+                # skip the title
+                line = next(filelines)
```
berquist: Can you use `self.skip_lines` here?

shivupa: So I could for the line that actually says [title], but we need to skip the line after it which is the real title and that could be anything. I was looking at [this](https://github.com/cclib/cclib/blob/5605bdd98085a7e41dee4965bf0573aeefe89b2e/cclib/parser/logfileparser.py) and it looks like I have to give it some pattern to skip right?

berquist: We could stand to update the docs here, because it also confused me for the longest time. Here is what it actually does. Let's say I write
```python
self.skip_lines(inputfile, ["d", "b", "Excitation energy in a.u."])
line = next(inputfile)
```
where `d` is for a line that contains only dashes (and possibly spaces, but excluding the newline) and `b` is for a "blank" line that contains zero or more spaces (excluding the newline). For dashes, stars, equals signs, or a blank line, we do a check to make sure that the line is exactly that. Even if it is not, the line is still skipped after the warning is printed. For anything else, like `"Excitation energy in a.u."` or `"authors"` or `"State   Energy Wavelength"`, it's a comment to the person reading the code about what that line contains, and no check is performed on the line before it's skipped. TL;DR it's just a comment meant for us, not a pattern to be matched.

langner: This is correct. We should update the docs, and maybe make the the syntax more explicit.

langner: Created #765 for this.

shivupa: Ok I think I understand how this work, but `Molden_reformatter` isn't of type `cclib.parser.logfileparser.Logfile` or any derived class. I'm still confused about how to use skip_lines outside of such a class?

langner: Ah, you're right. Scratch that then.

shivupa:
This should be ready to go. I think we should retrigger a build. Since the corresponding cclib-data PR went in I think this should pass all tests.

berquist:
The Travis failure
```
python coverage.py _build/cclib_dev > coverage_dev.rst
  File "coverage.py", line 55, in generate_coverage
    suite.testall()
  File "/home/travis/build/cclib/cclib/doc/sphinx/_build/cclib_dev/test/test_data.py", line 182, in testall
    datatype=test.datatype if hasattr(test, 'datatype') else None
  File "/home/travis/build/cclib/cclib/doc/sphinx/_build/cclib_dev/test/test_data.py", line 114, in getdatafile
    data = logfile.parse()
  File "/home/travis/build/cclib/cclib/doc/sphinx/cclib/parser/logfileparser.py", line 301, in parse
    inputfile = openlogfile(self.filename)
  File "/home/travis/build/cclib/cclib/doc/sphinx/cclib/parser/logfileparser.py", line 166, in openlogfile
    else FileWrapper(io.open(filename, "r", errors='ignore'))
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/cclib/cclib/doc/sphinx/_build/cclib_dev/data/GAMESS/basicGAMESS-US2014/dvb_sp.out'
```
is misleading, because the unit and regression tests pass for this branch, but when trying to build the development version of the docs, our Makefile always builds from master, which is currently broken. I've made issue https://github.com/cclib/cclib/issues/766 to fix this. Since everything else is working, I'm merging this. Thanks for fixing this Shiv.

PR #768, opened 2019-09-08
Update changelist and bump version to v1.6.2

langner:


berquist:
Travis is failing because it's trying to build docs for the tag `v1.6.2`, which doesn't exist yet.

langner:
I will create the tag shortly.

PR #769, opened 2019-09-13
Enable metadata tests

berquist:
as mentioned in https://github.com/cclib/cclib/pull/732#issuecomment-530198350.

I also found a few other skip decorators in the single-point tests that could be removed.

### DIFF
--- a/test/data/testSP.py
+++ b/test/data/testSP.py
@@ -293,20 +293,11 @@ class GenericSPTest(unittest.TestCase):
             for m in moment32:
                 self.assertEqual(m, 0.0)
 
-    @skipForParser('ADF', 'Does not support metadata yet')
-    @skipForParser('GAMESSUK', 'Does not support metadata yet')
-    @skipForParser('Molcas', 'The parser is still being developed so we skip this test')
-    @skipForParser('Molpro', 'Does not support metadata yet')
-    @skipForParser('NWChem', 'Does not support metadata yet')
-    @skipForParser('ORCA', 'Does not support metadata yet')
-    @skipForParser('Psi3', 'Does not support metadata yet')
-    @skipForParser('Psi4', 'Does not support metadata yet')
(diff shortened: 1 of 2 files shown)


test/data/testSP.py
```diff
@@ -293,20 +290,11 @@ def testmoments(self):
...
-    @skipForParser('GAMESSUK', 'Does not support metadata yet')
-    @skipForParser('Molcas', 'The parser is still being developed so we skip this test')
-    @skipForParser('Molpro', 'Does not support metadata yet')
-    @skipForParser('NWChem', 'Does not support metadata yet')
-    @skipForParser('ORCA', 'Does not support metadata yet')
-    @skipForParser('Psi3', 'Does not support metadata yet')
-    @skipForParser('Psi4', 'Does not support metadata yet')
-    @skipForParser('QChem', 'Does not support metadata yet')
-    @skipForParser('Turbomole', 'The parser is still being developed so we skip this test')
     def testmetadata(self):
```
langner: Do you think it makes sense to break this down into one method for each metadata attribute? Do we want to validate the value of any of the metadata?

berquist: > Do you think it makes sense to break this down into one method for each metadata attribute?

Yes.

> Do we want to validate the value of any of the metadata?

Yes, but related to https://github.com/cclib/cclib/pull/599#pullrequestreview-149546363 and https://github.com/cclib/cclib/issues/602, testing parser-specific results is messy right now, unless we restrict it to `SP` tests.

langner: I think it's fine to limit to `SP` test for now in that case. Maybe sprinkle with TODOs so we now where to go from here. Anyway, we can do this incrementally.

langner:
@berquist is this ready for v1.6.3?

berquist:
@langner Yes.

langner:
OK, here we go!

PR #770, opened 2019-09-14
Make Logfile base class abstract

berquist:
The cleanest way of enforcing that `normalisesym` is implemented for each parser is to make it an abstract method on `Logfile`. It also removes the need for testing to make sure it's implemented.

- We are no longer using doctests, so all references to using them were removed.
- Implements `normalisesym` for Molcas, Psi4, and Turbomole.

### DIFF
--- a/cclib/parser/logfileparser.py
+++ b/cclib/parser/logfileparser.py
@@ -17,6 +17,8 @@ import os
 import random
 import sys
 import zipfile
+from abc import ABCMeta, abstractmethod
+from six import add_metaclass
 
 if sys.version_info.major == 2:
     getargspec = inspect.getargspec
@@ -182,6 +184,7 @@ def openlogfile(filename, object=None):
         return fileobject
 
 
+@add_metaclass(ABCMeta)
 class Logfile(object):
     """Abstract class for logfile objects.
 
@@ -398,14 +401,9 @@ class Logfile(object):
                 self.progress.update(newstep, msg)
                 self.progress.step = newstep
 
+    @abstractmethod
     def normalisesym(self, symlabel):
-        """Standardise the symmetry labels between parsers.
-
-        This method should be overwritten by individual parsers, and should
-        contain appropriate doctests. If is not overwritten, this is detected
-        as an error by unit tests.
-        """
-        raise NotImplementedError("normalisesym(self, symlabel) must be overriden by the parser.")
+        """Standardise the symmetry labels between parsers."""
 
     def float(self, number):
         """Convert a string to a float.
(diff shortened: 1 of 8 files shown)


cclib/parser/logfileparser.py
```diff
@@ -398,16 +401,12 @@ def updateprogress(self, inputfile, msg, xupdate=0.05):
...
 
-        This method should be overwritten by individual parsers, and should
-        contain appropriate doctests. If is not overwritten, this is detected
-        as an error by unit tests.
-        """
-        raise NotImplementedError("normalisesym(self, symlabel) must be overriden by the parser.")
-
-    def float(self, number):
+    @staticmethod
+    def float(number):
```
langner: If this is not parser-specific then maybe move out of the class? Into `utils`?

cclib/parser/logfileparser.py
```diff
@@ -182,6 +184,7 @@ def openlogfile(filename, object=None):
         return fileobject
 
 
+@add_metaclass(ABCMeta)
```
langner: Can we use the standard `metaclass` syntax (https://docs.python.org/3/library/abc.html)?

cclib/parser/molcasparser.py
```diff
@@ -34,9 +34,13 @@ def __repr__(self):
         """Return a representation of the object."""
         return 'Molcas("%s")' % (self.filename)
 
-    #These are yet to be implemented.
     def normalisesym(self, label):
-        """Does Molcas require symmetry label normalization?"""
+        """Normalise the symmetries used by Molcas.
+
+        The labels are standardized except for the first character being
+        lowercase.
```
langner: I don't think we are being strict about the 80 column rule, especially for comments and docstrings.

cclib/parser/psi4parser.py
```diff
@@ -50,8 +50,11 @@ def after_parsing(self):
                 self.set_attribute('natom', len(self.atomnos))
 
     def normalisesym(self, label):
-        """Psi4 does not require normalizing symmetry labels."""
-        return label
+        """Psi4 does not require normalizing symmetry labels.
+
+        Only the Cs group (A prime and A double prime) need modification.
```
langner: It's not clear how the `pp` and `p` replacements below relate to Cs group.

cclib/parser/logfileparser.py
berquist: It's not parser-specific, but it's going to change ~80 lines in the parsers from `self.float(...)` to `utils.float(...)`. Are you ok with that?

cclib/parser/logfileparser.py
berquist: As long as we're supporting Python 2, we can't: https://stackoverflow.com/a/18513858/

cclib/parser/logfileparser.py
langner: I am - I think we do a lot of that for other functions. But maybe in a separate PR, then.

Issue #604, opened 2018-08-26
Replace parser.utils.PeriodicTable with periodictable


berquist:
As discussed in https://github.com/cclib/cclib/pull/568, our built-in periodic table should be replaced with [periodictable](https://pypi.org/project/periodictable/).

Rather than rewrite everywhere that our own table is used, PeriodicTable should be modified to call periodictable, so the interface doesn't need to be changed.

langner:
Another option that looks reasonable: https://github.com/MolSSI/QCElemental

berquist:
It does; it could also replace our unit converter. Are you ok with its dependency on Pint?

dudektria:
Hi, [mendeleev](https://mendeleev.readthedocs.io/en/stable/) might be also an alternative.

pandaxtc:
Hi, I'd like to take this issue! Is that alright?

berquist:
@pandaxtc Go right ahead! My thought is that we probably want to use QCElemental rather than `periodictable`, so the issue title is maybe misleading.

pandaxtc:
It looks like QCElemental is only Python 3 compatible, it's listed as such on PyPi and some of its dependencies are Python 3 only. Should I use `periodictable` instead, or should this be put on hold for now?

langner:
A good opportunity to drop Python2 support? :)

pandaxtc:
I've gone ahead and made a pull request using `periodictable` just for now!

berquist:
I am ok with depending on `periodictable` rather than QCElemental for now. The benefit would be that we could drop SciPy as a depedency, since QCElemental comes with the physical constants we need for the nuclear methods.

berquist:
I've made https://github.com/cclib/cclib/issues/776 so we can track a possible move to QCElemental there.

PR #774, opened 2019-10-05
Change PeriodicTable class to use periodictable library (#604)

pandaxtc:
Change parser.utils.PeriodicTable class to use periodictable library, as per #604. The interface for the class remains unchanged.

### DIFF
--- a/cclib/parser/utils.py
+++ b/cclib/parser/utils.py
@@ -164,8 +164,9 @@ class PeriodicTable(object):
         self.number = {}
         
         for e in periodictable.elements:
-            self.element.append(e.symbol)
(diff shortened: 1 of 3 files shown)


pandaxtc:
@berquist added!

berquist:
Are we testing that the zeroth element is `None`?

pandaxtc:
That was just the behavior of the previous implementation, I can add a test for this in `testutils.py` if you want.

berquist:
Yeah, just so that it makes noise if we ever decide to change the implementation. It can be an `assertNone` under `test_elements`.

berquist:
Thank you!

PR #775, opened 2019-10-06
Fix duplicate method name in testutils.py, fixes #773

### LINKED ISSUE #773: Duplicate test method name in testutils.py
In test/parser/testutils.py, there's two methods defined with the same name (test_elements): https://github.com/cclib/cclib/blob/173ee7091104a66506009c82f73b1b61991d174e/test/parser/testutils.py#L44-L58

Because of this, only the second test_elements actually runs. Should this be fixed?

pandaxtc:
Rename duplicate test method `test_elements` to `test_numbers`, because it tests atomic numbers in PeriodicTable.

Fixes #773


berquist:
Thank you!

Issue #766, opened 2019-09-07
Development docs always build from master, not the current checked-out branch


berquist:
as discovered in https://github.com/cclib/cclib/pull/762.

langner:
We might want to prioritize this for GSOC2020.

berquist:
@langner Do you want to take it, or should someone else?

langner:
> @langner Do you want to take it, or should someone else?

I can, although I don't know where to start. So if you have an idea, feel free to take the lead.

berquist:
Done in #1309

PR #781, opened 2019-10-26
Replace periodictable and SciPy with QCElemental

### LINKED ISSUE #776: Replace periodictable and SciPy with QCElemental
As discussed in https://github.com/cclib/cclib/issues/604, 

This would radically change library compatibility:
- Only Python >= 3.6 would be supported. Python 2 support would be dropped, which is perhaps ok, but 3.5 isn't EOL until September 2020 (https://devguide.python.org/#status-of-python-branches).
- Since Python 2 would be dropped, we can also drop `six`.
- Other old code can be dropped.
- It would bring in transitive dependencies on [`pint`](https://github.com/hgrecco/pint) and [`pydantic`](https://github.com/samuelcolvin/pydantic), neither of which have dependencies of their own. These are not as mature and stable as SciPy, though we wouldn't depend directly on their features. The physical constants might come with units attached, which could be nice.

I've set the milestone to 2.x, not because we want to wait a long time to do this, but because it's a breaking change in terms of Python support. We could finally use f-strings and type hints.

berquist:
Closes #776

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 dist: focal
 language: python
 python:
-  - 2.7
-  - 3.5
+  - 3.6
   - 3.7
 addons:
     apt:
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -15,9 +15,6 @@ import sys
 import re
 from tempfile import NamedTemporaryFile
 
-# We want this as long as we need to support both Python 2 and 3.
-from six import string_types
-
 # Python 2->3 changes the default file object hierarchy.
 if sys.version_info[0] == 2:
     fileclass = file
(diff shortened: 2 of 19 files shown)


berquist:
Looks like tests run fine but the docs builds are failing, need to temporarily reinstall periodictable.

shivupa:
The docs fail because of #766 right?

berquist:
@shivupa Yes, but more specifically, the install dependencies are changing, which I wasn't considering in #766. To do that properly, once would need an additional isolated environment, probably for building the previous release.

berquist:
Superseded by https://github.com/cclib/cclib/pull/972. In another PR, QCElemental will be used to replace periodictable, but we will keep SciPy as a hard dependency.

PR #783, opened 2019-11-22
adding parser for gaussian mulliken atom spins

pstjohn:
Adds a couple lines to enable parsing of mulliken atom spins from a gaussian log file. These look like this
```
 Mulliken charges and spin densities:
               1          2
     1  C   -0.387596   1.060148
     2  C   -0.200858  -0.093393
     3  C   -0.336819   0.029550
     4  C    0.349212  -0.000550
     5  O   -0.322174   0.002229
     6  O   -0.410581  -0.000111
     7  H    0.163961  -0.038402
     8  H    0.163228  -0.036199
     9  H    0.151451   0.022440
    10  H    0.157424   0.058876
    11  H    0.166521  -0.001892
    12  H    0.171347  -0.003006
    13  H    0.334883   0.000311
 Sum of Mulliken charges =  -0.00000   1.00000
```

### DIFF
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -1762,19 +1762,29 @@ class Gaussian(logfileparser.Logfile):
            line[1:23] == "Lowdin Atomic Charges:" or line[1:16] == "Lowdin charges:" or \
            line[1:37] == "Mulliken charges and spin densities:":
 
+            has_spin = 'spin densities' in line 
+
             if not hasattr(self, "atomcharges"):
                 self.atomcharges = {}
+
+            if has_spin and not hasattr(self, "atomspins"):
+                self.atomspins = {}
 
             ones = next(inputfile)
 
             charges = []
+            spins = []
             nline = next(inputfile)
             while not "Sum of" in nline:
                 charges.append(float(nline.split()[2]))
+                if has_spin:
+                    spins.append(float(nline.split()[3]))
                 nline = next(inputfile)
 
             if "Mulliken" in line:
                 self.atomcharges["mulliken"] = charges
+                self.atomspins["mulliken"] = spins
+
             else:
                 self.atomcharges["lowdin"] = charges
 
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -1783,7 +1783,8 @@ class Gaussian(logfileparser.Logfile):
 
             if "Mulliken" in line:
                 self.atomcharges["mulliken"] = charges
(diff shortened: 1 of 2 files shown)


berquist:
Thanks for the contribution Peter! The CI failure is unrelated (a dependency no longer supports one of the tested Python versions because we don't pin `requirements.txt` versions). A temporary workaround is to have `biopython==1.74` in `requirements.txt`.

It would also be good to have an `atomspins` test under https://github.com/cclib/cclib/blob/2f46d88f2f22966d75448809e869123396ad140c/test/data/testSPun.py#L179.

pstjohn:
Great! I figured the CI test wasn't due to the changes I made. Let me know if you want me to make the requirements.txt change, or if its easier to keep that decoupled from this PR

berquist:
Can you rebase this PR on top of the current master?

berquist:
This is failing because in some of the regression files, the header is
```
 Mulliken atomic spin densities:
```
so you will need to add another condition to the header for matching the section.

pstjohn:
Which regression files would those be? The gaussian parser doesn't appear to handle that case, but I want to know what the spin densities look like with that header

berquist:
Here is one example: https://github.com/cclib/cclib-data/blob/32708ce638ad34618aa454d1b3c9ad942119c11d/Gaussian/Gaussian09/dvb_un_sp_b_revA.02.log#L6853

pstjohn:
the test error doesn't seem to be in the code I changed:
```
>       assert len(logfile.data.entropy) == -numpy.inf
E       TypeError: object of type 'float' has no len()
test/regression.py:1651: TypeError
```

schneiderfelipe:
> the test error doesn't seem to be in the code I changed:
> 
> ```
> >       assert len(logfile.data.entropy) == -numpy.inf
> E       TypeError: object of type 'float' has no len()
> test/regression.py:1651: TypeError
> ```

This is due to #816 and is solved by #817.

pstjohn:
Are these regression failures from my changes?

langner:
> Are these regression failures from my changes?

It's hard to tell, because we had some failures on HEAD recently. I just restarted the Travis build.

langner:
> > Are these regression failures from my changes?
> 
> It's hard to tell, because we had some failures on HEAD recently. I just restarted the Travis build.

Looks like the failure is from your change.

pstjohn:
odd, not sure why that fixed it  🤷‍♂ 
But this LGTM from my end, let me know if there's anything else you'd like to see

langner:
I think this is good to go. Thanks again!

PR #784, opened 2019-11-25
Making parsing of free energy work for 4.2.0

shijunang:


berquist:
Thanks for the contribution! The CI failure is unrelated (a dependency no longer supports one of the tested Python versions because we don't pin `requirements.txt` versions). A temporary workaround is to have `biopython==1.74` in `requirements.txt`.

Can I reproduce this by simply running our frequency unit test with ORCA 4.2.0?

shivupa:
@berquist should this be rebased on master or is it ok? This is indeed an issue [4.1](https://github.com/shivupa/cclib/blob/orca4p2tests/data/ORCA/basicORCA4.1/dvb_ir.out) and [4.2](https://github.com/shivupa/cclib/blob/orca4p2tests/data/ORCA/basicORCA4.2/dvb_ir.out)

berquist:
@shijunang Can you rebase this PR on top of our current master?

Issue #782, opened 2019-11-09
Switch from `pip` to `conda` in Travis


berquist:
Because of https://github.com/openbabel/openbabel/issues/2067, we can no longer use `pip` to install Open Babel. However, thanks to Felipe's hard work in https://github.com/cclib/cclib/pull/758, it should be straightforward to make the full switch over.

Personally, I do not like `conda` at all, one reason being its slowness, especially on Travis (total time for a single build went up to 12 and a half minutes from 7 and a half on 3.7), but we don't have much choice unless we'd like to build just the Python bindings form scratch, and we're already using it for Psi4 anyway.

berquist:
Oh, and this is in order to upgrade from Open Babel 2.4.1 to 3.0.0, which...do we want to?

langner:
You mentioned in the other bug we're already using Conda... for what?

I personally don't mind much what we use on Travis, since this is just for testing, although I'd tend towards to lighter option (which means pip here). FTR though, I would like to continue supporting cclib in both pip and conda (although I understand that's not the issue at hand here).

berquist:
> You mentioned in the other bug we're already using Conda... for what?

Psi4; the alternative is to download a tarball with a pre-built version, but we don't have anywhere to host it. Building it from source every time is worse, even with caching.

For Open Babel 3.0, there are breaking Python API changes. It doesn't look like even the latest Ubuntu distro [has it available](https://packages.ubuntu.com/search?keywords=openbabel&searchon=names&suite=all&section=all) so changing the PPA wouldn't work to get the compiled library. `conda` might just be simpler here...

> FTR though, I would like to continue supporting cclib in both pip and conda (although I understand that's not the issue at hand here).

At the cost of some CI complexity, maybe we can have some runs that test all the functionality (like bridges) with `conda` and have other runs that test core functionality but installed with `pip`? We would use a matrix in `.travis.yml` rather than just Python versions, and need to introduce some conditional logic in our tests to skip them when dependencies aren't installed.

langner:
Ok then, let's do it.

berquist:
I think this was done in some combination of #758, #881, #939, though in between #939 and switching to GitHub Actions in #983 we never fully switched.

PR #785, opened 2019-11-25
Bump travis python version so biopython can install

shivupa:
Travis was failing on 3.4 because biopython needs 3.5 or later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 python:
   - 2.7
-  - 3.4
+  - 3.5
   - 3.7
 addons:
     apt:
--- a/test/parser/testlogfileparser.py
+++ b/test/parser/testlogfileparser.py
@@ -105,7 +105,10 @@ class LogfileTest(unittest.TestCase):
 
         parser.etenergies = [1, -1]
         parser.parse()
-        parser.logger.error.assert_called_once()
+        try:
+            parser.logger.error.assert_called_once()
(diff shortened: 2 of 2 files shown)


berquist:
I don't have the slightest idea what the problem with `mock` is, but can dig into it if you need me to.

shivupa:
Thanks! I'm planning to give this a look tonight, but I'll give you a shout if I get stuck.

shivupa:
Update `mock` doesn't have `assert_called_once`[ until 3.6](https://docs.python.org/3.6/library/unittest.mock.html) but I am confused why this didn't error before  on 3.4 since the line it has a problem with [has been there for ten months and we have been testing with 3.4 for a while....](https://github.com/cclib/cclib/blame/2f46d88f2f22966d75448809e869123396ad140c/test/parser/testlogfileparser.py#L108)

shivupa:
Let me know if this looks ok @berquist

berquist:
Just checked and 3.4 is [end-of-life](https://devguide.python.org/devcycle/#end-of-life-branches). 3.5 isn't EOL until this coming [September](https://devguide.python.org/#branchstatus). Thoughts:
- When I handle #782 I may reintroduce testing for 3.4 if all it means is installing an older BioPython version for testing.
- 3.5 is the first version to support inline type annotations, which I [started on](https://github.com/cclib/cclib/compare/master...berquist:mypy) not long ago.

PR #787, opened 2019-12-10
Fix for OB import change

shivupa:
The openbabel python interface changed.


berquist:
I added some comments about why the change was necessary, and why we need to use try/except instead of a version check. Thank you!

PR #788, opened 2019-12-12
Parsing ORCA MP2 gradients

alexmmaldonado:
The ORCA MP2 module uses a different flag other than "CARTESIAN GRADIENT" for gradients. Here, I just added another block to the ORCA parser that catches these MP2 gradients.


cclib/parser/orcaparser.py
```diff
@@ -403,6 +403,25 @@ def splitter(line):
             if not hasattr(self, 'grads'):
                 self.grads = []
             self.grads.append(grads)
+        
+        # Grab MP2 gradients
```
shivupa: This involves a little bit of code duplication. We could fix that by doing the following.

The `next` statements at 392 and 393 aren't rewriting `line` so at 394 we could add `if line[:18] == 'CARTESIAN GRADIENT' or line[:22] == 'The final MP2 gradient':` 

Then lines 413-424 could be removed.

Remember to take the comment here on 407 the comment and combine it with the comment on 390

Let me know if this makes sense

shivupa:
Thanks for this! I had a minor change. Also we don't currently have a test that tests this. Do you have an example that we could turn into a test? The way this would work would be to add the file to the relevant directory [here](https://github.com/cclib/cclib-data) and add a test in [here](https://github.com/cclib/cclib/blob/master/test/regression.py) that tests that the gradient is parsed correctly. Let me know if this makes sense.

cclib/parser/orcaparser.py
alexmmaldonado: Yes, that makes sense for the most part.

I combined the two signals for cartesian gradient and MP2 gradient and merged the common code. Let me know if this is what you were looking for, or what else you would like me to correct.

cclib/parser/orcaparser.py
```diff
@@ -388,22 +388,39 @@ def splitter(line):
...
 
             grads = []
-            line = next(inputfile).strip()
-            while line:
-                idx, atom, colon, x, y, z = line.split()
-                grads.append((float(x), float(y), float(z)))
 
+            if line[:18] == 'CARTESIAN GRADIENT':
+                next(inputfile)
+                next(inputfile)
```
berquist: I understand that the original code had these lines, but can you convert them to use `skip_lines`?

cclib/parser/orcaparser.py
```diff
@@ -388,22 +388,39 @@ def splitter(line):
...
+
+                line = next(inputfile).strip()
+                while line:
+                    idx, atom, colon, x, y, z = line.split()
```
berquist: You could do something like
```python
tokens = line.split()
x, y, z = float(tokens[-3]), float(tokens[-2]), float(tokens[-1])
```
and then both `while` blocks contain the same code, and you can combine the two `if line[...` blocks.

cclib/parser/orcaparser.py
alexmmaldonado: Not a problem.

cclib/parser/orcaparser.py
alexmmaldonado: Yes, that works great.

alexmmaldonado:
I added, to the best of my knowledge, the regression test to the output file I added to cclib-data. Let me know if there is something I am missing.

PR #790, opened 2020-01-09
Fixing problem with units on moment of inertia tensor

jamesETsmith:
Hi all, I ran across an issue with the function `Nuclear.principal_moments_of_inertia()`. I found that the units for this function were not correct.

Here is a script which demonstrates the original problem:
```python
import cclib
from cclib.method import Nuclear

file_path = (
    "/home/james/apps/pull_requests/cclib/data/Gaussian/basicGaussian16/dvb_ir.out"
)
data = cclib.io.ccread(file_path)
nuc = Nuclear(data)
print(nuc.principal_moments_of_inertia()[0])
```

Output:
```
[109.23249971 737.88033329 847.112833  ]
```

Comparing these to the Gaussian16 output moments (shown below) I believe the moments from cclib are incorrect.
```
 Principal axes and moments of inertia in atomic units:
                           1         2         3
     Eigenvalues --   390.07631 2635.01852 3025.09483
           X            0.02413   0.99971   0.00000
           Y            0.99971  -0.02413   0.00000
           Z           -0.00000  -0.00000   1.00000
```

Since `Nuclear.moment_of_inertia_tensor()` is used elsewhere (e.g. `Nuclear.rotational_constants()`) and behaves correctly, I think that the problem is isolated in the unit conversions in `Nuclear.principal_moments_of_inertia()`. After correcting these I get the following output from my script above:
```
[ 390.07633792 2635.01850639 3025.09484431]
````

I've run the unit tests and found that it breaks the `NuclearTest.test_principal_moments_of_inertia` unit test. Let me know how you'd like me to address these testing issues.

### DIFF
--- a/cclib/method/nuclear.py
+++ b/cclib/method/nuclear.py
@@ -66,7 +66,7 @@ class Nuclear(Method):
 
     def __init__(self, data, progress=None, loglevel=logging.INFO, logname="Log"):
 
-        self.required_attrs = ("natom", "atomcoords", "atomnos", "charge")
+        self.required_attrs = ('natom','atomcoords','atomnos','charge')
 
         super(Nuclear, self).__init__(data, progress, loglevel, logname)
 
@@ -86,16 +86,16 @@ class Nuclear(Method):
 
         formula = ""
         elcount = lambda el, c: "%s%i" % (el, c) if c > 1 else el
-        if "C" in elements:
-            formula += elcount("C", counts["C"])
-            counts.pop("C")
-            if "H" in elements:
-                formula += elcount("H", counts["H"])
-                counts.pop("H")
+        if 'C' in elements:
+            formula += elcount('C', counts['C'])
+            counts.pop('C')
+            if 'H' in elements:
+              formula += elcount('H', counts['H'])
+              counts.pop('H')
         for el, c in sorted(counts.items()):
             formula += elcount(el, c)
 
-        if getattr(self.data, "charge", 0):
+        if getattr(self.data, 'charge', 0):
             magnitude = abs(self.data.charge)
             sign = "+" if self.data.charge > 0 else "-"
             formula += "(%s%i)" % (sign, magnitude)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cclib/method/nuclear.py [hunk 1]
```diff
@@ -148,46 +148,46 @@ def moment_of_inertia_tensor(self, atomcoords_index=-1):
...
-        if units == 'amu_angstrom_2':
+        if units == "amu_bohr_2":
             _check_scipy(_found_scipy)
-            bohr2ang = spc.value('atomic unit of length') / spc.angstrom
-            conv = bohr2ang ** 2
-        if units == 'g_cm_2':
+            bohr2ang = spc.value("atomic unit of length") / spc.angstrom
+            conv = 1 / bohr2ang ** 2
+        if units == "amu_angstrom_2":
+            conv = 1
```
berquist: So I swapped the two cases and forgot to take the inverse of one of them...good job past me.

jamesETsmith:
Ah apologies about all the formatting changes, I've got things set up so that black reformats on save. I'll turn it off and redo the changes.

To your comment on the tests, I'll change the function call. Specifying "amu_angstrom_2" is also what's causing the problem in the rotational constants so I will change it there too.

jamesETsmith:
@berquist I reverted the previous commits that reformatted things and made the changes w/o formatting the files. I also updated the calls to `principal_moments_of_inertia()` in `rotational_constants()` and the unit test. Let me know if there is anything else I should do!

berquist:
I think https://github.com/cclib/cclib/pull/790/commits/da29ebbf8b981db9ed8c09fc634207de2d2548c1 (revert of a completely separate ORCA change) was a mistake.

jamesETsmith:
Sorry about that, I just reverted the accidental revert!

berquist:
@jamesETsmith It just occurred to me that we should have a test, in `test_nuclear`, that does exactly what you did in https://github.com/cclib/cclib/pull/790#issue-361212226. Can you write one?

jamesETsmith:
I will make a separate PR with the test and try to add it before the end of the week

Issue #657, opened 2018-11-20
Support multistep jobs generally


langner:
There are already issues for fragments (#131) and hybrid jobs (#132), this is to cover other types of multistep computations. The one I recently ran into was G3 in #656.

I think we've outlined some possible solutions in #131, and it seems each step should be a separate ccData object, and it remains to be seen if it should structured in a list, dictionary or some general object (I prefer this, since it looks like we'll go in this direction with attributes as well).

Issue #419, opened 2017-08-18
Structure of future attributes


berquist:
This is for discussion of how attributes will be represented in the future. Some relevant starting points are https://github.com/cclib/cclib/pull/398#issuecomment-323166398 and https://github.com/cclib/cclib/pull/227/.

langner:
I think the design constraints are something like this:
* Attributes are extensible, in order to parse and provide access to variantss that are rare or program-specific, but nonetheless clearly the same attribute. The two examples we've run into so far have been atomic charges (which is currently a dictionary) and transition properties (now being discussed in #398).
* Variants are accessible directly as part of the data structure (ccData), although possibly by calling some method if that makes sense.

It's pretty clear it'll be hard to make this change backwards compatible, so let's plan this for a 2.x branch.

It might make sense to think about the recently added `metadata` as well - should it remain a dictionary or adhere to this new paradigm (whatever it is)?

langner:
Some discussion from #227:
> > I'm opposed to creating multiple attributes for the same "thing". All these representation as, in principle, equivalent.
> >
> > How about we just go forward with generalizing our attributes into classes? We could do it for this attribute now, and move atomcharges and whatever else would be backwards incompatible in 2.x. This will take > some coding, but seem fairly straightforward. Actually, I see several options:
> >
> > * turn attributes into classes (changes API a lot but provides lots of flexibility)
> > * make attributes into function calls (changes API, allows us to hide multiplicity)
> > * subclass int/float/ndarray/whatever and modify getitem (does not change API, hides multiplicity)
>
> I can try and make a mockup based on top of this PR. I don't like 3, because now is the only chance to change the API. I can't visualize what 2 would look like, is there anything it could do that 1 couldn't? I'm currently in favor of 1.

I think I'm leaning towards 1 as well since it's the most flexible, but I wanted to consider the alternatives.

@ATenderholt any other ideas or suggestions at this point? What do you think about this endeavor in general?

berquist:
I have started working on this, for something to discuss: https://github.com/berquist/cclib/compare/cjson...berquist:qcjson?expand=1

Each attribute should eventually have its own `to_molden()`, `to_cjson()`, `to_qcjson()`, or whatever specialized format we want conversion to, and then the writers will collect all the available relevant attributes and put them in the right hierarchy; the formatting will be abstracted to the attribute classes. They may also contain information about the hierarchy.

[...]

Issue #99, opened 2014-04-22
Future of nmo, nbasis and similar attributes


langner:
Following #94 and similar bugs in the past, there was an idea to change `nmo` into an array, since it can change between SCF cycles as orbtials are dropped for various reasons. This issue is meant to collect our discussion on what to do with this.

Two aspects to consider:
- As Noel pointed out, changing the API would infer a major version change to cclib 2.x
- Changing `nmo` in this way suggests other attributes should also change that can behave similarly, including `nbasis`.

berquist:
There are a number of breaking changes I have in mind, along with the ones already presented here. Most of them are related to the way data is handled during geometry optimizations. Where should we have this discussion?

langner:
Perhaps a more general discussion is better suited for the mailing list.

My rough idea of the way forward is to merge in the pending branches (writers, DALTON) and do as many things as possible with backwards compatability for 1.4. After that, it might be clearer what we should change for 2.x. What do you think?

berquist:
Yes, let's do this.

PR #794, opened 2020-01-31
Added gradient parsing to molcasparser

dpadula85:
Added code to parse gradients in Molcas. I also added the code to parse QM/MM gradients computed with the ESPF method, I left it commented because maybe it is the case to create an additional attribute containing these data.


shivupa:
Hi Thanks for this and sorry for the delay on the review. This looks great. I have two things. First can you [delete this line](https://github.com/cclib/cclib/blob/1c1671bf09256f615291268c78e2aebb0f27dc35/test/data/testGeoOpt.py#L212) to enable a test for this? Secondly, can you provide more context on the QM/MM ESPF method you mention as I have never used it? When this is used each atom will only have 1 gradient? Or are the qm contributions printed first followed by the mm contributions. The comment you made in the file makes it seem like that isn't the case.  Could you perhaps run the [geometry optimization from the test suite](https://github.com/cclib/cclib/blob/master/data/Molcas/basicOpenMolcas18.0/dvb_gopt.in) with this method to compare or is it too small of a test case.

dpadula85:
Hi, thanks for the suggestions. I removed the line you asked. Before pushing again, I would like to agree with you on the test.
The documentation and examples for the ESPF method, either using the Tinker interface or not, are described here https://www.molcas.org/documentation/manual/node79.html

It's easier if you run the calculation yourself to see what is the output. Basically the program prints the gradients twice: once before the ESPF and once after. The table containing the gradients is the same, but now you have two sets of gradients in the output.
The code I wrote parses gradients for QM calculations in vacuo and only the gradient before the ESPF in QM/MM calculations. The commented code, instead, parses also the gradients after the QM/MM correction.

Since I don't know exactly how you dealt with similar situations for other codes, probably you want to parse both sections and keep them in separate attributes. Or, in case of a QM/MM calculation, store the gradients after QM/MM correction in the main attribute and keep the other ones in a "hidden" attribute (e.g. ``self._grads``).

If you want exactly the same test from the cclib suite, we have to modify it to include the ESPF section. I suppose, otherwise, we can use the test provided on the Molcas website.

Let me know what you prefer, so I can add the test in my branch and repush it.

langner:
Hi @dpadula85 - I don't think you removed that line?

I would recommend first commiting the obvious change - adding the normal gradients. And then proposing another change with the additional gradients, with a recommendation on whether to return multiple gradients or overwrite the previous ones.

dpadula85:
Hi @langner, so I pushed two changes. The first removing the line for the tests. The second is the code to parse QM/MM gradients and overwrite the previous ones, since I think this is preferable.

Thanks for the help!

cclib/parser/molcasparser.py
```diff
@@ -686,6 +686,74 @@ def extract(self, inputfile, line):
...
+            _grads = []
+            line = next(inputfile)
+            while len(line.split()) == 4:
+                tmpgrads = list(map(float, line.split()[1:]))
+                _grads.append(tmpgrads)
+                line = next(inputfile)
+
+            self._grads.append(_grads)
+
+        # Replace the main attribute with QM/MM gradients
```
langner: Could you just accumulate into a local variable in the block above, and then do `self.set_attribute('grads', grads)`? That is enough to ensure that the ESPF gradients are parsed and provided when available.

langner:
> Hi @langner, so I pushed two changes. The first removing the line for the tests. The second is the code to parse QM/MM gradients and overwrite the previous ones, since I think this is preferable.

Yup, that's great. Are you able to provide an example logfile with ESPF gradients, which we could add as a regression?

langner:
Note: I think handling this case more generally is similar to other issues we've encountered in the past, for example #99, #657 and #419 most generally. We'll address this as time permits for version 2.x

berquist:
Semi-related, but could you also fix the incorrect docstring here? https://github.com/cclib/cclib/blob/ebf1309e2c4a49d23735689cc7293418dd111c19/test/data/testGeoOpt.py#L251

cclib/parser/molcasparser.py
dpadula85: I had to do a bit differently to overwrite all previous gradients parsed to be overwritten.

berquist:
@dpadula85 Thanks for your patience!

Issue #791, opened 2020-01-11
Raman activities marked as Raman intensities


marcin-witkowski:
https://cclib.github.io/data.html states that Raman intensities are returned with vibramans, which is not true, as Raman activities (calculated with CC software) are returned. Raman intensities are dependent on the excitation lines used in the experiment.

langner:
Hi Marcin - thanks for the report. Please be more specific, in particular:
* Which parser are you referring to?
* Do you have an example logfile we can use?

marcin-witkowski:
>     * Which parser are you referring to?

I am working with Gaussian parser, but I presume the issue is universal, as it is a problem with the documentation (particularly the file linked in the original post), rather than parsers themselves.

>     * Do you have an example logfile we can use?

To be honest, the files you already have are sufficient. Consider the file cclib\data\Gaussian\basicGaussian16\dvb_raman.out, particularly lines (e.g.) 982 and 983. cclib returns the contents of such lines with vibirs and vibramans. Line 982 contains _IR Inten(sities)_, and the output of vibirs is documented accordingly. However, line 983 contains _Raman Activ(ities)_, while the cclib documentation mistakenly states that vibramans returns Raman intensities. For the discussion on the relation between Raman activities and intensities see e.g. 10.1016/j.cplett.2004.12.096 (especially eq. 7).

langner:
Thanks for clarifying, Marcin, I glanced over your original report much too quickly! That particular mention should be addressed by #791, and I changed two others I found with a brief search.

If you find anything else wrong, please do let us know. In particular, are the units for IR intensities and Raman activities correct?

langner:
Should be all good here now.

PR #795, opened 2020-02-07
[docs] Fix Raman intensities->activities (as per #791)

langner:
This addresses #791

### DIFF
--- a/cclib/parser/daltonparser.py
+++ b/cclib/parser/daltonparser.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2017, the cclib development team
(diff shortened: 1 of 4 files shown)


langner:
> This comment also needs changing:
> 
> https://github.com/cclib/cclib/blob/ebf1309e2c4a49d23735689cc7293418dd111c19/cclib/parser/gamessparser.py#L750
> 
> (and line 756)
> I don't know how it got there, since the only GAMESS example output uses activity:
> 
> https://github.com/cclib/cclib/blob/ebf1309e2c4a49d23735689cc7293418dd111c19/data/GAMESS/basicFirefly8.0/dvb_raman.out#L4964

Indeed that's rather odd, fixed it.

PR #796, opened 2020-02-09
Update context for github links in documentation

langner:
We should be pointing to the cclib repo ever since we moved the documentation sources there from the cclib/cclib.github.io repository.

Here's a screenshot below. The link that's not working is in the upper right corner:
screencapture-cclib-github-io-data-html-1581217926904

### DIFF
--- a/doc/sphinx/conf.py
+++ b/doc/sphinx/conf.py
@@ -243,9 +243,9 @@ texinfo_documents = [
 
 # Update template context with project information.
 context = {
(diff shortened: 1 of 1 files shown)


langner:
Eh, looks like I made a typo and didn't check - submitted the typo [fix directly](https://github.com/cclib/cclib/commit/e3915b20f61a2af08e8e6220ee9ad2e90f42842d).

langner:
Confirmed that it works now.

PR #798, opened 2020-02-17
QCSchema: add output writer

### LINKED ISSUE #643: Plan building out support for QCSchema JSON
We write out CJSON, but we should think about eventually supporting the QCSchema variants as well: https://github.com/MolSSI/QCSchema

Bert wrote a converter from NWChem output: https://github.com/wadejong/NWChemOutputToJson

berquist:
~~This is just a proof-of-concept to show in https://github.com/cclib/cclib/issues/643.~~ Not a proof-of-concept, rather the first implementation.


cclib/io/qcjsonwriter.py
```diff
@@ -0,0 +1,101 @@
...
+
+        qcschema_dict["schema_name"] = "qcschema_output"
+        qcschema_dict["schema_version"] = 1
+
+        qcschema_dict["molecule"] = dict()
+        qcschema_dict["molecule"]["schema_name"] = "qcschema_molecule"
+        qcschema_dict["molecule"]["schema_version"] = 2
+
+        # FIXME hack
+        qcschema_dict["driver"] = "energy"
```
shivupa, line 35: Is the thought here to populate based on `optdone` and the presence of certain attributes or by parsing `metadata->input_file_contents` or something else?

berquist:
Looks like we will need to do extra work to keep the 2.x branch up to date, or only make PRs from the head of 2.x rather than master.

cclib/io/qcjsonwriter.py
```diff
@@ -0,0 +1,101 @@
...
+        qcschema_dict["model"]["method"] = method
+
+        # FIXME
+        qcschema_dict["model"]["basis"] = ""
+
+        qcschema_dict["provenance"] = dict()
+        qcschema_dict["provenance"]["creator"] = metadata["package"]
+        qcschema_dict["provenance"]["version"] = metadata["package_version"]
+        # FIXME
+        qcschema_dict["provenance"]["routine"] = ""
```
shivupa, line 63: Would we want something to indicate that this was created through cclib rather than by the `metadata['package']` directly?

cclib/io/qcjsonwriter.py
berquist: I think we need to parse based on what's happened in the output file, rather than what's specified in the input file if possible, since programs can have a complex "path" through them when you specify even simple stuff. For example, if you say "I want a CCSD relaxed dipole moment", which is an "energy" calculation, you'll get the gradient, since calculating the coupled cluster density is expensive compared to calculating the integral derivatives, so your energy calculation has automatically turned into a gradient calculation even though it wasn't explicitly requested.

cclib/io/qcjsonwriter.py
```diff
@@ -0,0 +1,101 @@
...
+        qcschema_dict["properties"] = dict()
+        qcschema_dict["properties"]["calcinfo_nbasis"] = self.ccdata.nbasis
+        qcschema_dict["properties"]["calcinfo_nmo"] = self.ccdata.nmo
+        # TODO nalpha and nbeta
+        qcschema_dict["properties"]["calcinfo_natom"] = self.ccdata.natom
+
+        qcschema_dict["properties"]["scf_total_energy"] = convertor(
+            self.ccdata.scfenergies[-1], "eV", "hartree"
+        )
+
```
shivupa, line 74: TODO mpenergies and mpenergy for `return_result`

(You might have planned to do this, just adding a note!)

cclib/io/qcjsonwriter.py
berquist: In an ideal world, that's extra information that goes in the provenance in addition to the package. I can upload the provenance notes I took at the meeting, but I don't think the provenance substructure requires it right now.

I can try adding fields to the output that aren't part of the schema and see if it passes validation, I don't know if the validator ignores extra fields.

Also, this is a reminder that if we think something is missing from the schema, we (cclib) can request it to be added.

cclib/io/qcjsonwriter.py
berquist: Problem: untangling correlation level for Moller-Plesset and coupled cluster energies within cclib...

ghutchis:
@berquist - I don't know if you have thoughts about the QCSchema. I mentioned this work over at https://github.com/MolSSI/QCSchema/pull/77

(e.g., how much do you care if it's a community-led schema vs. one pushed by MolSSI)

berquist:
This is on hold until https://github.com/MolSSI/QCSchema/pull/81 or a variation of it is merged.

cclib/io/qcschemawriter.py
```diff
@@ -0,0 +1,251 @@
...
+                "eV",
+                "hartree",
+            )
+            return_energy = ccsd_total_energy
+        else:
+            raise RuntimeError("Don't know what to do with method {}".format(method))
+
+        scf_dipole_moment = None
+        if hasattr(self.ccdata, "moments"):
+            # FIXME We currently have no easy way to tell if the moments from
```
shivupa, line 125: We should open an issue about this. I don't this there is an easy way to fix this. If we move to class based properties maybe we could fix this?

cclib/io/qcschemawriter.py
```diff
@@ -0,0 +1,251 @@
...
+        qcschema_dict["properties"] = {
+            "calcinfo_nalpha": self.ccdata.homos[0] + 1,
+            "calcinfo_natom": self.ccdata.natom,
+            "calcinfo_nbasis": self.ccdata.nbasis,
+            "calcinfo_nbeta": self.ccdata.homos[-1] + 1,
+            "calcinfo_nmo": self.ccdata.nmo,
+            "return_energy": return_energy,
+            "scf_iterations": self.ccdata.scfvalues[-1].shape[0],
+            "scf_total_energy": scf_total_energy,
+            # TODO These properties aren't parsed yet.
```
shivupa, line 140: Same here. Maybe an issue with a check list.

shivupa (approved):
Good fix for testing. Thanks. LGTM

cclib/io/qcschemawriter.py
berquist: Yes. Under the current idea, we would have all the different moments grouped together, but in a true provenance model, we would have all properties associated with an (un)relaxed density as children of said density.

berquist:
This will be reworked to use the https://github.com/openqcschema/openqcschema (QCElemental) schema.

berquist:
Things I need to fix:
- [x] pre-commit on every commit
- [x] gpg signature
- [ ] review all of Shiv's comments again
- [ ] docs on what parts of QCSchema this covers
  - despite this not being the one from QCElemental it is similar enough that this code should go in as something preliminary
- [ ] figure out what to do with the QCSchema package dependency

PR #800, opened 2020-02-21
Ensure ccframe consistently overwrites existing data files

### LINKED ISSUE #797: ccframe seems to append data to file instead of overwriting
The problem is the following:

```bash
% rm data.h5 
% ccframe -O data.h5 opt/**/*out
% du -h data.h5 
23M	data.h5
% ccframe -O data.h5 opt/**/*out
% du -h data.h5 
45M	data.h5
% ccframe -O data.h5 opt/**/*out
% du -h data.h5 
66M	data.h5
```

Somehow the `data.h5` file keeps accumulating unnecessary data.

I'll try to do a PR to fix this as soon as I can.

schneiderfelipe:
`ccframe` had an unexpected behaviour when writing `hdf5` files (when compared to other file types).
When the output data file already existed, it didn't overwrite, but *appended* the data (because [that's panda's default](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_hdf.html)).

The problem is that, as this might get unnoticed, files keep growing, making the posterior reading process potentially slower.

This is fixed with this PR, making all output file types behave the same with regard to overwriting policy:

```bash
% rm data.h5 
% ccframe -O data.h5 opt/**/*out
% du -h data.h5 
23M	data.h5
% ccframe -O data.h5 opt/**/*out
% du -h data.h5 
23M	data.h5
```

This fixes #797.

### DIFF
--- a/cclib/scripts/ccframe.py
+++ b/cclib/scripts/ccframe.py
@@ -63,7 +63,12 @@ def main():
                         help=('name of sheet which will contain DataFrame, if '
                               'writing to an Excel file, or identifier for '
                               'the group in HDFStore, if writing a HDF file'))
+    parser.add_argument('-f', '--force', action='store_true',
+                        help=('overwrite output file in case it already exists'))
     args = parser.parse_args()
+    if args.output is not None and not args.force and os.path.exists(args.output):
+        parser.exit(1, 'failure: exiting to avoid overwriting existing file "{}"\n'.format(args.output))
+
     process_logfiles(args.compchemlogfiles, args.output, args.identifier)
 
 
--- a/cclib/scripts/ccframe.py
+++ b/cclib/scripts/ccframe.py
@@ -31,15 +31,15 @@ def process_logfiles(filenames, output, identifier):
             )
 
         if outputtype in {'csv'}:
-            df.to_csv(output)
+            df.to_csv(output, mode='w')
         elif outputtype in {'h5', 'hdf', 'hdf5'}:
-            df.to_hdf(output, key=identifier)
+            df.to_hdf(output, mode='w', key=identifier)
         elif outputtype in {'json'}:
             df.to_json(output)
         elif outputtype in {'pickle', 'pkl'}:
             df.to_pickle(output)
         elif outputtype in {'xlsx'}:
-            writer = pd.ExcelWriter(output)
+            writer = pd.ExcelWriter(output, mode='w')
             # This overwrites previous sheets
             # (see https://stackoverflow.com/a/42375263/4039050)
             df.to_excel(writer, sheet_name=identifier)
--- a/doc/sphinx/how_to_parse.rst
+++ b/doc/sphinx/how_to_parse.rst
@@ -232,8 +232,9 @@ Since the pandas library is not a dependency of cclib, `it must be installed <ht
 
 A complete data table can be parsed from many output files by following this format::
 
-    ccframe -O <OutputDest> <CompChemLogFile> [<CompChemLogFile>...]
+    ccframe [--force|-f] -O <OutputDest> <CompChemLogFile> [<CompChemLogFile>...]
 
 The argument for ``-O`` indicates the data file to be written and its extension specifies the filetype (e.g. csv, h5/hdf/hdf5, json, pickle/pkl, xlsx).
+An error will be thrown if ``<OutputDest>`` already exists, but you can force overwriting by using the ``--force`` (or ``-f``) flag.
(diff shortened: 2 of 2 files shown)


shivupa:
This seems fine.. Travis is failing very early on for some reason that seems unrelated. @langner @berquist I don't know if it is possible to retrigger a travis build, but that might fix it?

berquist:
For reasons I don't understand, despite Travis saying that's a PR build and not a branch build, the build isn't on top of the latest master. @schneiderfelipe can you rebase and force push?

langner:
> This seems fine.. Travis is failing very early on for some reason that seems unrelated. @langner @berquist I don't know if it is possible to retrigger a travis build, but that might fix it?

@shivupa it is possible, from the Travis UI. Check [this](https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit) out. If it doesn't work for you let us know, perhaps we need to update permissions somewhere.

> For reasons I don't understand, despite Travis saying that's a PR build and not a branch build, the build isn't on top of the latest master. @schneiderfelipe can you rebase and force push?

It does seem like a rebase is simple needed.

schneiderfelipe:
I finally managed to do this rebase. Thanks for the patience!

langner:
Thanks!

Two questions:
1. Do you think we should force 'w' mode for all file formats?
2. Do you think it'd be useful to check if the file exists before writing at all?

schneiderfelipe:
> Thanks!
> 
> Two questions:
> 
>     1. Do you think we should force 'w' mode for all file formats?

Excellent question! I forgot to address a strange fact in `pandas`. I checked and, for all `to_?` methods we use, the only one whose default isn't 'w' is [`to_hdf`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_hdf.html) (it is 'a' for `to_hdf`, compare with [`to_csv`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html)). I don't know why it is so, but I think it is a rather strange asymmetry in their API.

So, unless something changes in `pandas` in the future, we are already (implicitly) enforcing 'w' mode for all file formats with this PR.

>     2. Do you think it'd be useful to check if the file exists before writing at all?

I prefer documenting the behavior and keep enforcing writing.

On the other hand, I could do a new PR adding a new command-line option that ensures safe behavior (`--safe|-s`) and quits if the file already exists. If we define for this as default behavior we would need another option to `--force` writing.

langner:
> So, unless something changes in `pandas` in the future, we are already (implicitly) enforcing 'w' mode for all file formats with this PR.

That makes sense (I guess), but forcing 'w' on our end won't hurt in case the default changes in the future.

langner:
> On the other hand, I could do a new PR adding a new command-line option that ensures safe behavior (`--safe|-s`) and quits if the file already exists. If we define for this as default behavior we would need another option to `--force` writing.

I'm OK, although it feels more prudent to not overwrite by default, so if you're up for changing that I'd be happy to see it change. I don't think it's documented anywhere so the risk in doing this is small.

schneiderfelipe:
I thought about it and opted for a `--force|-f` flag that allows overwriting. Having a non-overwriting policy was easier to implement actually because neither `to_json` nor `to_pickle` accept the `mode` parameter (they overwrite by default).

langner:
> I thought about it and opted for a `--force|-f` flag that allows overwriting. Having a non-overwriting policy was easier to implement actually because neither `to_json` nor `to_pickle` accept the `mode` parameter (they overwrite by default).

OK, sounds good. Just waiting for Travis to be green now.

cclib/scripts/ccframe.py
```diff
@@ -63,7 +63,12 @@ def main():
                         help=('name of sheet which will contain DataFrame, if '
                               'writing to an Excel file, or identifier for '
                               'the group in HDFStore, if writing a HDF file'))
+    parser.add_argument('-f', '--force', action='store_true',
+                        help=('overwrite output file in case it already exists'))
     args = parser.parse_args()
+    if args.output is not None and not args.force and os.path.exists(args.output):
+        parser.exit(1, f'failure: exiting to avoid overwriting existing file "{args.output}"\n')
```
langner: There's a syntax error on this line: https://travis-ci.org/github/cclib/cclib/jobs/661763079#L1077

It might be prudent to add a unit test covering this new code.

schneiderfelipe: Oops sorry, Python 2.7 doesn't support f-strings.

schneiderfelipe:
Fixed. This should be fine now.

berquist:
Thanks!

PR #801, opened 2020-03-05
Fix Typo (numpy version in doc)

mscho527:
It looks like there is a tiny typo in the numpy version listed in the doc on line 11. Judging from line 41 and the fact that latest version of numpy being 1.19, it seems reasonable that the intended version was 1.15.


shivupa:
Thanks for this. There's an issue with Travis right now, but once we fix that up this can go in!

berquist:
Hi Minsik, can you rebase your branch against master?

mscho527:
@berquist Took a little trial and error but I believe it is all set. Thanks!

berquist:
Thank you for being so responsive!

PR #802, opened 2020-03-06
Update adding apt resource in .travis.yml

shivupa:
We needed to install a newer version of OpenBabel than was available in the apt resources of the default python Travis image. This required us to add the apt resource for Ubuntu Cosmic Cuttlefish, which is now EOL. Updating this to Eoan Ermine should fix this. It can be removed once the default python Travis image updates to the Focal LTS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ cache:
     - $HOME/miniconda
 
 before_install:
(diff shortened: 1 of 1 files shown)


shivupa:
Relevant PR (#760)

PR #804, opened 2020-03-09
ORCA parser fails for long (1,000+ lines) inputs

aalexmmaldonado:
One of the ways the ORCA parser terminates parsing input coordinates is with
```python
if line[0] == '*' or line.strip() == "end":
    break
```
However, when the input file is longer than 1,000 lines (multiple jobs in one ORCA input file), cclib will have `' *'` instead of `'*'`. By changing orcaparser to have
```python
if '*' in line or line.strip() == "end":
    break
```
we capture all circumstances.

Here is the output file from ORCA 4.2.0 that originally caused this error:
[out-CD-2MeOH-300K-1.txt](https://github.com/cclib/cclib/files/4308555/out-CD-2MeOH-300K-1.txt)

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -183,7 +183,7 @@ class ORCA(logfileparser.Logfile):
                                 continue
                             if line[0] == '#' or line.strip(' ') == '\n':
                                 continue
-                            if '*' in line or line.strip() == "end":
+                            if line.strip()[0] == '*' or line.strip() == "end":
                                 break
                             # Strip basis specification that can appear after coordinates
                             line = line.split('newGTO')[0].strip()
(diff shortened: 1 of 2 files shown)


langner:
Thanks for fixing this! One thought: using `line.strip()[0] == '*'` instead of `'*' in line` is more specific but still captures both cases. Do you think we should use that instead?

aalexmmaldonado:
More specific is always better in my opinion, so I made the switch back to using strip.

shivupa:
Can you add this test as a regression test by adding it to [cclib-data](https://github.com/cclib/cclib-data) and adding it to [regression.py](https://github.com/cclib/cclib/blob/master/test/regression.py). I can help more if needed.

aalexmmaldonado:
No problem. I believe I added everything.

langner:
Thanks for the extra effort!

PR #811, opened 2020-03-10
Make makecclib the inverse of makeopenbabel

schneiderfelipe:
(This substitutes and implements the same changes as #759. All suggestions have been included here.)

It's currently impossible to call `makecclib` from the result of `makeopenbabel` since `makeopenbabel` demands the dimension of `atomcoords` to be three. In contrast, the return object of `makecclib` is always two dimensional.

This PR ensures `np.asarray(atomcoords).ndim == 3` for the return object of ` cclib.bridge.cclib2openbabel.makecclib`.


langner:
Is there a test you can add that covers this?

schneiderfelipe:
> Is there a test you can add that covers this?

I wrote a simple test. Also, `makecclib` lacked a test, so it ended up being a test for it as well.

langner:
Looks good, but please wait for #816 and Travis to be green before merging.

PR #812, opened 2020-03-10
Improve parsing of entropy (and enthalpy and freeenergy)

### LINKED ISSUE #663: "entropy" has two meanings depending on the computational package
Hi, it's me again.

Entropies extracted from Gaussian (09C01) or ORCA (4.0.1.2) logfiles have different meanings:

```
% # First a Gaussian logfile:
% ccget entropy enthalpy freeenergy gaussian_out.out
Attempting to read gaussian_out.out
entropy
0.00016293141036397934
enthalpy
-552.415496
freeenergy
-552.464074
```

Results from Gaussian assume $G = H - T S$ as usual (`entropy` means $S$ here):

```
In [1]: -298.15 * 0.00016293141036397934 + -552.415496                          
Out[1]: -552.464074
```

```
% # Now a ORCA logfile:
% ccget entropy enthalpy freeenergy orca_out.out
Attempting to read orca_out.out
entropy
0.04797662
enthalpy
-612.38317299
freeenergy
-612.43114961
```

But in results from ORCA, `entropy` means $T S$!

```
In [2]: -0.04797662 + -612.38317299                                             
Out[2]: -612.43114961
```

Is this wanted? I can imagine algorithms using ccclib not working across packages because of this.

schneiderfelipe:
This should fix #663.


test/data/testvib.py
```diff
@@ -130,7 +130,7 @@ def testenthalpy(self):
 
     def testentropy(self):
          """Is the entropy reasonable"""
-         self.assertAlmostEqual(0.03601749, self.data.entropy, self.entropy_places)
+         self.assertAlmostEqual(0.00012080325339594164, self.data.entropy, self.entropy_places)
```
schneiderfelipe: Aren't we rounding too soon with `entropy_places = 3` (line 116)?

schneiderfelipe: There are uneven rounding places for entropy across programs. Gamess for instance is tested against 6 decimal places.

langner: What do you recommend we do instead?

langner: Do you know what the units here are? It is kcal/(mol K) like the logfile suggests?

test/data/testvib.py
```diff
@@ -67,6 +67,16 @@ def testvibsyms(self):
...
 
+    entropy_places = 3
+
+    def testtemperature(self):
+        """Is the temperature 298.15 K?"""
+        self.assertAlmostEqual(298.15, self.data.temperature)
+
+    def testentropy(self):
+         """Is the entropy reasonable"""
+         self.assertAlmostEqual(0.0001462623335480945, self.data.entropy, self.entropy_places)
```
langner: This looks like a test we could move to GenericIRTest, and just customize the value in the package-specific test classes.

cclib/parser/data.py
```diff
@@ -35,7 +35,7 @@ class ccData(object):
         charge -- net charge of the system (integer)
         coreelectrons -- number of core electrons in atom pseudopotentials (array[1])
         enthalpy -- sum of electronic and thermal enthalpies (float, hartree/particle)
-        entropy -- entropy (float, hartree/particle)
+        entropy -- entropy (float, hartree/particle*kelvin)
```
langner: 1. Is it customary to add the "particle" bit for entropy?
2. Shouldn't his be hartree/(particle*Kelvin) - currently looks a little like K is in the numerator if read programmatically.

test/data/testvib.py
schneiderfelipe: As long as the numbers are reliable, we could attempt to be as strict as possible: maximum number of common digits. (The figures I came up with were calculated independently so I believe they are reliable.)

Another possibility would be to set an energy/entropy cut, e.g. 1e-3 hartree (1e-4 hartree) ensures an error smaller than 0.63 kcal/mol (0.063 kcal/mol).

Entropy cuts need then to be calculated separately: by using a cut of 1e-3 hartree/K, we are accepting errors around 187 kcal/mol at 298.15 K! We need 1e-6 hartree/K (1e-7 hartree/K) for errors less than 0.2 kcal/mol (0.02 kcal/mol).

 I prefer 1e-4 hartree and 1e-7 hartree/K.

schneiderfelipe: True. I insisted a bit in hartree/(particle*kelvin) to keep consistency with other fields in the docs such as `enthalpy` and `freeenergy`, but we can change them all. Essentially, "particle" here means "unity".

cclib/parser/data.py
schneiderfelipe: >     1. Is it customary to add the "particle" bit for entropy?

I used to keep consistency with [the docs](https://cclib.github.io/data.html). I think it is fine if we remove.

>     2. Shouldn't his be hartree/(particle*Kelvin) - currently looks a little like K is in the numerator if read programmatically.

You're right. I'll fix that.

langner: I don't mind. Actually I would defer to your judgement whether the particle bit should stay, just would press on trying to keep being consistent throughout the docs and any comments (it's always a struggle).

schneiderfelipe: OK. So I would say to just add the parenthesis and leave the "particle" as it is.

langner: Sounds good.

cclib/parser/molcasparser.py
```diff
@@ -483,7 +483,8 @@ def extract(self, inputfile, line):
                 if line[1:48] == 'Molecular Partition Function and Molar Entropy:':
                     while 'TOTAL' not in line:
                         line = next(inputfile)
-                    entropy_values.append(utils.convertor(float(line.split()[2]), 'kcal/mol', 'hartree'))
+                    # Molcas reports entropy values in kcal/mol*K but actually writes them in cal/mol*K
+                    entropy_values.append(utils.convertor(float(line.split()[2]), 'kcal/mol', 'hartree') / 1000)
```
langner: Do we need to sum the entropy components for Moclas as well?

schneiderfelipe: I don't think so. The output of Molcas produces values for a set of temperatures, we end up getting the entropy value for 298.15 K in https://github.com/cclib/cclib/blob/b50fdfd557077e9dfbd3d1605c07b5f82d90fb04/cclib/parser/molcasparser.py#L495-L513

Is that what you mean?

test/data/testvib.py
schneiderfelipe: Which tests do you think we could move to GenericIRTest? We could move tests for temperature and pressure as well, for instance.

langner: I was thinking only about the entropy at this time, since that is what you are changing.

test/data/testvib.py
langner: SGTM

cclib/parser/molcasparser.py
langner: I see. And is there something we can do if 298.15 is not there but other temperature are?

schneiderfelipe: Not sure, I've never used Molcas. Is it possible that 298.15 K is always reported?

langner: I guess the comment says it all. Looks fine as far as your changes are concerned.

test/data/testvib.py
langner: OK then, do you want to make the default for `entropy_places` more stringent and loosen it for select packages?

schneiderfelipe: Yes, just to be sure that the parsing is working within a precise but reasonable energy deviation.

langner:
@schneiderfelipe sorry but there seem to have been conflicts merged in recently - possibly mine.

You need to merge or rebase manually - let me know if you need help with that.

schneiderfelipe:
This is failing and I don't know why. Errors are not due to decimal places as they didn't change after the last commit.

Failures are due to `dvb_ir.out` ORCA logfiles, but the newest version (4.0) is the only one that doesn't fail. In summary, errors happen with (2 failed tests each):

- `ORCA2.6/dvb_ir.out`, associated with class `OrcaIRTest_old_coordsOK` (inherits from `OrcaIRTest`)
- `ORCA2.8/dvb_ir.out`, associated with class `OrcaIRTest_old` (inherits from `OrcaIRTest` too)
- `ORCA2.9/dvb_ir.out`, associated with class `OrcaIRTest`
- `ORCA3.0/dvb_ir.out`, associated with class `OrcaIRTest` as well

`ORCA4.0/dvb_ir.out` presents no error, but is **associated with class `GenericIRTest`**, which doesn't test for entropy/enthalpy/freeenergy values. Furthermore, `OrcaIRTest` inherits from `GenericIRTest`.

This is clearly inconsistent. Is there a reason for this? What should we do?

langner:
> This is failing and I don't know why. Errors are not due to decimal places as they didn't change after the last commit.
> 
> Failures are due to `dvb_ir.out` ORCA logfiles, but the newest version (4.0) is the only one that doesn't fail. In summary, errors happen with (2 failed tests each):
> 
> * `ORCA2.6/dvb_ir.out`, associated with class `OrcaIRTest_old_coordsOK` (inherits from `OrcaIRTest`)
> * `ORCA2.8/dvb_ir.out`, associated with class `OrcaIRTest_old` (inherits from `OrcaIRTest` too)
> * `ORCA2.9/dvb_ir.out`, associated with class `OrcaIRTest`
> * `ORCA3.0/dvb_ir.out`, associated with class `OrcaIRTest` as well
> 
> `ORCA4.0/dvb_ir.out` presents no error, but is **associated with class `GenericIRTest`**, which doesn't test for entropy/enthalpy/freeenergy values. Furthermore, `OrcaIRTest` inherits from `GenericIRTest`.
> 
> This is clearly inconsistent. Is there a reason for this? What should we do?

Well, I think the cause of this is the improved accuracy in your checks. We've changed the coordinates and such over the years so that unit test logfiles for different packages are more aligned. So I'm not surprised we would see slight differences for these older logfiles which are now in the regression suite.

You can adjust the test class attributes in a custom test class and assign it to the old unit test in the regression suite. See this for an example: https://github.com/cclib/cclib/blob/master/test/regression.py#L2927. I know this is a little clumsy, and we need to improve the usability, plus that file is getting mighty large... and that will eventually get improved.

P. S. Trying to add more traceback information to the regression suite run on Travis in #834. Also, for some reason I don't see the Travis links/icons in github PRs - do you guys see them?

langner:
> > Well, I think the cause of this is the improved accuracy in your checks.
> 
> I don't think so. This is the [Travis build](https://travis-ci.org/github/cclib/cclib/builds/671655276) for [1ed29be](https://github.com/cclib/cclib/commit/1ed29bee41d2b00a5fb3fe220bb5ab7b7baf8ded) (prior to the change in accuracy). Here is the [Travis build for the last commit](https://travis-ci.org/github/cclib/cclib/builds/671669604) ([96e33a2](https://github.com/cclib/cclib/commit/96e33a22b291d82e79c7e95025f85637b18b31fa)), including the improved accuracy. The errors are the same.

Hmm.. need to look into this then.

> > P. S. Trying to add more traceback information to the regression suite run on Travis in #834.
> 
> Thanks. Let's see what the traceback info gives us.

Unfortunately my PR broke Travis so I reverted it. Need to figure out how to pass an additional flag using pytest.

> > Also, for some reason I don't see the Travis links/icons in github PRs - do you guys see them?
> 
> I don't see anything as well.

Seems there was some kind of migration from travis-ci.org to travis-ci.com. Removed and re-added the app integration and works again now from the new domain.

schneiderfelipe:
> Unfortunately my PR broke Travis so I reverted it. Need to figure out how to pass an additional flag using pytest.

I think adding something like

```
[pytest]
addopts = --tb=long
```

to [`pytest.ini` might work](https://docs.pytest.org/en/latest/customize.html#adding-default-options).

> Seems there was some kind of migration from travis-ci.org to travis-ci.com. Removed and re-added the app integration and works again now from the new domain.

Yes, I see it now. Thanks!

langner:
> > Unfortunately my PR broke Travis so I reverted it. Need to figure out how to pass an additional flag using pytest.
> 
> I think adding something like
> 
> ```
> [pytest]
> addopts = --tb=long
> ```

Unfortunately not. I'm trying to add use a flag that is defined in `regression.py`.

schneiderfelipe:
> Unfortunately not. I'm trying to add use a flag that is defined in `regression.py`.

Sorry, I thought that was the pytest `--tb` flag!

berquist (approved):
Once this is rebased and the CI failure is fixed, I'm happy with this.

cclib/parser/orcaparser.py
```diff
@@ -947,7 +947,7 @@ def splitter(line):
 
             # ORCA prints -inf for single atom entropy.
             if self.natom > 1:
-                self.entropy = float(next(inputfile).split()[4])
+                self.entropy = float(next(inputfile).split()[4]) / self.temperature
```
schneiderfelipe, line 950: ORCA reports entropy as `T S`.

berquist (approved):
Sorry, I was hacking on this to remove most of the duplication in the inherited tests, then I think I discovered some other issues, not sure...but that's not worth dragging out this particular PR any further. Thanks for doing this!

PR #813, opened 2020-03-11
Correct vibfreqs, vibdisps, vibirs and vibramans for single atom

### LINKED ISSUE #778: Frequency calculation with single atom crashes cclib (ORCA)
The attached output ([cl.zip](https://github.com/cclib/cclib/files/3743973/cl.zip)) file throws an error when parsed:

```python
import cclib
cclib.ccopen("../../data/cl.out").parse()
```

The error:

```python
...
.../cclib/cclib/parser/logfileparser.py in parse(self, progress, fupdate, cupdate)
    327             #   in data._attrlist will be moved to final data object that is returned.
    328             try:
--> 329                 self.extract(inputfile, line)
    330             except StopIteration:
    331                 self.logger.error("Unexpectedly encountered end of logfile.")

.../cclib/cclib/parser/orcaparser.py in extract(self, inputfile, line)
   1009 
   1010             nonzero = numpy.nonzero(vibfreqs)[0]
-> 1011             self.first_mode = nonzero[0]
   1012             # Take all modes after first
   1013             # Mode between imaginary and real modes could be 0

IndexError: index 0 is out of bounds for axis 0 with size 0
```

The problem seems to be the number of frequencies (it's a single atom!):

```
...
-----------------------
VIBRATIONAL FREQUENCIES
-----------------------

Scaling factor for frequencies =  1.000000000  (already applied!)

   0:         0.00 cm**-1
   1:         0.00 cm**-1
   2:         0.00 cm**-1


------------
NORMAL MODES
------------
...
```

Could a `try...except` clause solve it? Some parts of the output are also affected by the fact that it's a single atom, such as entropy calculations, which could also throw new errors:

```
...
The entropies will be listed as multiplied by the temperature to get
units of energy

Electronic entropy                ...      0.00065446 Eh      0.41 kcal/mol
Vibrational entropy               ...            -inf Eh      -inf kcal/mol
Rotational entropy                ...      0.00000000 Eh      0.00 kcal/mol
Translational entropy             ...      0.01740262 Eh     10.92 kcal/mol
-----------------------------------------------------------------------
Final entropy term                ...            -inf Eh      -inf kcal/mol


-------------------
GIBBS FREE ENTHALPY
-------------------
...
```

The correct behaviour is to consider `Vibrational entropy` equal to zero and sum all lines again into `Final entropy term`. (Enthalpies look fine, but the `Final Gibbs free enthalpy` is also `inf` in the output....)

schneiderfelipe:
This PR corrects `vibfreqs`, `vibdisps`, `vibirs` and `vibramans` in ORCA, for logfiles with only one atom.

This should fix #778.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -1059,20 +1059,24 @@ States  Energy Wavelength    D2        m2        Q2         D2+m2+Q2       D2/TO
             if float(self.metadata["package_version"][:3]) > 4.0:
                 self.skip_lines(inputfile, ['Scaling factor for frequencies', 'b'])
 
-            vibfreqs = numpy.zeros(3 * self.natom)
-            for i, line in zip(range(3 * self.natom), inputfile):
-                vibfreqs[i] = float(line.split()[1])
+            if self.natom > 1:
+                vibfreqs = numpy.zeros(3 * self.natom)
+                for i, line in zip(range(3 * self.natom), inputfile):
+                    vibfreqs[i] = float(line.split()[1])
 
-            nonzero = numpy.nonzero(vibfreqs)[0]
-            self.first_mode = nonzero[0]
-            # Take all modes after first
-            # Mode between imaginary and real modes could be 0
-            self.num_modes = 3*self.natom - self.first_mode
-            if self.num_modes > 3*self.natom - 6:
-                msg = "Modes corresponding to rotations/translations may be non-zero."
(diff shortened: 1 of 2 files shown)


schneiderfelipe:
This PR currently leaves `entropy` and `freeenergy` as `-inf` and `inf`, respectively. This behavior is tested, however.
The code for solving this is simple, but I would rather wait for #812 to avoid conflicts.

langner:
Thanks, looks generally good!

Issue #814, opened 2020-03-12
missing `comment` key in parsed `ORCA` `ccData`


pzarabadip:
I just realized that the parsed object data from `ORCA` (`version 4.2.1`) is missing `comments` key which will result in throwing `KeyError` in the case of writing xyz file. 
https://github.com/cclib/cclib/blob/87abf82c6a06836a2e5fb95a64cdf376c5ef8d4f/cclib/io/xyzwriter.py#L84
Currently, my not-clean work around to continue my work is using `setattribute` method to artificially add it to `metadata` dictionary but I thought to open this issue to report the case and keeping it here until implementing better solution and possibly opening a PR. 
I will be happy to hear opinions on this matter for solving the issue.

schneiderfelipe:
It also fails for other logfiles:

```python3
>>> import cclib
>>> cclib.io.XYZWriter(cclib.ccopen("data/Molcas/basicOpenMolcas18.0/dvb_ir.out").parse()).generate_repr()
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-23-fafefcf9e718> in <module>()
----> 1 cclib.io.XYZWriter(cclib.ccopen("data/Molcas/basicOpenMolcas18.0/dvb_ir.out").parse()).generate_repr()

/mnt/Files/Dropbox/Dropbox/cclib/cclib/io/xyzwriter.py in generate_repr(self)
     71             indices = [-1]
     72         for i in indices:
---> 73             xyzblock.append(self._xyz_from_ccdata(i))
     74 
     75         # Ensure an extra newline at the very end.

/mnt/Files/Dropbox/Dropbox/cclib/cclib/io/xyzwriter.py in _xyz_from_ccdata(self, index)
     82 
     83         atomcoords = self.ccdata.atomcoords[index]
---> 84         existing_comment = "" if not self.ccdata.metadata["comments"] \
     85             else self.ccdata.metadata["comments"][index]
     86 

KeyError: 'comments'
```

The same error also happens with `data/Gaussian/basicGaussian16/dvb_ir.out`.

pzarabadip:
Thanks @schneiderfelipe. Then, I will open a PR in a while which would fix this issue.

PR #815, opened 2020-03-12
fixing an issue in xyzwriter

pzarabadip:
Hi,
This PR aims to fix the #814.

### DIFF
--- a/cclib/io/xyzwriter.py
+++ b/cclib/io/xyzwriter.py
@@ -81,7 +81,7 @@ class XYZ(filewriter.Writer):
         """Create an XYZ file of the geometry at the given index."""
 
         atomcoords = self.ccdata.atomcoords[index]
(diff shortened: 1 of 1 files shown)


langner:
Thanks for the fix! Are you able to also add a unit test for this?

pzarabadip:
Sure @langner 
I thought there are tests already in place for this. It may be a bit delayed from my side but it would quite helpful if you can point me to similar ones.

langner:
> Sure @langner
> I thought there are tests already in place for this. It may be a bit delayed from my side but it would quite helpful if you can point me to similar ones.

The existing tests for this module are here: https://github.com/cclib/cclib/blob/master/test/io/testxyzwriter.py

It would be ideal if we could get a test that covers the case you're fixing with this PR.

berquist:
@pzarabadip Poke.

langner:
This was done in #989

PR #817, opened 2020-03-13
Correct parsing of entropy for single atoms in ORCA

### LINKED ISSUE #816: "TypeError: object of type 'float' has no len()" for -inf entropy (single atom in ORCA)
The following comes from Travis:

```python3
    def testORCA_ORCA4_1_single_atom_freq_out(logfile):

        """ORCA frequency with single atom."""

        assert len(logfile.data.vibdisps) == 0

        assert len(logfile.data.vibfreqs) == 0

        assert len(logfile.data.vibirs) == 0

>       assert len(logfile.data.entropy) == -numpy.inf

E       TypeError: object of type 'float' has no len()

test/regression.py:1651: TypeError
```

I want to address it in #812. It is a simple fix, just summing the actual contributions for entropy.
(Giving `-inf` entropy on single atom calculations is [an issue on ORCA's side](https://orcaforum.kofo.mpg.de/viewtopic.php?f=8&t=5436).)

schneiderfelipe:
This should fix #816.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -780,7 +780,8 @@ class ORCA(logfileparser.Logfile):
             line = next(inputfile)
             while line[:17] != 'Electronic energy':
                 line = next(inputfile)
-            self.zpe = next(inputfile).split()[4]
+            self.electronic_energy = float(line.split()[3])
+            self.zpe = float(next(inputfile).split()[4])
             thermal_vibrational_correction = float(next(inputfile).split()[4])
             thermal_rotional_correction = float(next(inputfile).split()[4])
             thermal_translational_correction = float(next(inputfile).split()[4])
@@ -793,7 +794,11 @@ class ORCA(logfileparser.Logfile):
                 line = next(inputfile)
             thermal_enthalpy_correction = float(next(inputfile).split()[4])
             next(inputfile)
-            self.enthalpy = float(next(inputfile).split()[3])
+
+            if self.natom > 1:
+                self.enthalpy = float(next(inputfile).split()[3])
+            else:
+                self.enthalpy = self.electronic_energy + thermal_translational_correction
 
             # Entropy
             line = next(inputfile)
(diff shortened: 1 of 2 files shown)


cclib/parser/orcaparser.py
```diff
@@ -804,12 +809,20 @@ def splitter(line):
             rotational_entropy = float(next(inputfile).split()[3])
             translational_entropy = float(next(inputfile).split()[3])
             next(inputfile)
-            self.entropy = float(next(inputfile).split()[4])
+
+            if self.natom > 1:
+                self.entropy = float(next(inputfile).split()[4])
+            else:
+                self.entropy = (electronic_entropy + translational_entropy) / self.temperature
```
langner: Is this right? Do you need to divide by the temperature here?

cclib/parser/orcaparser.py
```diff
@@ -793,7 +794,11 @@ def splitter(line):
                 line = next(inputfile)
             thermal_enthalpy_correction = float(next(inputfile).split()[4])
             next(inputfile)
-            self.enthalpy = float(next(inputfile).split()[3])
+
+            if self.natom > 1:
+                self.enthalpy = float(next(inputfile).split()[3])
+            else:
+                self.enthalpy = self.electronic_energy + thermal_translational_correction
```
langner: Enthalpy doesn't seem to be affected (at least in [the regression file](https://github.com/cclib/cclib-data/blob/master/ORCA/ORCA4.1/single_atom_freq.out#L879)). So do we need this?

langner:
I'm going to merge this in without resolving the last comment, just to get Travis green again...

cclib/parser/orcaparser.py
langner: OK, it is :)

cclib/parser/orcaparser.py
langner: Removed in #818.

schneiderfelipe: This looks innocent but is in fact a dangerous value. It includes the [total free energy (i.e. inner energy)](https://github.com/cclib/cclib-data/blob/25cb852e143e4e2e790aa415b1b0e0ca6c190603/ORCA/ORCA4.1/single_atom_freq.out#L876), which includes the [spurious vibrational correction](https://github.com/cclib/cclib-data/blob/25cb852e143e4e2e790aa415b1b0e0ca6c190603/ORCA/ORCA4.1/single_atom_freq.out#L855) that ORCA thought was nonzero (because [it could not remove the translational modes](https://github.com/cclib/cclib-data/blob/25cb852e143e4e2e790aa415b1b0e0ca6c190603/ORCA/ORCA4.1/single_atom_freq.out#L773-L777)).

This means that the enthalpy ORCA reports is wrong by 1.78 kcal/mol. It even calculates a ZPE which, although almost zero in this regression file, might not be in others (I'm not sure, depends on the spurious translational frequencies, which probably depend on atomic masses, I would have to look closer).

In the end, the [only correction to electronic energy that is valid is the translational one](https://github.com/cclib/cclib-data/blob/25cb852e143e4e2e790aa415b1b0e0ca6c190603/ORCA/ORCA4.1/single_atom_freq.out#L854-L857), which what I included.

PR #818, opened 2020-03-20
Clean up ORCA thermochemistry section a bit

langner:
This is a follow up to #817.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -763,15 +763,39 @@ class ORCA(logfileparser.Logfile):
                 self.gbasis.append(gbasis_tmp[bas_atname])
             del self.tmp_atnames
 
-        """ Banner announcing Thermochemistry
+        """
         --------------------------
         THERMOCHEMISTRY AT 298.15K
         --------------------------
-        """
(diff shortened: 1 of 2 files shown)


cclib/parser/orcaparser.py
```diff
@@ -785,40 +809,37 @@ def splitter(line):
...
             while line[:17] != 'Total free energy':
                 line = next(inputfile)
             thermal_enthalpy_correction = float(next(inputfile).split()[4])
             next(inputfile)
-
-            if self.natom > 1:
-                self.enthalpy = float(next(inputfile).split()[3])
-            else:
-                self.enthalpy = self.electronic_energy + thermal_translational_correction
+            self.enthalpy = float(next(inputfile).split()[3])
```
schneiderfelipe: This enthalpy would be wrong for single atoms due to spurious translational frequencies ORCA doesn't remove in this case ([as discussed here](https://github.com/cclib/cclib/pull/817#discussion_r395715175)).

langner: Gotcha - added a comment about this then, in the code and in `regressions.py`. Thanks for point this out!

berquist:
Looks good to me. @schneiderfelipe can you take one more look?

PR #799, opened 2020-02-17
Add Psi4 1.3.1 unit tests

berquist:

PR #819, opened 2020-03-20
[Psi4] Remove 1.0 unit test logfiles and add them to the regression suite

langner:
This is a follow up to #799 and needs to go in after cclib/cclib-data#126


shivupa:
Am I correct in thinking this looks ready to merge?

langner:
> Am I correct in thinking this looks ready to merge?

Yes, please do.

PR #821, opened 2020-03-26
Created geometry.py containing various geometrical parameters

### LINKED ISSUE #489: New method: calculate all possible distances, angles, dihedrals
As discussed on the dev mailing list, we should have a method that calculates all possible distances, angles, dihedrals, etc.

Possibly useful references:
* https://github.com/CrawfordGroup/ProgrammingProjects/tree/master/Project%2301
* https://github.com/berquist/programming_party/blob/master/eric/molecule.py
* https://stackoverflow.com/q/20305272/3249688

From Karol:
> Regarding the coordinate-to-angle calculations - they'll be so fast for 99% of the cases that it's not worthwhile to think about calculating things on-the-fly.

So things should be calculated all at once.

GomaruSai:
distance, angle and dihedral

I have created methods for calculating the distance, angle and dihedral angle in this file called geometry.py

I believe this partially resolves #489 

I have yet to change the writer to incorporate these into the final output file. I request help regarding how to start doing this.

### DIFF
--- a/cclib/method/geometry.py
+++ b/cclib/method/geometry.py
@@ -1,0 +1,36 @@
+import numpy
+
+def getDistance(coords0, coords1):
+    """Distance between atoms 0-1"""
+    return numpy.linalg.norm(numpy.subtract(coords0 - coords1))
+
+def getAngle(coords0, coords1, coords2):
+    """Angle(in radians) between atoms 0-1-2"""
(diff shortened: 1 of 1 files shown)


langner:
Why shouldn't this stuff go into [nuclear.py](https://github.com/cclib/cclib/blob/master/cclib/method/nuclear.py)?

GomaruSai:
Yes, I guess that would make more sense.
I will work on it.

GomaruSai:
I have made another pull request putting these methods into nuclear.py #837

PR #822, opened 2020-03-26
Print all citations to the cclib paper in rst file

### LINKED ISSUE #610: Expand on literature using/citing cclib
(this is moved from cclib/cclib.github.io#19)

This could go in the docs somewhere.

GomaruSai:
The above api called scholarly can be used to retreive the list of citations from google scholar. This code may partially resolve #610 
However, it requires the installation of the arrow, Beautiful Soup, bibtexparser, and requests[security] libraries, which may not all be dependancies of cclib.
This has to be implemented into the rst file of the overview section in the docs

### DIFF
--- a/doc/sphinx/citations.py
+++ b/doc/sphinx/citations.py
@@ -1,0 +1,13 @@
+import scholarly
+
+# Retrieve the author's data and fill-in
+search_query = scholarly.search_author('Karol M. Langner')
+author = next(search_query).fill()
+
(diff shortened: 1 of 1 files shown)


langner:
What is the benefit of having the list in our docs above just linking to Google Scholar?

langner:
I'm going to close. If other devs feel this would be valuable for some reason, we can reconsider.

PR #823, opened 2020-03-26
Add Gradient & Laplacian in Cartesian Coordinates

GomaruSai:
I have implemented gradient and laplacian in cartesian coordinate system for the periodic boundary condition.
I believe this will aid in further calculations.


langner:
Was there somethingwrong?

PR #826, opened 2020-04-01
cElementTree has been deprecated since Python 3.3 and removed in Python 3.9

### LINKED ISSUE #824: xml.etree.cElementTree removed from Python 3.9 in favour of xml.etree.ElementTree
https://github.com/cclib/cclib/blob/428b4cf07dfd91a6b28c49c63d93ddc0175b5cf2/cclib/io/cmlwriter.py#L10

Ref : https://github.com/python/cpython/pull/19108

tirkarthi:
Fixes #824

### DIFF
--- a/cclib/io/cmlwriter.py
+++ b/cclib/io/cmlwriter.py
@@ -7,7 +7,7 @@
 
 """A writer for chemical markup language (CML) files."""
 
-import xml.etree.cElementTree as ET
+import xml.etree.ElementTree as ET
 
 from cclib.io import filewriter
 from cclib.parser.utils import find_package


cclib/io/cmlwriter.py
```diff
@@ -7,7 +7,10 @@
 
 """A writer for chemical markup language (CML) files."""
 
-import xml.etree.cElementTree as ET
+try:
```
langner, line 10: Could you add a comment explaining why this is necessary?

tirkarthi: Added a code comment about deprecation. Please let me know if you have any suggestions, Thanks.

langner: If you're up for it, it would actually be better to just check the Python version, instead of catching the exception. I believe the fast version is used whenever possible starting from version 3.3

tirkarthi: I have seen this to be suggested as a common idiom

https://bugs.python.org/msg365079

ATenderholt: A couple of thoughts:

* For Python 3, we're only supporting versions > 3.3.
* Python 2 is EOL'd.

Can't we just use `xml.etree.ElementTree`? Python 3 is already using the accelerated version under the hood, and do we really care about performance of Python 2 at this point?

langner: Oh, I thought this is because something didn't work.... if that's not the case then yes, please, just use the non-c sytnax.

tirkarthi: Thanks, changed the import as suggested.

langner:
Thanks, this is ready to go. Could you squash your commits into the last one - this change is minimal.

tirkarthi:
Thanks @langner, squashed the commits.

PR #837, opened 2020-04-10
Added Geometrical parameters to nuclear.py

### LINKED ISSUE #489: New method: calculate all possible distances, angles, dihedrals

GomaruSai:
Added distance, angle and dihedral methods to nuclear.py.
This will parially resolve #489
What's left is to figure out a way to represent this data in the final output file.

### DIFF
--- a/cclib/method/nuclear.py
+++ b/cclib/method/nuclear.py
@@ -100,6 +100,91 @@ class Nuclear(Method):
             sign = "+" if self.data.charge > 0 else "-"
             formula += "(%s%i)" % (sign, magnitude)
         return formula
+    
+    def get_distances(self, atomcoords_index = -1):
+        '''Returns a matrix with distances between i_th and j_th atoms at index [i][j], where i < j
+        *Other indices will give wrong values*'''
+        coords = self.data.atomcoords[atomcoords_index]
+        number_of_atoms = self.data.atomcoords.size[1]
+        distance_matrix = np.empty((number_of_atoms, number_of_atoms))
+        i = 0
+        j = 1
+        while(i < number_of_atoms):
+            while(j < number_of_atoms):
+                distance_matrix[i][j] = np.linalg.norm(np.subtract(coords[j], coords[i]))
+                j += 1
+            i += 1
+            j = i + 1
+        return distance_matrix
+
+    def get_angles(self, atomcoords_index = -1):
+        """Returns a matrix with Angles(in radians) between atoms i-j-k at index [i][j][k], where i < j < k
+        *Other indices will give wrong values*"""
+        coords = self.data.atomcoords[atomcoords_index]
+        number_of_atoms = self.data.atomcoords.size[1]
+        angle_matrix = np.empty((number_of_atoms, number_of_atoms, number_of_atoms))
+        i = 0
+        j = 1
+        k = 2
+        while(i < number_of_atoms):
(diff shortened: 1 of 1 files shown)


cclib/method/nuclear.py
```diff
@@ -100,6 +100,91 @@ def stoichiometry(self):
             sign = "+" if self.data.charge > 0 else "-"
             formula += "(%s%i)" % (sign, magnitude)
         return formula
+    
+    def get_distances(self, atomcoords_index = -1):
```
langner, line 104: We should just call this `distances` to keep in line with `stiochiometry` in this class and the no-get pattern we have in other places.

cclib/method/nuclear.py
```diff
@@ -100,6 +100,91 @@ def stoichiometry(self):
...
+        '''Returns a matrix with distances between i_th and j_th atoms at index [i][j], where i < j
+        *Other indices will give wrong values*'''
```
langner, line 106: What do you mean by wrong?

cclib/method/nuclear.py
```diff
@@ -100,6 +100,91 @@ def stoichiometry(self):
...
+        coords = self.data.atomcoords[atomcoords_index]
+        number_of_atoms = self.data.atomcoords.size[1]
```
langner, line 108: Why not use `natom`?

cclib/method/nuclear.py
```diff
@@ -100,6 +100,91 @@ def stoichiometry(self):
...
+        distance_matrix = np.empty((number_of_atoms, number_of_atoms))
+        i = 0
+        j = 1
+        while(i < number_of_atoms):
```
langner, line 112: In Python it is more more conventional to use `for` loops for these types of things.

cclib/method/nuclear.py
GomaruSai: Yes. I guess that makes sense. I have made changes in the new commit.

cclib/method/nuclear.py
GomaruSai: I used np.empty and did not populate the matrix at those indices. However, I have resolved this in the new commit. There is no longer any restriction in the indices to be used, unless it means the same thing.

cclib/method/nuclear.py
GomaruSai: Yes. I guess that makes sense. I have made changes in the new commit.

cclib/method/nuclear.py
GomaruSai: Yes. I guess that makes sense. I have made changes in the new commit.

cclib/method/nuclear.py
GomaruSai: *as long as it means the same thing

berquist:
Since there was never any response on this, in the interest of letting someone else work on #489, I'll close it.

PR #839, opened 2020-04-14
Add point group detection, take 2: DALTON

berquist:
This does three things:
- Adds a test for detected (full) and used (most likely largest Abelian subgroup) point group
- Enables the test for DALTON, which has point group detection implemented.
- Updates the DALTON no symmetry regression test to look for C1 symmetry rather than C2h (the DVB point group). In doing this, I discovered that it used `Ag` as the no-symmetry irrep symbol when it should be `A`.

### DIFF
--- a/cclib/parser/daltonparser.py
+++ b/cclib/parser/daltonparser.py
@@ -46,13 +46,12 @@ class DALTON(logfileparser.Logfile):
         # would like to use as triggers.
         self.section = None
 
-        # If there is no symmetry, assume this.
+        # If there is no symmetry (point group is C1), assume this.
         self.symlabels = ['Ag']
 
         # Is the basis set from a single library file? This is true
         # when the first line is BASIS, false for INTGRL/ATOMBASIS.
         self.basislibrary = True
-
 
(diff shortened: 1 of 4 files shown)


cclib/parser/daltonparser.py
```diff
@@ -48,14 +48,13 @@ def before_parsing(self):
         # would like to use as triggers.
         self.section = None
 
-        # If there is no symmetry, assume this.
-        self.symlabels = ['Ag']
+        # If there is no symmetry (point group is C1), assume this.
```
langner, line 51: Could you expand this comment? It's not clear what "assume this" means for the reset of the code.

langner (approved):

langner:
From discussion: don't populate with C1 if not parsed.

shivupa (approved):

shivupa:
I think this just needs a rebase, but I can try to resolve conflicts later tonight and get this in

berquist:
Nope, I've done it, also I need to check if

> From discussion: don't populate with C1 if not parsed.

has been addressed.

Issue #843, opened 2020-04-21
Bridge for ASE trajectory files


schneiderfelipe:
I would like to build a bridge from cclib to the [Atomic Simulation Environment (ASE)](https://wiki.fysik.dtu.dk/ase/index.html). That would consist of a small module (`cclib.cclib.bridge.cclib2ase`) with support of creating `ase.Atoms` objects and writing ASE trajectory files. ASE would be kept as a soft dependency, of course.

`ase.Atoms` objects hold only structural data (atomic numbers, masses, geometries, etc.), but [trajectory files](https://wiki.fysik.dtu.dk/ase/ase/io/trajectory.html) can store a [small set of calculated properties](https://gitlab.com/ase/ase/-/blob/master/ase/calculators/calculator.py#L118), such as energies, gradients, dipoles, charges, and free energies. Those properties would be taken from cclib whenever available.

Such a bridge would make cclib interoperable with any package expecting ASE trajectory files as data sources. For example, one could directly use data read with cclib (e.g., geometries, energies, and gradients) to train neural networks for fast energy and gradient predictions, as the [Atomistic Machine-learning Package (Amp)](https://amp.readthedocs.io/en/latest/) expects train data in the ASE trajectory format.

I would be happy to submit a PR if you believe that it is something interesting/useful.

berquist:
Oops, fat fingered that one...yes, that would be a really nice feature to have.

langner:
Indeed, this would be nice.

berquist:
Done in #845

PR #845, opened 2020-04-22
Adding a bridge module to the Atomic Simulation Environment (ASE)

schneiderfelipe:
This PR solves #843. It currently implements conversions from ASE `Atoms` objects to `ccData` and vice-versa. Structural data (atomic numbers, masses, geometries, etc.) in ASE `Atoms` objects is both read and written. Still, calculated properties (potential energy, forces, dipole moment, etc.) depend on whether a calculator is attached to `Atoms`. Since the conversion from `ccData` to `Atoms` doesn't attach any calculator, not all data are written both ways, so we have this limitation.

I assumed potential energies coming from ASE are equivalent to `scfenergies` in cclib. We might want to change that in the future and sum dispersion corrections and other contributions when converting to ASE.

I made an effort to ensure units are properly converted when data goes back and forth. Please let me know if you spot any errors in the units.


schneiderfelipe:
I pushed the implementation for reading/writing trajectory files. There's no need for an attached calculator for properties to be stored in `.traj` files.

The following entries are written, when available: `atomcoords`, `atomnos`, `atomcharges`, `atomspins`, `atommasses`, `scfenergies`, `grads`, `moments` (dipole only) and `freeenergy`. No `charge` or `mult` is written since ASE calculates them from `atomcharges` and `atomspins` (see [here](https://gitlab.com/ase/ase/-/blob/master/ase/io/gaussian.py#L105) for an example).

cclib/bridge/cclib2ase.py
```diff
@@ -0,0 +1,238 @@
...
+            ccdata.atommasses,
+        )
+
+        properties = {}
+        if hasattr(ccdata, "scfenergies"):
+            properties.update({"energy": ccdata.scfenergies[i]})
+        if hasattr(ccdata, "grads"):
+            try:
+                properties.update(
+                    {"forces": -ccdata.grads[i] * units.Hartree / units.Bohr}
```
schneiderfelipe, line 97: This unit conversion is in line with [the one done in the ORCA calculator](https://gitlab.com/ase/ase/-/blob/master/ase/calculators/orca.py#L123).

cclib/bridge/cclib2ase.py
```diff
@@ -0,0 +1,238 @@
...
+                )
+            except IndexError:
+                pass
+
+        if i == len(ccdata.atomcoords) - 1:  # last geometry
+            if hasattr(ccdata, "moments"):
+                properties.update({"dipole": ccdata.moments[1] * units.Bohr})
```
schneiderfelipe, line 104: [ASE's default units](https://wiki.fysik.dtu.dk/ase/ase/units.html) are eV and angstrom, so this needs to be converted from a.u. ([in line with the XTB calculator](https://github.com/grimme-lab/xtb/blob/master/python/xtb/calculators.py#L261)).

schneiderfelipe:
Travis is failing for Python 2.7 for the following reason:

```
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Collecting https://sourceforge.net/projects/pyquante/files/PyQuante-1.6/PyQuante-1.6.5/PyQuante-1.6.5.tar.gz (from -r requirements.txt (line 15))
  Downloading https://sourceforge.net/projects/pyquante/files/PyQuante-1.6/PyQuante-1.6.5/PyQuante-1.6.5.tar.gz (681kB)
Collecting packaging>=19.0 (from -r requirements.txt (line 1))
  Using cached https://files.pythonhosted.org/packages/62/0a/34641d2bf5c917c96db0ded85ae4da25b6cd922d6b794648d4e7e07c88e5/packaging-20.3-py2.py3-none-any.whl
Requirement already satisfied: numpy in /home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages (from -r requirements.txt (line 2)) (1.16.2)
Requirement already satisfied: six in /home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages (from -r requirements.txt (line 4)) (1.11.0)
Collecting scipy (from -r requirements.txt (line 7))
  Using cached https://files.pythonhosted.org/packages/24/40/11b12af7f322c1e20446c037c47344d89bab4922b8859419d82cf56d796d/scipy-1.2.3-cp27-cp27mu-manylinux1_x86_64.whl
Collecting periodictable (from -r requirements.txt (line 8))
Collecting ase (from -r requirements.txt (line 11))
  Downloading https://files.pythonhosted.org/packages/95/c0/4a38c15ec0e4d8fc7d470d9e02cabe7a84ee0c4dd6bfe761eb553510e49d/ase-3.19.1.tar.gz (1.9MB)

    Complete output from command python setup.py egg_info:
    Python 3.5 or later is required
```

shivupa:
> Should I test for Python 3.5+?

On Travis the 3.5+ tests builds are passing. This might need a longer timeline for a cclib v2 release (#518)

schneiderfelipe:
> > Should I test for Python 3.5+?
> 
> On Travis the 3.5+ tests builds are passing. This might need a longer timeline for a cclib v2 release (#518)

What about doing the same thing that is [done for PyQuante in `requirements.txt`](https://github.com/cclib/cclib/blob/df9af5126ffb139a8505f4e7d8551788f5b345dc/requirements.txt#L14) (the general case is defined in [PEP 508](https://www.python.org/dev/peps/pep-0508/#environment-markers))? As such, the tests would only run under Python 3.5+.

shivupa:
> > > Should I test for Python 3.5+?
> > 
> > 
> > On Travis the 3.5+ tests builds are passing. This might need a longer timeline for a cclib v2 release (#518)
> 
> What about doing the same thing that is [done for PyQuante in `requirements.txt`](https://github.com/cclib/cclib/blob/df9af5126ffb139a8505f4e7d8551788f5b345dc/requirements.txt#L14) (the general case is defined in [PEP 508](https://www.python.org/dev/peps/pep-0508/#environment-markers))? As such, the tests would only run under Python 3.5+.

Oh yes I understand now! That is a better solution

schneiderfelipe:
> Oh yes I understand now! That is a better solution

@shivupa Done! :)

cclib/bridge/cclib2ase.py
```diff
@@ -0,0 +1,243 @@
...
+    return Atoms(
+        positions=atomcoords,
+        numbers=atomnos,
+        masses=atommasses,
+        charges=atomcharges,
+        magmoms=atomspins,
+    )
+
+
+def writetraj(filename, ccdata, popname="mulliken", index=None):
```
langner, line 54: Perhaps call this `write_trajectory`? I don't think the abbreviation is valuable here.

cclib/bridge/cclib2ase.py
```diff
@@ -0,0 +1,243 @@
...
+                properties.update({"dipole": ccdata.moments[1] * units.Bohr})
+            if hasattr(ccdata, "free_energy"):
+                properties.update(
+                    {"free_energy": ccdata.freeenergy * units.Hartree}
+                )
+
+        traj.write(atoms, **properties)
+
+
+def readtraj(filename):
```
langner, line 118: Same for this and other methods.

test/bridge/testase.py
```diff
@@ -0,0 +1,360 @@
...
+        assert np.isclose(data.natom, 2)
+        assert np.isclose(data.charge, 0)
+        assert np.isclose(data.mult, 1)
+        assert np.isclose(data.temperature, 0)
+
+        # front page example of ASE (https://wiki.fysik.dtu.dk/ase/)
+        h2.calc = Psi4(
+            label="h2", method="PBE", basis="3-21G", num_threads="max"
+        )
+        opt = GPMin(h2)  # optimize geometry
```
langner, line 44: Why do we need to optimize a geometry to test code that converts data structures?

schneiderfelipe: No need actually, I just wanted to ensure the returned object would work as expected. Should we remove it? I could make tests calculate energies and forces only.

cclib/bridge/cclib2ase.py
schneiderfelipe: Sure, done!

cclib/bridge/cclib2ase.py
schneiderfelipe: I changed both to `write_trajectory` and `read_trajectory`. The other three functions (`_check_ase`, `makease` and `makecclib`) are in line with how they are called in other bridges (for instance, [`cclib2openbabel._check_openbabel`](https://github.com/cclib/cclib/blob/master/cclib/bridge/cclib2openbabel.py#L23), [`cclib2openbabel.makeopenbabel`](https://github.com/cclib/cclib/blob/master/cclib/bridge/cclib2openbabel.py#L50), etc.). Please let me know if this needs to change as well.

test/bridge/testase.py
schneiderfelipe: > Why do we need to optimize a geometry to test code that converts data structures?

Removed all optimizations.

test/bridge/testase.py
```diff
@@ -0,0 +1,236 @@
...
+        assert np.allclose(
+            trajdata.atomcharges["mulliken"], data.atomcharges["mulliken"]
+        )
+        # no atomspins here
+
+    def test_write_and_read_opt_trajectories(self):
+        """Ensure write and read trajectory files with optimizations."""
+        # geometry optimization
+        data = ccopen("data/ORCA/basicORCA4.2/dvb_gopt.out").parse()
+        cclib2ase.write_trajectory("dvb_gopt.traj", data)
```
langner, line 212: Do you have any native ASE trajectory you could use for testing here?

schneiderfelipe: I generated a simple optimization trajectory and included in the tests. The trajectory file was generated as follows:

```python3
from ase import Atoms
from ase.optimize import BFGS
from ase.calculators.orca import ORCA

h2o = Atoms("OH2", positions=[[0, 0, 0], [1, 0, 0], [0, 1, 0]])
h2o.calc = ORCA(orcasimpleinput="PM3")
opt = BFGS(h2o, trajectory="h2o.traj")
opt.run(fmax=0.02)
# =>
#       Step     Time          Energy         fmax
# BFGS:    0 16:00:28     -324.613009        4.3722
# BFGS:    1 16:00:28     -324.787779        1.4193
# BFGS:    2 16:00:28     -324.853574        1.0747
# BFGS:    3 16:00:29     -324.903836        0.4422
# BFGS:    4 16:00:29     -324.907186        0.1344
# BFGS:    5 16:00:29     -324.907382        0.0226
# BFGS:    6 16:00:30     -324.907387        0.0000
```

langner:
Sorry for the delay. I haven't really dug into this much, but at a glance it looks good. I wonder @berquist and @shivupa are you OK with merging this now for v1.6.3? I am.

langner (approved):

test/bridge/testase.py
```diff
@@ -0,0 +1,261 @@
...
+            ase_data.atomcharges["mulliken"], data.atomcharges["mulliken"]
+        )
+        assert np.allclose(
+            ase_data.atomspins["mulliken"], 0
+        )  # spin densities not supported but overwritten by EMT
+        assert np.isclose(ase_data.charge, data.charge)
+        assert np.isclose(ase_data.mult, 1)
+        assert np.isclose(ase_data.natom, len(data.atomnos))
+        assert np.isclose(ase_data.temperature, 0)
+        assert np.allclose(ase_data.scfenergies, [7.016800805424298])
```
berquist, line 98: Can you comment that this value and the gradients came from the EMT calculation? I was confused by this at first.

cclib/bridge/cclib2ase.py
```diff
@@ -0,0 +1,243 @@
...
+    attributes["mult"] = ccdata.mult
+    attributes["natom"] = ccdata.natom
+
+    return ccData(attributes)
+
+
+def makecclib(atoms, popname="mulliken"):
+    """Create cclib attributes and return a ccData from an ASE Atoms object.
+
+    Available potential energy is assumed to be from SCF (see
```
berquist, line 176: Can you mention forces/gradients here as well?

test/bridge/testase.py
```diff
@@ -0,0 +1,261 @@
...
+                    [0.05412095, 0.06691358, -0.0],
+                    [-0.01446204, 0.04423546, -0.0],
+                    [0.01446204, -0.04423546, -0.0],
+                    [0.04468942, -0.00502891, -0.0],
+                    [-0.04468942, 0.00502891, -0.0],
+                ]
+            ],
+        )
+        assert (
+            np.abs(dvb_sp_un.get_forces()).sum() > np.abs(ase_data.grads).sum()
```
berquist, line 127: I'm not sure this check is meaningful, since on my machine it's comparing 2.4424906541753444e-15 to -1.5612511283791264e-17 for a quantity itself (EMT gradient) that isn't meaningful for the system. I'd replace both this assert and the full one above with a check for the gradient existing with the correct shape, and *not* being present on the original `data`.

berquist (changes requested):
Yes, I am overall pretty happy with this.

I needed to add
```python
from .bridge.testase import *
```
in `test_bridge.py` behind a version guard to have the tests run.

cclib/bridge/cclib2ase.py
```diff
@@ -0,0 +1,243 @@
...
+    try:
+        attributes["atomspins"] = {popname: atoms.get_magnetic_moments()}
+    except (PropertyNotImplementedError, RuntimeError):
+        attributes["atomspins"] = {
+            popname: atoms.get_initial_magnetic_moments()
+        }
+
+    # the following is how ASE determines charge and multiplicity from initial
+    # charges and initial magnetic moments in its Gaussian interface
+    # (https://gitlab.com/ase/ase/-/blob/master/ase/io/gaussian.py#L105)
```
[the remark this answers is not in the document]
schneiderfelipe: Sure, done.

test/bridge/testase.py
schneiderfelipe: Done!

cclib/bridge/cclib2ase.py
schneiderfelipe: Sure, done.

test/bridge/testase.py
schneiderfelipe: Right, done. Let me know if what I did is what you've expected (I tested for `AttributeError` in the case of the absent `grads` property).

schneiderfelipe:
> Yes, I am overall pretty happy with this.
> 
> I needed to add
> 
> ```python
> from .bridge.testase import *
> ```
> 
> in `test_bridge.py` behind a version guard to have the tests run.

I added that to the commit (the guard matches the version 3.5+ in `requirements.txt`).

requirements.txt
```diff
@@ -9,6 +9,8 @@ periodictable
 
 # bridges
 biopython==1.76
+openbabel==2.4.1
```
berquist, line 12: Our setup is super confusing right now, but the Open Babel line actually needs to be removed; it's what's causing CI to fail.

berquist (approved):

requirements.txt
langner: It eludes me what this is the case, but for the purposes of this PR we can submit once you remove the line and it fixes Travis.

langner (changes requested):

requirements.txt
schneiderfelipe: Right, removed.

langner (approved):

langner:
Thanks again, this is great!

PR #847, opened 2020-04-28
Add parser for Hirshfeld Analysis

GomaruSai:
Parses charges and spins from Hirshfeld Analysis to atomcharges and atomspins.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -1255,6 +1255,21 @@ States  Energy Wavelength    D2        m2        Q2         D2+m2+Q2       D2/TO
         #   are not printed (there is a blank line at the end).
         if line[:22] == "LOEWDIN ATOMIC CHARGES":
             self.parse_charge_section(line, inputfile, 'lowdin')
+        # ------------------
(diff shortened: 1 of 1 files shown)


cclib/parser/orcaparser.py
```diff
@@ -1612,6 +1627,17 @@ def parse_charge_section(self, line, inputfile, chargestype):
         elif chargestype == 'chelpg':
             should_stop = lambda x: x.startswith('---')
             start, stop = 11, 26
+        elif chargestype == 'hirshfeld':
+            # stops when blank line encountered
+            should_stop = lambda x: not bool(x.strip())
+            start, stop = 9, 17
+            for i in range(0, 3):
```
GomaruSai, line 1634: 'i' should vary from 0 to 2 rather than 0 to 3 to take all atoms into account.
The line should instead read: for i in range(0, 2):

GomaruSai:
I would like to apologize for a mistake I have made in the code. I have commented at the line which requires the change. I shall close this PR and make a new one with the change.
Sorry for the hassle caused by this.
I confirm that the new one gives the desired output.

PR #753, opened 2019-08-08
Gaussian zero-point correction

kuriba:
Parsing support for zero-point correction from Gaussian frequency calculations.

PR #854, opened 2020-05-02
Gaussian zpe

langner:
This fixes the typos and stuff in #753 so that we can merge it in for v1.6.3. I didn't bother cleaning up the end line noise, we can do that later.

Note: Travis will fail until #853 is merged in.


shivupa:
Travis looks clean.

PR #857, opened 2020-05-07
[Gaussian] Revert 'Elmar Neumann' exception for which there is no regression

### LINKED ISSUE #756: Gaussian natural orbital parsing
Hi cclib community, when I am trying to parse my gaussian ub3lyp calculation with `pop=naturalorbitals` enabled, I get the following error:

```
~/opt/cclib/cclib/parser/gaussianparser.py in natural_orbital_single_spin_parsing(inputfile, updateprogress_title)
   1513                                 atombasis.append(basisonatom)
   1514                             basisonatom = []
-> 1515                             atomname = "%s%s" % (parts[2], parts[1])
   1516                         orbital = line[11:20].strip()
   1517                         aonames.append("%s_%s" % (atomname, orbital))

IndexError: list index out of range
```
when I revert the "Elmar Neumann's example" change back to `:12` in the following code
```
1507                    if base == 0:
1508                        # Changed below from :12 to :11 to deal with Elmar Neumann's example.
1509                        parts = line[:11].split()
1510                        # New atom.
1511                        if len(parts) > 1:
1512                            if i > 0:
1513                                atombasis.append(basisonatom)
1514                            basisonatom = []
1515                            atomname = "%s%s" % (parts[2], parts[1])
1516                        orbital = line[11:20].strip()
```
everything works.

Here's the log file [g.log](https://github.com/cclib/cclib/files/3516989/g.log)

langner:
So this little change fixes #756 as pointed out there. Since we don't have any regression as far as I can tell from Elmar Neumann - or at least this change doesn't break anything we have - I think it's fair to ignore the old warning. If something comes up that we can debug, we can look for a better fix.

### DIFF
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -1595,8 +1595,7 @@ class Gaussian(logfileparser.Logfile):
                     line = next(inputfile)
                     # Just do this the first time 'round.
                     if base == 0:
-                        # Changed below from :12 to :11 to deal with Elmar Neumann's example.
-                        parts = line[:11].split()
+                        parts = line[:12].split()
                         # New atom.
                         if len(parts) > 1:
                             if i > 0:


langner:
This needs to go in before cclib/cclib-data#132

PR #858, opened 2020-05-07
Log last parsed line after encountering an exception during parsing

langner:


langner:
This is a little improvement I fond useful when debugging some recent regressions.

PR #859, opened 2020-05-14
Add support for ORCA geometry scans (unrelaxed and relaxed)

shivupa:
I added support for a parameter scan calculation for ORCA. 

I would appreciate input/advice on the following points:

The ORCA parser was also adding coordinates into the metadata which seemed wrong to me. I removed this. Is that reasonable?

Also `scancoords` and `scanenergies` seem redundant to me? The gaussian parser seems to support this since `scancoords` are set from `atomcoords`. Also if the energies should be stored in `scfenergies`, `mpenergies` etc.


shivupa:
I'll add a example output tomorrow!

shivupa:
`scanenergies` and `scancoords` are only redundant in the _unrelaxed_ scan case. In the _relaxed_ scan case, the `scfenergies` will contain energies for each of the optimization steps for each step in the parameters. I believe `scancoords` and `scanenergies` should only contain the 'final' geometries and energies at each parameter value.

`testScan.py` doesn't seem to be testing this?

langner:
> `scanenergies` and `scancoords` are only redundant in the _unrelaxed_ scan case. In the _relaxed_ scan case, the `scfenergies` will contain energies for each of the optimization steps for each step in the parameters. I believe `scancoords` and `scanenergies` should only contain the 'final' geometries and energies at each parameter value.
> 
> `testScan.py` doesn't seem to be testing these parsed values?

I believe that's accurate. Could you add this information to the docs right away?

cclib/parser/orcaparser.py
```diff
@@ -144,54 +144,77 @@ def extract(self, inputfile, line):
                 elif line[0] == '%':
                     pass
 
-                # Geometry block
```
berquist, line 147: Where did this block run off to?

shivupa: Sorry I should have put a comment about this. This block parses the geometry _only_ for storing in the metadata. This isn't done in any of the other parsers, and I didn't see the use of it so I removed it. If this is something that is desired, then I could add it back in. It would have to be generalized a bit since it is incompatible with parameter scan calculations.

shivupa:
> The ORCA parser was also adding coordinates into the metadata which seemed wrong to me. I removed this. Is that reasonable?

This is the only thing remaining on this PR. Does anyone have thoughts on this?

berquist:
I think we *do* want to have the input coordinates stored in the metadata, because
- we're moving to have the input file parse into `metadata`, and
- the input coordinates may be different from the coordinates we parse, even in a single-point calculation, such as if they are rotated or otherwise altered due to symmetry.

langner:
@shivupa Is there much to do here before it can go in?

shivupa:
> @shivupa Is there much to do here before it can go in?

Nope! @berquist does this look good? I added the parsing the geometry for the metadata back in

langner (approved):

berquist:
Yes, I am happy with the implementation, but I just noticed that there aren't any tests. What do we want to do about them?

langner:
> Yes, I am happy with the implementation, but I just noticed that there aren't any tests. What do we want to do about them?

Let's wait for 1.7.1 then.

berquist:
Poke for tests.

shivupa:
This works as is, but it would be nice to have `scanenergies` parsed for ORCA eventually.

berquist (approved):

PR #862, opened 2020-05-24
[Gaussian] Append list instead of map to scanparm

### LINKED ISSUE #851: scanparm is a list of map objects
When importing a Gaussian PES scan, scanparm is a list of map objects. I can't find a clean way to extract a list of scan parameters without `list(map(list, molecule.scanparm))`. I'm sure I'm making that transform more complicated than necessary, but the fact remains that `molecule.scanparm` should consist of a flat list of numbers. 

https://github.com/cclib/cclib/blob/409e7f9ee87c02240b78eb1845177656e844159b/cclib/parser/gaussianparser.py#L977

langner:
Fixed #851


langner:
Included in #864.

PR #863, opened 2020-05-24
[Gaussian] Convert value put into scanenergies

### LINKED ISSUE #850: Gaussian scanenergies aren't converted to eV.
The documentation for cclib states that all energies  are in eV, but the scanenergies values parsed from Gaussian output files don't get converted. I'm guessing that line 976 needs to have a conversion function added to it. 

https://github.com/cclib/cclib/blob/409e7f9ee87c02240b78eb1845177656e844159b/cclib/parser/gaussianparser.py#L976

langner:
Fixed #850 as far as I can tell, although there are not tests for this code yet.


langner:
Included in #864.

PR #864, opened 2020-05-24
[Gaussian] Parse scannames and scanparm for relaxed PES scans, and add more unit tests

### LINKED ISSUE #850: Gaussian scanenergies aren't converted to eV.

langner:
Note that this ~~still only tests relaxed scans~~tests both relaxes and unrelaxed scans (Thanks Shiv!) since the formatting is different.

Also fixes #850 and #851


langner:
OK, this is ready to go in.

langner:
Redirecting this to the v1.6.3 branch, which I'll merge back into master after the release.

PR #866, opened 2020-05-27
Fixes issue with dealing with reduced list of atoms

MDAddario:
## Issue

When working with an atom selection in Gaussian of the form

```
 AtmSel:  Line="noatoms atoms=X-Y"
```

the block of code

```
if line.strip().split()[0:3] == ["Atom", "AN", "X"]:
```

from `gaussianparser.py` will run for `self.natom` iterations.
However, since the number of atoms has been reduced, the block of code will read too many lines and cause an error:

```
ValueError: could not convert string to float: 'A'
```

## Fix

To fix this, the parameter `NQMF` is directly read from the .out file and stored as an instance attribute.
This parameter corresponds to the number of reduced atoms.
This is then used when iterating in a 

```
if line.strip().split()[0:3] == ["Atom", "AN", "X"]:
```

block of code, allowing for the correct number of lines to be read.

### DIFF
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -272,8 +272,9 @@ class Gaussian(logfileparser.Logfile):
 
             # Necessary for `if line.strip().split()[0:3] == ["Atom", "AN", "X"]:` block
             if not hasattr(self, 'nqmf'):
-                nqmf = int(re.search('NQMF=\s*(\d+)', line).group(1))
-                self.set_attribute('nqmf', nqmf)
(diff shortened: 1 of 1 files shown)


langner:
Thanks for the report and fix! I think you need to be a little bit more subtle with the regex, because sometimes it doesn't match which makes the code fail (no `group` attribute). See Travis for details.

langner:
@MDAddario could you rebase with master? Travis was failing and a fix was merged in a few days ago.

langner:
Looks like the source repo was actually deleted. Patched and rebased in #891 

Would be nice to have a regression, but let's not wait on it.

PR #869, opened 2020-06-05
Feat: Bickelhaupt Charges

mscho527:
This adds a new routine `cclib.cclib.method.bpa` which calculates Bickelhaupt charges as described in Bickelhaupt, F. M.; Hommes, N. J. R.; Guerra, C. F.; Baerends, E. The Carbon-Lithium Electron Pair Bond in (CH3Li)_n (n=1,2,4). _Organometallics_ **1996**, _15_ (13), 2923-2931.

Tests are invoked in Travis using inputs from Gaussian 16 and are compared against the values obtained using Multiwfn.


langner:
Is there a package that cclib currently support that already prints these charges, so that we can compare directly against parsed values?

mscho527:
> Is there a package that cclib currently support that already prints these charges, so that we can compare directly against parsed values?

@langner Not that I know of. I'll double check and comment soon.

mscho527:
> > Is there a package that cclib currently support that already prints these charges, so that we can compare directly against parsed values?
> 
> @langner Not that I know of. I'll double check and comment soon.

@langner I think the closest thing that is available among the packages cclib currently support is Gaussian 16's IOp(6/27)=1000 option (as described [here](https://gaussian.com/overlay6/)). This is an option that allows custom weightings for Mulliken calculation. I am not entirely sure how about the syntax of inputting the custom weightings though.

cclib/method/bpa.py
```diff
@@ -0,0 +1,120 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2018, the cclib development team
```
langner, line 3: nit: update to 2020

test/method/testpopulation.py
```diff
@@ -5,7 +5,7 @@
 # This file is part of cclib (http://cclib.github.io) and is distributed under
```
langner, line 5: nit: update copyright year to 2020 :)

test/method/testpopulation.py
```diff
@@ -126,8 +126,54 @@ def testsumspins(self):
         totalspin = sum(self.analysis.fragspins)
         self.assertAlmostEqual(totalspin, formalspin, delta=1.0e-3)
 
-
-tests = [GaussianMPATest, GaussianLPATest, GaussianCSPATest]
+class GaussianBPATest(unittest.TestCase):
+    """Bickelhaupt Population Analysis test"""
+    
+    def setUp(self):
```
langner, line 132: Just to be prudent, add a `super(GaussianBPATest, self).setUp()` call. We should be doing that everywhere.

cclib/method/bpa.py
```diff
@@ -0,0 +1,120 @@
...
+"""Calculation of Bickelhaupt population analysis (BPA) based on data parsed by cclib."""
+
+import random
+
+import numpy
+
+from cclib.method.population import Population
+
+
+class BPA(Population):
```
langner, line 17: I would prefer to have a more verbose name here, perhaps just `Bickelhaput`? I know we've been using abbreviations, but I don't think `BPA` is that common this will likely break down in the future for the more advanced methods. @berquist @shivupa thoughts?

berquist: +1, I also think (separately) that `MPA` should be renamed to Mulliken, etc.

langner: Yup, we should definitely rename and leave aliases to the old abbreviations, which we could drop in 2.x.

mscho527:
I just received a response from Professor Bickelhaupt that clarifies the way weights are calculated for the charges. The way weights are calculated in this function matches what he has explained in the response.

cclib/method/bickelhaupt.py
```diff
@@ -0,0 +1,120 @@
...
+                    if hasattr(self.data, "aooverlaps"):
+                        overlaps = self.data.aooverlaps[a]
+                    # handle spin-unrestricted beta case
+                    elif hasattr(self.data, "fooverlaps2") and spin == 1:
+                        overlaps = self.data.fooverlaps2[a]
+
+                    elif hasattr(self.data, "fooverlaps"):
+                        overlaps = self.data.fooverlaps[a]
+
+                    self.aoresults[spin][i][a] = numpy.sum(numpy.multiply(numpy.multiply(numpy.multiply(w[spin][a], self.data.mocoeffs[spin][i][a]), self.data.mocoeffs[spin][i]), overlaps))
```
langner, line 82: Could you break this up into several line, for better readability?

mscho527: I've just done it. Thanks!

cclib/method/bickelhaupt.py
```diff
@@ -0,0 +1,122 @@
...
+    """Bickelhaupt population analysis."""
+
+    def __init__(self, *args):
+
+        # Call the __init__ method of the superclass.
+        super(Bickelhaupt, self).__init__(logname="Bickelhaupt Population Analysis", *args)
+
+    def __str__(self):
+        """Return a string representation of the object."""
+        return "Bickelhaupt charges of %s" % (self.data)
```
shivupa, line 27: Can this (and others) be updated to the new string formatting (e.g. `"Bickelhaupt charges of {}".format(self.data)`)

cclib/method/bickelhaupt.py
```diff
@@ -0,0 +1,122 @@
...
+        step = 0
+
+        # First determine the weight matrix
+        for spin in range(len(self.data.mocoeffs)):
+            for i in range(self.data.nbasis):
+                if self.progress and random.random() < fupdate:
+                    self.progress.update(step, "Bickelhaupt Population Analysis")
+                for j in range(self.data.nbasis):
+                    # W_{ab} = 2 * \sum_k c_{ak}^2 / (\sum_i c_{ai}^2 + \sum_j c_{bj}^2)
+                    # W is (nbasis)x(nbasis)
```
shivupa, line 63: Can you clarify this to "The size of W..."

doc/sphinx/methods.rst
```diff
@@ -144,6 +144,66 @@ The LPA class available from cclib.method performs Löwdin population analysis a
     m = LPA(d)
     m.calculate()
 
+.. index::
+    single: methods; Bickelhaupt Population Analysis
+
+Bickelhaupt Population Analysis
+-------------------------------
+
+The Bickelhaupt class available from cclib.method performs Bickelhaupt population analysis that has been proposed in *Organometallics* 1996, 15, 13, 2923–2931. `doi:10.1021/om950966x <https://pubs.acs.org/doi/abs/10.1021/om950966x>`_
```
shivupa, line 153: alternatively the link could point to [https://doi.org/10.1021/om950966x](https://doi.org/10.1021/om950966x). Doesn't need to change just a tip if you come across a DOI with no link in the future

cclib/method/bickelhaupt.py
mscho527: @shivupa Thanks for the comments! I've updated them just now.

cclib/method/bickelhaupt.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # First determine the weight matrix
+        for spin in range(len(self.data.mocoeffs)):
+            for i in range(self.data.nbasis):
+                if self.progress and random.random() < fupdate:
+                    self.progress.update(step, "Bickelhaupt Population Analysis")
+                for j in range(self.data.nbasis):
+                    # W_{ab} = 2 * \sum_k c_{ak}^2 / (\sum_i c_{ai}^2 + \sum_j c_{bj}^2)
+                    # The dimension of W is (nbasis)x(nbasis)
+                    aTerm = numpy.dot(self.data.mocoeffs[spin][0:(self.data.homos[spin]+1),i], self.data.mocoeffs[spin][0:(self.data.homos[spin]+1),i])
```
langner, line 64: Can you break this up into several lines as well (use variables for the indices)?

cclib/method/bickelhaupt.py
```diff
@@ -0,0 +1,122 @@
...
+
+        step = 0
+
+        # First determine the weight matrix
+        for spin in range(len(self.data.mocoeffs)):
+            for i in range(self.data.nbasis):
+                if self.progress and random.random() < fupdate:
+                    self.progress.update(step, "Bickelhaupt Population Analysis")
+                for j in range(self.data.nbasis):
+                    # W_{ab} = 2 * \sum_k c_{ak}^2 / (\sum_i c_{ai}^2 + \sum_j c_{bj}^2)
```
langner, line 62: Perhaps move this to the comment at the top of the block, and perhaps also add a short description of the weight matrix.

cclib/method/bickelhaupt.py
```diff
@@ -0,0 +1,122 @@
...
+                if spin == 0:
+                    alpha = numpy.add(alpha, temp)
+                elif spin == 1:
+                    beta = numpy.add(beta, temp)
+
+        if not unrestricted:
+            self.fragcharges = numpy.multiply(self.fragcharges, 2)
+        else:
+            self.logger.info("Creating fragspins: array[1]")
+            self.fragspins = numpy.subtract(alpha, beta)
```
berquist, line 120: I'd replace all usage of `add`, `subtract`, and `multiply` with their overloaded versions (`+`, `-`, `*`). I always need to look up whether or not the named functions are elementwise or something else weird...

cclib/method/bickelhaupt.py
```diff
@@ -0,0 +1,122 @@
...
+                    # handle spin-unrestricted beta case
+                    elif hasattr(self.data, "fooverlaps2") and spin == 1:
+                        overlaps = self.data.fooverlaps2[a]
+
+                    elif hasattr(self.data, "fooverlaps"):
+                        overlaps = self.data.fooverlaps[a]
+                    
+                    temp = numpy.multiply(w[spin][a], self.data.mocoeffs[spin][i][a])
+                    temp = numpy.multiply(temp, self.data.mocoeffs[spin][i])
+                    self.aoresults[spin][i][a] = numpy.sum(numpy.multiply(temp, overlaps))
```
berquist, line 84: Do you think we can drop `temp` from these three lines, and just do it all in one go? It'll look a lot more like the equation then.

cclib/method/bickelhaupt.py
berquist: Since this is a fresh file, what do you think about running Black on it with a line length of 100?

cclib/method/__init__.py
```diff
@@ -19,4 +19,5 @@
 from cclib.method.nuclear import Nuclear
 from cclib.method.opa import OPA
 from cclib.method.orbitals import Orbitals
+from cclib.method.bickelhaupt import Bickelhaupt
```
berquist, line 22: Can you put this in alphabetical order?

doc/sphinx/methods.rst
```diff
@@ -144,6 +144,66 @@ The LPA class available from cclib.method performs Löwdin population analysis a
...
+
+The contribution of the a-th atomic orbital to the i-th molecular orbital in this method is written in terms of the molecular orbital coefficients, c, and the overlap matrix, S:
+
+.. math:: \Phi_{ai,\alpha} = \sum_b w_{ab,\alpha} c_{ai,\alpha} c_{bi,\alpha} S_{ab}
+
+where the weights :math:`w_{ab}` that are applied on the Mulliken atomic orbital contributions are defined as:
+
+.. math:: w_{ab,\alpha} = 2 \frac{\sum_k c_{ak,\alpha}^2}{\sum_i c_{ai,\alpha}^2 + \sum_j c_{bj,\alpha}^2}
```
berquist, line 161: Related to https://github.com/cclib/cclib/pull/869#issuecomment-642323035, it is not immediately obvious to me how equation 11 in the paper is converted to this. Could you at least say that this leads to the same results?

cclib/method/bickelhaupt.py
shivupa: I'd be in support of this!

cclib/method/bickelhaupt.py
langner: I was the one who asked to break this up. I'm OK with merging back into one expression, as long as the formatting is done better. Another option would be to use better variable names (do those intermediate products have names?).

mscho527: I've merged it back to one expression after replacing `numpy.multiply` with `*` and running `Black` as suggested by @berquist . The whole expression fits in one line (of length 100) this way. Let me know if using different variable names and splitting into 2-3 lines would be more readable.

cclib/method/bickelhaupt.py
```diff
@@ -0,0 +1,138 @@
...
+        return "Bickelhaupt({})".format(self.data)
+
+    def calculate(self, indices=None, fupdate=0.05):
+        """Perform a Bickelhaupt population analysis."""
+        # Bickelhaupt population analysis uses the relative magnitude of the diagonal terms to
+        # partition off-diagonal terms. The weights are therefore calculated using the following
+        # formula:
+        # W_{ab} = 2 * \sum_k c_{ak}^2 / (\sum_i c_{ai}^2 + \sum_j c_{bj}^2)
+        # The dimension of W is (nbasis)x(nbasis).
+        # Then, each terms are calculated similarly to Mulliken charges, in a form of
```
langner, line 40: ...each term is...

cclib/method/bickelhaupt.py
berquist: > do those intermediate products have names?

The quantity that isn't multiplied by the weight is the "unsummed" Mulliken population over one index (don't have it right in front of me), but that's not a very useful intermediate.

PR #873, opened 2020-06-10
Add horton 2 --> cclib bridge, along with tests

mscho527:
Related Issue : #880

This PR is related to Issue #880 and builds upon #882 . This PR adds a bridging function that constructs ccData object from horton 2's IOData object.

Test compares the parsed results (from cclib.io) and converted results (IOData converted using the bridge function).

### DIFF
--- a/cclib/bridge/cclib2horton.py
+++ b/cclib/bridge/cclib2horton.py
@@ -6,6 +6,7 @@
 # the terms of the BSD 3-Clause License.
 
 """Bridge for using cclib data in horton (http://theochem.github.io/horton)."""
+# Support for horton 2 will likely be dropped when Python 2 support is dropped from cclib.
 
 import numpy
 from cclib.parser.data import ccData
@@ -13,11 +14,13 @@ from cclib.parser.utils import find_package
 
 # First check horton version
 _old_horton = False
-_found_horton = find_package('horton')
-_found_iodata = find_package('iodata')
+_found_horton = find_package("horton")
+_found_iodata = find_package("iodata")
 
-if _found_horton: # Detect whether horton 2 is present or not
-    try: # Older horton packages do not have __version__, causing exceptions.
+# Detect whether horton 2 is present or not
+if _found_horton:
+    # Older horton packages do not have __version__, causing exceptions.
+    try:
         from horton import __version__
     except:
         _old_horton = True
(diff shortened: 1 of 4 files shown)


cclib/bridge/__init__.py
```diff
@@ -21,5 +21,13 @@
...
 
-
+if find_package("horton"):
+    try:
+        from horton import __version__
+    except Exception:
+        pass
+    if (__version__[0] == '2'):
+        from cclib.bridge.cclib2horton import makehorton
+    # TODO: Add something for iodata (Horton 3) later
```
langner, line 31: Could you reference an issue here (maybe create one)?

cclib/bridge/cclib2horton.py
```diff
@@ -0,0 +1,92 @@
...
+    _old_horton = False
+    # TODO: Add appropriate routines for Horton3 IOData
+    
+def _check_horton(found_horton, found_iodata, old_horton):
+    if (old_horton):
+        raise ImportError("You must have at least version 2 of `horton` to use this function.")
+    elif (not found_horton and not found_iodata):
+        raise ImportError("You must install `horton` to use this function.")
+
+def makehorton(atomcoords = None, atomnos = None, mocoeffs = None, mult = None, coreelectrons = None, atomcharges = None, polarizability = None):
```
langner, line 37: Would it be better to just pass in a ccData object?

mscho527: @langner I was thinking of doing that (and actually wrote the first version this way) but then found out that all other bridges are implemented this way. I chose to keep it consistent with other bridge methods, but I see that passing in a `ccData` object would likely be cleaner. I am open to making modifications so that `ccData` object is passd instead.

cclib/bridge/__init__.py
mscho527: @langner I am planning to add bridge for horton 3 today or tomorrow. Thanks!

cclib/bridge/cclib2horton.py
```diff
@@ -0,0 +1,145 @@
...
+                attributes['mo'].coeffs.append(mocoeffs[1].T)
+        # if multiplicity known; float / should not be set when mocoeffs present
+        if (mult != None and not isinstance(mocoeffs, numpy.ndarray)):
+            attributes['spinpol'] = mult - 1 # horton has 2S+1, iodata has 2S
+        # if pseudopotentials exist; numpy.ndarray (size natom)
+        if (isinstance(coreelectrons, numpy.ndarray)):
+            attributes['atcorenums'] = coreelectrons
+        # if mulliken charges are known; dict of numpy.ndarrays (size natom)
+        if (isinstance(atomcharges, dict)):
+            attributes['atcharges'] = atomcharges
```
berquist, line 99: Can you add an `else:` that raises an error?

cclib/bridge/cclib2horton.py
```diff
@@ -0,0 +1,145 @@
...
+        if (hasattr(iodat, 'mo')):
+            attributes['mocoeffs'] = [iodat.mo.coeffs[:iodat.mo.norba].T]
+            if iodat.mo.kind == 'unrestricted':
+                attributes['mocoeffs'].append(iodat.mo.coeffs[iodat.mo.norba:].T)
+        if (hasattr(iodat, 'spinpol')):
+            attributes['mult'] = iodat.spinpol + 1 # iodata has 2S, cclib has 2S+1
+        if (hasattr(iodat, 'atcorenums')):
+            attributes['coreelectrons'] = iodat.atcorenums
+        if (hasattr(iodat, 'atcharges')):
+            attributes['atomcharges'] = iodat.atcharges
```
berquist, line 143: Same as above.

cclib/bridge/cclib2horton.py
mscho527: As far as I know, `atomcharges` are not always present in input data? The if statement is supposed to detect whether any valid information was passed on or not (since default value for `atomcharges` is `None`). I see that this may not be the cleanest way to do this -- let me know I could modify this further. Thanks!

cclib/bridge/__init__.py
langner: This would be a good thing to send in a separate PR.

cclib/bridge/cclib2horton.py
langner: I think this might be a good place to not be consistent; IMO passing the object would be better and we just didn't think about it before. We can update the remaining bridges later. @berquist for thoughts

cclib/bridge/cclib2horton.py
langner: This would be more natural if we passed in the ccData object.

cclib/bridge/__init__.py
mscho527: Yes -- I will try to split this into a separate PR tomorrow. The commits for horton 2 and 3 are readily separable, so I think should be able to do so without much issue. Thanks.

cclib/bridge/cclib2horton.py
```diff
@@ -0,0 +1,92 @@
...
+# the terms of the BSD 3-Clause License.
+
+"""Bridge for using cclib data in Horton (http://theochem.github.io/horton)."""
+
+import numpy
+from cclib.parser.data import ccData
+from cclib.parser.utils import find_package
+
+_found_horton = find_package("horton")
+if _found_horton:
```
langner, line 15: Could you add a comment here shortly explaining the difference between horton and the iodata you try to import below?

test/bridge/testhorton.py
```diff
@@ -0,0 +1,104 @@
...
+import numpy
+
+from cclib.bridge import cclib2horton
+from ..test_data import getdatafile
+
+
+class HortonTest(unittest.TestCase):
+    """Tests for the horton bridge in cclib."""
+
+    # Both packages are able to read in fchk
```
langner, line 17: Could you be more specific here? Which "both" packages?

Also, please end sentences in comments with periods.

cclib/bridge/cclib2horton.py
```diff
@@ -0,0 +1,84 @@
...
+if _found_horton: # Detect whether horton 2 is present or not
+    try: # Older horton packages do not have __version__, causing exceptions.
+        from horton import __version__
+    except:
+        _old_horton = True
+    else:
+        if __version__[0] == "2":
+            from horton.io.iodata import IOData
+
+if _found_iodata: # Detect whether iodata (part of horton 3) is present or not; Horton 3 is divided into smaller (sub)packages that each take different functionalities.
```
berquist, line 28: In general, can you put comments like this on the line above of the code they're describing, and also reflow them so they don't go over ~100 characters?

test/bridge/testhorton.py
```diff
@@ -0,0 +1,89 @@
...
+                    from horton import log
+                    
+                    log.set_level(0) # This suppresses horton outputs so that they do not flood the test log.
+                    self._hortonver = 2
+                    self.iodat = IOData.from_file(inputfile)
+        if self._found_iodata:
+            # Add horton 3 import lines
+            pass
+        
+    def compareArrays(self, arr1, arr2, delta = 1.0e-3):
```
berquist, line 48: I think in other parts of the code we use assertion functions from `numpy.testing`. There should be at least one that is "almost equal" with a delta or tolerance that can be specified.

cclib/bridge/cclib2horton.py
```diff
@@ -0,0 +1,84 @@
...
+            attributes['atomcoords'] = [iodat.coordinates] # cclib parses the whole history of coordinates in the list
+        if (hasattr(iodat, 'orb_alpha')):
+            attributes['mocoeffs'] = [iodat.orb_alpha]
+        if (hasattr(iodat, 'orb_beta')):
+            attributes['mocoeffs'].append(iodat.orb_beta)
+        if (hasattr(iodat, 'mulliken_charges')):
+            attributes['atomcharges'] = {'mulliken': iodat.mulliken_charges}
+            if (hasattr(iodat, 'npa_charges')):
+                attributes['atomcharges']['natural'] = iodat.npa_charges
+        elif (hasattr(iodat, 'npa_charges')):
```
berquist, line 78: I feel there must be a better way to handle the case when the dictionary hasn't been made yet, so you don't need the NPA logic in two places.

mscho527: I think 'DefaultDict' would make this part of the code cleaner but I'm not sure if it is worth importing another module for this. I also feel that there must be something else -- I'll look for it and comment later.

test/bridge/testhorton.py
mscho527: > I think in other parts of the code we use assertion functions from `numpy.testing`. There should be at least one that is "almost equal" with a delta or tolerance that can be specified.

Looks like `numpy.testing.assert_array_almost_equal` would do. Thanks!

cclib/bridge/cclib2horton.py
mscho527: > In general, can you put comments like this on the line above of the code they're describing, and also reflow them so they don't go over ~100 characters?

Yes - I've just fixed them and ran `black` on the files that were modified in this PR.

cclib/bridge/cclib2horton.py
```diff
@@ -0,0 +1,101 @@
...
+
+    hortonver = check_horton()
+    attributes = {}
+
+    if hortonver == 2:
+        # For a few attributes, a simple renaming suffices
+        renameAttrs = {
+            "numbers": "atomnos",
+            "ms2": "mult",
+            "pseudo_numbers": "coreelectrons",
```
mscho527, line 73: @mscho527 Note for myself -- `pseudo_numbers` attribute has different syntax (atomic number - ignored e) in `horton`. This should be corrected.

cclib/bridge/cclib2horton.py
berquist: Yes, I agree.

berquist: Actually, this would lead down a path that isn't worthwhile right now, it'd take too much time.

PR #874, opened 2020-06-14
Make sure we have updated pytest{,-cov} under Travis

berquist:
This is to fix the currently broken `master`.


shivupa:
Thank you!

Issue #880, opened 2020-06-19
Bridge between cclib and horton for Hirshfeld-related charges calculation


mscho527:
_This issue is based on the GSoC proposal._

The aim is to create a bridge between `cclib` and `horton` (both 2 and 3) so that the data parsed with `cclib` can be used to perform Hirshfeld-related population analyses implemented in horton, such as Hirshfeld charges, Iterative Stockholder charges, and Minimal Basis Iterative Stockholder charges.

- [x] #881 Add horton into Travis configuration
- [x] #882 Add stencil code that detects horton version
- [x] #873 Add horton 2 --> cclib bridge, along with tests
- [x] #883 Add cclib --> horton 2 bridge, along with tests
- [x] #884 Add horton 3 --> cclib bridge, along with tests
- [x] #879 Add cclib --> horton 3 bridge, along with tests
- [x] #888 Write documentation for the bridge and how the bridge can be utilized

mscho527:
Closing #892 due to incomplete parsing of .log files in horton 3 at the moment.

mscho527:
Closing issue since each to-do list item has been merged.

PR #881, opened 2020-06-19
Add horton in Travis configurations to enable testing

mscho527:
Related Issue : #880 

This PR is the first step in the Issue #880. This adds horton 2 into the list of packages that will be fetched for Python 2.7 test and adds IOData, which is part of horton 3, into the list of packages that will be fetched for Python 3.7 test.

`Cython` is added in Python 3.7 test to satisfy requirements for IOData package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,8 @@ before_install:
   - sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu eoan universe"
   - sudo apt update
   - sudo apt install libopenbabel-dev
-
+  # For Python 3.6+ machine, horton 3 IOData is installed from source. For Python 2.7 machine, horton 2 is installed from conda.
+  # From the documentation, horton 3 works with Python 3.6+ and horton 2 does with Python 2.7
   - |
     if [[ $(echo "$TRAVIS_PYTHON_VERSION >= 3.6" | bc -l) == 1 ]];
     then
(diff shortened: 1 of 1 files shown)


mscho527:
> Correct me if I am wrong, but this will install Horton 2.1.0a1 only correct? The horton 3 interface will be untested?

This will install horton 2 for Python 2.7 machine and will install horton 3 (IOData package specifically) for Python 3.7 machine. horton 2 does not work in Python 3.x.x and horton 3 does not in Python 2.x.x so they are mutually exclusive for the test cases.

.travis.yml
```diff
@@ -28,15 +28,36 @@ before_install:
         conda config --set always_yes yes --set changeps1 no
         conda update -q conda
         conda info -a
-        conda create -q -n p4env python=$TRAVIS_PYTHON_VERSION ci-psi4 psi4 numpy matplotlib jupyter scipy -c psi4/label/dev -c defaults -c conda-forge
+        conda create -q -n p4env python=$TRAVIS_PYTHON_VERSION ci-psi4 psi4 numpy matplotlib jupyter scipy cython -c psi4/label/dev -c defaults -c conda-forge -c anaconda
 
         source activate p4env
         conda install pytest pytest-cov codecov -c conda-forge
         pip install pytest-shutil
+        pip install git+https://github.com/theochem/iodata.git
```
mscho527, line 36: @shivupa This is the line that would trigger installation of Horton 3's iodata package.

langner:
> > Correct me if I am wrong, but this will install Horton 2.1.0a1 only correct? The horton 3 interface will be untested?
> 
> This will install horton 2 for Python 2.7 machine and will install horton 3 (IOData package specifically) for Python 3.7 machine. horton 2 does not work in Python 3.x.x and horton 3 does not in Python 2.x.x so they are mutually exclusive for the test cases.

Minsik, could you please put this information as a comment in the file right above the configuration? We will forget...

.travis.yml
berquist: According to https://github.com/theochem/iodata we can also do
```bash
conda install -c theochem iodata
```
but I'm not sure if we want the one from Anaconda or the master branch. Maybe this is fine.

mscho527: @berquist From what I have tested, `iodata` on conda is old and the name of the attributes have changed since when the latest conda package was built. I think that the `pip` install line can be pegged into a specific commit though, if that fits better.

mscho527:
> > > Correct me if I am wrong, but this will install Horton 2.1.0a1 only correct? The horton 3 interface will be untested?
> > 
> > 
> > This will install horton 2 for Python 2.7 machine and will install horton 3 (IOData package specifically) for Python 3.7 machine. horton 2 does not work in Python 3.x.x and horton 3 does not in Python 2.x.x so they are mutually exclusive for the test cases.
> 
> Minsik, could you please put this information as a comment in the file right above the configuration? We will forget...

I've just added the comment on `.travis.yml` just above the configuration line.

langner:
> I'm happy with this once Karol's request has been addressed.



> I'm happy with this once Karol's request has been addressed.

It's been addressed and since we already have one green light I'm going to merge in. Good work!

PR #883, opened 2020-06-19
Add horton 2 <-- cclib bridge, along with tests

mscho527:
Related Issue : #880

This PR is related to Issue #880 and builds upon #873 . This PR adds a bridging function that constructs horton 2's IOData object from ccData object.

Test compares the parsed results (from horton.io.IOData.from_file) and converted results (ccData object converted using the bridge function).

### DIFF
--- a/cclib/bridge/cclib2horton.py
+++ b/cclib/bridge/cclib2horton.py
@@ -37,7 +37,9 @@ if _found_iodata:
 
 def check_horton():
     if _old_horton:
-        raise ImportError("You must have at least version 2 of `horton` to use this function.")
+        raise ImportError(
+            "You must have at least version 2 of `horton` to use this function."
+        )
     elif not _found_horton and not _found_iodata:
         raise ImportError("You must install `horton` to use this function.")
     if _found_iodata:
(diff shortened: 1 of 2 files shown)


mscho527:
Rebased against master, reflecting changes made on PR #873

mscho527:
Note for myself -- reflow line 22 on `testhorton` [Suggested by @berquist during Zoom session]

cclib/bridge/cclib2horton.py
```diff
@@ -53,7 +55,32 @@ def makehorton(ccdat):
     attributes = {}
```
langner, line 55: So I'm a little confused, because IIUC we want to pass in attribute to this function but we are passing a ccData object. What will you do in the end? For your project, I don't think it matters much.

For long term standardization, created #890

mscho527: @langner I've modified the function so accept `ccData` object after your suggestion on PR #873. Modifying from this to a function that asks for the attributes directly should not be too difficult, if that is the way we will be standardizing the bridge functions.

cclib/bridge/cclib2horton.py
```diff
@@ -53,7 +55,32 @@ def makehorton(ccdat):
...
+            # cclib parses the whole history of coordinates in the list, horton keeps the last one.
+            attributes["coordinates"] = ccdat.atomcoords[-1]
+        if hasattr(ccdat, "mocoeffs"):
+            attributes["orb_alpha"] = ccdat.mocoeffs[0]
+            if len(ccdat.mocoeffs) == 2:
+                attributes["orb_beta"] = ccdat.mocoeffs[1]
+        if hasattr(ccdat, "coreelectrons") and isinstance(
+            ccdat.coreelectrons, numpy.ndarray
+        ):  # type-checked
+            attributes["pseudo_numbers"] = ccdat.atomnos - ccdat.coreelectrons
```
berquist, line 78: If you get rid of the type check, you can do
```suggestion
            attributes["pseudo_numbers"] = ccdat.atomnos - np.asanyarray(ccdat.coreelectrons)
```
which will avoid making a copy if `ccdat.coreelectrons` is already an ndarray, but do the conversion if it isn't.

cclib/bridge/cclib2horton.py
berquist: This is tricky because this bridge can take many more attributes than all the other existing ones. It might be unreasonable to ask users to come up with all of these function arguments and have most default to `None` (or require explicitly setting to `None`). I agree that we shouldn't fuss over it and any changes we want can be made later.

cclib/bridge/cclib2horton.py
mscho527: @berquist Thanks! I didn't know about `asanyarray`. That definitely simplifies and make it more efficient!

cclib/bridge/cclib2horton.py
langner: Sounds good.

PR #884, opened 2020-06-19
Add horton 3 --> cclib bridge, along with tests

mscho527:
Related Issue : #880

This PR is related to Issue #880 and builds upon #883 . This PR adds a bridging function that constructs ccData object from Horton 3 IOData object.

Test compares the parsed results (cclib.io) and converted results (horton 3 IOData object converted into ccData object using the bridge function).

### DIFF
--- a/cclib/bridge/cclib2horton.py
+++ b/cclib/bridge/cclib2horton.py
@@ -37,9 +37,7 @@ if _found_iodata:
 
 def check_horton():
     if _old_horton:
-        raise ImportError(
-            "You must have at least version 2 of `horton` to use this function."
-        )
+        raise ImportError("You must have at least version 2 of `horton` to use this function.")
     elif not _found_horton and not _found_iodata:
         raise ImportError("You must install `horton` to use this function.")
     if _found_iodata:
@@ -48,7 +46,7 @@ def check_horton():
         return 2
 
 
-def makehorton(ccdat):
+def makehorton(data):
     """ Create horton IOData object from ccData object """
 
     hortonver = check_horton()
@@ -56,7 +54,7 @@ def makehorton(ccdat):
 
     if hortonver == 2:
         renameAttrs = {"atomnos": "numbers", "mult": "ms2", "polarizability": "polar"}
-        inputattrs = ccdat.__dict__
+        inputattrs = data.__dict__
 
         attributes = dict(
             (renameAttrs[oldKey], val)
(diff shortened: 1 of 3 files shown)


mscho527:
Rebased against master (reflecting changes in PR #883)

cclib/bridge/cclib2horton.py
```diff
@@ -72,10 +72,8 @@ def makehorton(ccdat):
             attributes["orb_alpha"] = ccdat.mocoeffs[0]
             if len(ccdat.mocoeffs) == 2:
                 attributes["orb_beta"] = ccdat.mocoeffs[1]
-        if hasattr(ccdat, "coreelectrons") and isinstance(
-            ccdat.coreelectrons, numpy.ndarray
-        ):  # type-checked
-            attributes["pseudo_numbers"] = ccdat.atomnos - ccdat.coreelectrons
+        if hasattr(ccdat, "coreelectrons"):
```
langner, line 75: Could we rename `ccdat` to just `data`?

cclib/bridge/cclib2horton.py
```diff
@@ -126,7 +124,26 @@ def makecclib(iodat):
...
+            attributes["mocoeffs"] = [iodat.mo.coeffs[: iodat.mo.norba].T]
+            if iodat.mo.kind == "unrestricted":
+                attributes["mocoeffs"].append(iodat.mo.coeffs[iodat.mo.norba :].T)
+        if hasattr(iodat, "spinpol"):
+            # IOData stores 2S, ccData stores 2S+1.
+            attributes["mult"] = iodat.spinpol + 1
+        if hasattr(iodat, "atnums"):
+            attributes["atnums"] = iodat.atnums
+        if hasattr(iodat, "atcorenums"):
+            attributes["coreelectrons"] = iodat.atnums - iodat.atcorenums
```
langner, line 146: I'm not familiar with Horton, but `atcorenums` sounds like it could be the number of core electrons. Is that correct?

cclib/bridge/cclib2horton.py
```diff
@@ -126,7 +124,26 @@ def makecclib(iodat):
...
+        if hasattr(iodat, "atcharges"):
+            attributes["atomcharges"] = iodat.atcharges
```
langner, line 148: In cclib, `atomcharges` is a dictionary - is that what it ends up here?

cclib/bridge/cclib2horton.py
mscho527: Yes, I've just renamed it to `data`. Thanks!

cclib/bridge/cclib2horton.py
mscho527: Yes, in `horton`, it looks like # of electrons after applying the pseudopotentials are stored (from what is parsed using its `from_file` method). `cclib` stores # of electrons that are approximated out with the pseudopotentials. I've added a comment right above the line to reduce confusion. Thanks!

cclib/bridge/cclib2horton.py
mscho527: Yes. Both packages use `dict` here.

PR #887, opened 2020-06-23
Proposing wrapper methods for population analysis methods from horton

mscho527:
Related Issue: #880 

This is a draft PR intended to ask for input on whether it would be necessary to have wrapper methods that basically take care of the steps needed to calculate population analysis methods in `horton`, using the bridge function. An alternative is to provide a similar example code in the documentation.

If we decide to create a wrapper, some decisions may have to be made on whether a parent class for `cclib.method.Population` needs to be created. (b/c some charges are not based on the basis sets and will not have `aoresults`.)

Implementation not intended to work / tested at the moment

### DIFF
--- a/cclib/bridge/__init__.py
+++ b/cclib/bridge/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2017, the cclib development team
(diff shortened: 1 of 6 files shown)


mscho527:
Closed for now to do a few more tests myself on the `ProAtomDB` object in `horton`. Will be updated soon

PR #891, opened 2020-06-24
Fixes issue with dealing with reduced list of atoms

langner:
This is a rebased version of #866

### DIFF
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -281,7 +281,9 @@ class Gaussian(logfileparser.Logfile):
             if not hasattr(self, 'nqmf'):
                 match = re.search('NQMF=\s*(\d+)', line)
                 if match is not None:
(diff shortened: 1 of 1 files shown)


langner:
Hmm.. I restricted the code to scenarios where NQMF>0, because some unit tests were failing. I worry that this might not correspond to the initial edge case the original PR was supposed to cover. But, there is no reference regression logfile, so I'm comfortable doing this.

shivupa:
Do we want to generate a regression test for this?

langner:
> Do we want to generate a regression test for this?

I didn't plan to; if you have time, then that'd be great. Maybe @MDAddario can send one.

MDAddario:
I unfortunately cannot provide a reference logfile. This is why I was unable to pursue the original PR. If ever I am able to get a sample logfile in the future I will update you accordingly.

PR #879, opened 2020-06-18
Add cclib --> horton 3 bridge, along with tests

mscho527:
Related Issue : #880

This PR is related to Issue #880 and builds upon #884 . This PR adds a bridging function that constructs Horton 3 IOData object from ccData object.

Test compares the parsed results (iodata.api.load_one) and converted results (horton 3 IOData object converted from ccData object using the bridge function).

PR #892, opened 2020-06-26
Improve unit tests for horton bridge

mscho527:
Related Issue : #880

This PR is related to Issue #880 and builds upon #879 . This PR attempts to improve unit tests in `horton` bridge.

WIP


mscho527:
Comment for future reference -- mixing use of `fchk` and `log` does not help as two files often have atom coordinates with different reference points.

mscho527:
`.log` IO in horton 3 seems not to be ready yet (it does not import some major attributes like atom numbers). Closing this for now.

Issue #886, opened 2020-06-19
Add method to calculate Bader's QTAIM charges


mscho527:
_This issue is based on the GSoC proposal._

Implementation of Bader's QTAIM charges will add modern charge analysis tool for `cclib`. Grid-based integration is also need for determining QTAIM charges.

**References**

1. Calculation of atomic integration data. _J. Computational Chem._ ([Journal](https://onlinelibrary.wiley.com/doi/abs/10.1002/1096-987X%28200009%2921%3A12%3C1040%3A%3AAID-JCC2%3E3.0.CO%3B2-8))
2. Bader Analysis: Calculating the Charge on Individual Atoms in Molecules & Crystals ([Presentation](https://notendur.hi.is/egillsk/stuff/annad/egillsk_bader_pres_150206.pdf))
3. A fast and robust algorithm for Bader decomposition of charge density. _Computational Materials Science_ ([Journal](http://theory.cm.utexas.edu/henkelman/code/bader/download/henkelman06_354.pdf))

**Checklist**

- [x] #893 Add pyquante2 to Travis configuration
- [x] #894 Update pyquante bridge for pyquante2 support
- [x] #896 Update volume method for updated cclib2pyquante bridge
- [x] #897 Add pyquante2 routines in volume method
- [x] #898 \<Bug Fix for existing code\> Fix formula that determine number of grid points in volume function
- [x] #899 Update test routine for volume method to include verify data from each point on a grid
- [x] #900 Add QTAIM method
- [x] #907  Add function to read cube files and generate Volume object; Add the ability to use precalculated charge densities in Bader method
- [x] #902 Add a few more tests for Bader's QTAIM method
- [x] #903 Add docs for QTAIM method

Issue #757, opened 2019-08-21
PyQuante does not support Python 3 (yet)


schneiderfelipe:
Hi all,

I got myself interested in obtaining densities (I believe some QTAIM/ELF/NCI stuff can be easily implemented on top of cclib) and the code in `cclib.method.volume` seems perfect for that.
Unfortunately, it's heavily based on PyQuante (the lines [208](https://github.com/cclib/cclib/blob/master/cclib/method/volume.py#L208) and [227](https://github.com/cclib/cclib/blob/master/cclib/method/volume.py#L227) are critical), a Python 2 only library.

Is there a way around this?  It looks like PyQuante is only used for the spatial representation of wavefunctions. Could it be replaced (or a fallback be used in case PyQuante isn't available)?

shivupa:
Hey sorry I saw your PR as well, but you could also try to add pyscf integration. There is the`eval_gto` functionality that could help.

schneiderfelipe:
> Hey sorry I saw your PR as well, but you could also try to add pyscf integration. There is the`eval_gto` functionality that could help.

@shivupa Do you mean [eval_ao](https://sunqm.github.io/pyscf/dft.html?highlight=eval#pyscf.dft.numint.eval_ao)? By the way, thanks for mentioning pyscf, I didn't know about it.

shivupa:
No worries and yes that's the one I was thinking of!

berquist:
An option is to add [pyquante2](https://github.com/rpmuller/pyquante2/blob/6e34cb4480ae7dbd8c5e44d221d8b27584890c83/pyquante2/basis/cgbf.py#L57)/Psi4/pyscf/etc bridges, implement `volume.getbfs` for each one, and then pass this as a function pointer to `volume.wavefunction` and `volume.electrondensity` to abstract away this part of the implementation.

berquist:
We have dropped Python 2 support, moved away from PyQuante to pyquante2, and Psi4 and pyscf bridges exist. https://github.com/cclib/cclib/issues/757#issuecomment-529069235 is still missing.

amandadumi:
I just wanted to chime in and say I have a some (messy) work started that does exactly this for pyscf! I was working on mimicking the current data structures, but if you plan on changing this, I can certainly adjust.

PR #893, opened 2020-06-29
Add pyquante2 to Travis configuration

mscho527:
Related Issue: #886 

This is a first PR associated with issue #886. It adds pyquante2 into the list of packages to install for Travis builds. Subsequent PRs will build upon this PR.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,6 +12,7 @@ biopython==1.76
 openbabel==2.4.1
 # For some reason PyQuante can't be installed from PyPI.
 https://sourceforge.net/projects/pyquante/files/PyQuante-1.6/PyQuante-1.6.5/PyQuante-1.6.5.tar.gz  ; python_version <= '2.7'
+git+https://github.com/rpmuller/pyquante2.git ; python_version > '3.0'
 
 # For building the documentation
 sphinx


mscho527:
Just noticed #757 and `cclib.method.volume` -- looks like it would be better starting from there 😲

.travis.yml
```diff
@@ -35,6 +36,7 @@ before_install:
         conda install pytest pytest-cov codecov -c conda-forge
         pip install pytest-shutil
         pip install git+https://github.com/theochem/iodata.git
+        pip install git+https://github.com/rpmuller/pyquante2.git
```
berquist, line 39: I think what we should do instead of having this added in two separate ways (`pip` installing `master` and `conda` installing an older tagged version) is to have this `pip install git...` line after lines 13 and 14 in `requirements.txt`: https://github.com/cclib/cclib/blob/0217c94cb06e1c812459fc87ee65b877c7ba3599/requirements.txt#L14
so that we can just drop PyQuante (1) later if possible.

mscho527: I've made the changes that you suggested -- PyQuante 1.6.5 should be installed for Python 2.x.x machine in Travis (from sourceforge) and pyquante2 should be installed for Python 3.x.x machine (from github) in Travis.

PR #896, opened 2020-07-02
Update volume method for updated cclib2pyquante bridge

mscho527:
_Related Issue: #886
Builds upon: #894_

This PR updates volume method to match updated pyquante bridge.

Subsequent PR would build upon this PR and would add pyquante2 equivalent code for `volume.py`.

### DIFF
--- a/cclib/method/volume.py
+++ b/cclib/method/volume.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2017, the cclib development team
+# Copyright (c) 2020, the cclib development team
 #
 # This file is part of cclib (http://cclib.github.io) and is distributed under
 # the terms of the BSD 3-Clause License.
@@ -29,7 +29,6 @@ if _found_pyvtk:
 def _check_pyvtk(found_pyvtk):
     if not found_pyvtk:
         raise ImportError("You must install `pyvtk` to use this function")
-
 
 class Volume(object):
     """Represent a volume in space.
@@ -136,9 +135,16 @@ def scinotation(num):
     return ("%sE%s%s" % (broken[0], sign, broken[1][-2:])).rjust(12)
 
 
-def getbfs(coords, gbasis):
-    """Convenience function for both wavefunction and density based on PyQuante Ints.py."""
-    mymol = cclib2pyquante.makepyquante(coords, [0 for _ in coords])
+def getbfs(ccdata):
+    """Convenience function for both wavefunction and density based on PyQuante Ints.py.
+    
+    Input:
+        ccdata -- ccData object
+        
+    Output:
+        bfs -- list of PyQuante CGBF objects
+    """
+    pymol = cclib2pyquante.makepyquante(ccdata)
 
     sym2powerlist = {
         'S' : [(0, 0, 0)],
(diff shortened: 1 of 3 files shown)


cclib/method/volume.py
```diff
@@ -224,11 +253,11 @@ def electrondensity(coords, mocoeffslist, gbasis, volume):
...
-                        data[i, j, :] = tmp
-                data *= mocoeff[bs]
+                            tmp.append(bfs[bs].amp(xval, yval, zval)) # first get amplitudes
+                        data[i, j, :] = tmp # this contains the amplitudes
+                data *= mocoeff[bs] # then multiply with mocoeffs until homos
                 wavefn += data
-            density.data += wavefn ** 2
+            density.data += wavefn ** 2 # lastly put them into second power
 
     # TODO ROHF
```
mscho527, line 262: I'm not too sure why this comment was left in the original implementation -- `electrondensity` function calculates the charge density from a user-designated list of orbitals anyway (which could be entire occupied orbitals or just a few frontier orbitals). If the calculation is restricted, it would just be fine to multiply by two here.

shivupa: Wouldn't that only be the case if it was a restricted closed shell system? For ROHF, we would need to only multiply the `wavefn**2` by 2 for the doubly occupied orbitals and by 1 otherwise?

mscho527: Yes -- but I thought that what the original author meant was to call the function using something like this for ROHF:
`volume.electrondensity(data.atomcoords[0], [data.mocoeffs[0][:data.homos[0]], data.mocoeffs[1][:data.homos[1]]
`
to indicate different number of orbitals to consider for ROHF calculations.

This is how the function was used in the test suite:
```
frontierorbs = [data_sp.mocoeffs[0][(data_sp.homos[0] - 3):(data_sp.homos[0] + 1)]]
        density = volume.electrondensity(data_sp.atomcoords[0],
                                         frontierorbs, data_basis.gbasis, vol)
```

Checking that inside the function makes a lot sense though and adding it would be pretty simple.

mscho527:
Rebased against master to reflect changes made in PR#894

cclib/method/volume.py
```diff
@@ -149,29 +155,36 @@ def getbfs(coords, gbasis):
...
-        for sym, prims in bs:
-            for power in sym2powerlist[sym]:
-                bf = CGBF(atom.pos(), power)
-                for expnt, coef in prims:
-                    bf.add_primitive(expnt, coef)
-                bf.normalize()
-                bfs.append(bf)
+    
+    #PyQuante
+    if _found_pyquante:
```
berquist, line 160: It would be better to have this guard the full definition of `getbfs`, rather than call the function and have nothing happen. (I know that's not how it's done in other parts of the code.)

cclib/method/volume.py
```diff
@@ -181,21 +194,34 @@ def wavefunction(coords, mocoeffs, gbasis, volume):
...
-    for bs in range(len(bfs)):
-        data = numpy.zeros(wavefn.data.shape, "d")
-        for i, xval in enumerate(x):
-            for j, yval in enumerate(y):
-                for k, zval in enumerate(z):
-                    data[i, j, k] = bfs[bs].amp(xval, yval, zval)
-        data *= mocoeffs[bs]
-        wavefn.data += data
+    #PyQuante
+    if _found_pyquante:
```
berquist, line 198: Same comment here.

cclib/method/volume.py
```diff
@@ -224,11 +253,11 @@ def electrondensity(coords, mocoeffslist, gbasis, volume):
                     for j, yval in enumerate(y):
                         tmp = []
                         for zval in z:
-                            tmp.append(bfs[bs].amp(xval, yval, zval))
-                        data[i, j, :] = tmp
-                data *= mocoeff[bs]
+                            tmp.append(bfs[bs].amp(xval, yval, zval)) # first get amplitudes
+                        data[i, j, :] = tmp # this contains the amplitudes
+                data *= mocoeff[bs] # then multiply with mocoeffs until homos
```
berquist, line 258: I've noticed a bunch of these comments, but don't think they add much. Did you have a hard time understanding the code? I'd rather have a descriptive code comment toward the top of the function in that case.

cclib/method/volume.py
berquist: We concluded that we aren't going to worry about this right now, right?

mscho527: @berquist Yes, that's correct.

cclib/method/volume.py
mscho527: If `_found_pyquante` or `_found_pyquante2` (in the next PR) are not `True`, `cclib2pyquante.makepyquante`, which is the first line to be executed in both functions should trigger an Exception. Would the exception be enough? Adding something like `if not _found_pyquante: raise someException` could be more explicit?

cclib/method/volume.py
mscho527: No -- I've written those when I was reading it and forgot to remove them back. Thanks.

cclib/method/volume.py
berquist: No, the exception is fine for now.

PR #897, opened 2020-07-06
Add pyquante2 routines in volume method

mscho527:
_Related Issue: #886
Builds upon: #896_ 

This PR updates volume method to support pyquante2.

Subsequent PR would build upon this PR and would add more testing to the volume method (which built upon existing method and was updated in #896 and #897) so that the calculated electron densities are verified for individual points in Cartesian coordinates.

### DIFF
--- a/cclib/method/volume.py
+++ b/cclib/method/volume.py
@@ -12,23 +12,92 @@ import copy
 
 import numpy
 
+from cclib.bridge import cclib2pyquante
 from cclib.parser.utils import convertor
 from cclib.parser.utils import find_package
+
+sym2powerlist = {
+    'S' : [(0, 0, 0)],
+    'P' : [(1, 0, 0), (0, 1, 0), (0, 0, 1)],
+    'D' : [(2, 0, 0), (0, 2, 0), (0, 0, 2), (1, 1, 0), (0, 1, 1), (1, 0, 1)],
+    'F' : [(3, 0, 0), (2, 1, 0), (2, 0, 1), (1, 2, 0), (1, 1, 1), (1, 0, 2),
+           (0, 3, 0), (0, 2, 1), (0, 1, 2), (0, 0, 3)]
+    }
+
 
 _found_pyquante = find_package("PyQuante")
 if _found_pyquante:
     from PyQuante.CGBF import CGBF
-    from cclib.bridge import cclib2pyquante
+    
+    def getbfs(ccdata):
+        pymol = cclib2pyquante.makepyquante(ccdata)
+        
+        bfs = []
+        for i, atom in enumerate(pymol):
+            bs = ccdata.gbasis[i]
+            for sym, prims in bs:
+                for power in sym2powerlist[sym]:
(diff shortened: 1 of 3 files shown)


cclib/method/volume.py
```diff
@@ -168,10 +172,34 @@ def getbfs(ccdata):
...
+                    basisfunction = cgbf(atom.atuple()[1:4], powers = power, exps=exponentlist, coefs=coefficientlist)
+                    basisfunction.normalize()
+                    bfs.append(basisfunction)
+                    
     return bfs
 
+# Small wrapper PyQuante & pyquante2 function that evaluates basis function on a given point
+# Used in both `wavefunction` and `electrondensity`
+def pyamp(bfs, bs, x, y, z):
+    if _found_pyquante2:
```
berquist, line 197: I'm more convinced now that these checks should be outside the function definition for performance reasons and general consistency with what functions return (the desired value vs. `None`).

cclib/method/volume.py
```diff
@@ -253,17 +279,43 @@ def electrondensity(ccdata, volume, mocoeffslist):
...
             density.data += wavefn ** 2
 
-    # TODO ROHF
-    if len(mocoeffslist) == 1:
-        density.data *= 2.0
-
     return density
 
+def electrondensity(ccdata, volume, mocoeffslist):
+    """Calculate the magnitude of the electron density at every point in a volume for either up or down spin
```
berquist, line 291: Shouldn't this be
```suggestion
    """Calculate the magnitude of the total electron density at every point in a volume
```
?

cclib/method/volume.py
mscho527: I've added a check that raises ImportError when pyquante is not present outside the functions. Thanks.

cclib/method/volume.py
mscho527: Looks like I forgot to change the comment here. Thanks for pointing this out!

cclib/method/volume.py
```diff
@@ -253,17 +282,43 @@ def electrondensity(ccdata, volume, mocoeffslist):
...
+    Note: mocoeffs is a list of NumPy arrays. The list will be of length 1
+          for restricted calculations, and length 2 for unrestricted.
+    """
+
+    if len(mocoeffslist) == 2:
+        return electrondensity_spin(ccdata, volume, [mocoeffslist[0]]) + electrondensity_spin(ccdata, volume, [mocoeffslist[1]])
+    else:
+        edens = electrondensity_spin(ccdata, volume, [mocoeffslist[0]])
+        edens.data *= 2
+        return edens
```
shivupa, line 320: This looks good for now but in the future this should be modified to take a `coeffslist` that would either be 2D or 1D and an occupation number list that scales the spin density by (2D case) 1*occno[i] or (1D case) 2*occno[i]. This would require changes in `electrondensity_spin` but it would allow us to use NOs as well as MOs. No necessary for this PR, but I thought I would add a note.

PR #898, opened 2020-07-07
Fix formula that determine number of grid points in volume function

mscho527:
_Related Issue: #886
Builds upon: #897_ 

This small PR fixes the formula used in `wavefunction` and `electrondensity` functions to determine the number of grid points.
Current formula uses `topcorner+spacing` as the end point for numpy's `arange` function to make sure that `topcorner` is included as one of grid points when `topcorner-origin` is divisible by `spacing`. This causes off-by-one error for some grids where `topcorner-origin` is indivisible by `spacing`. Substituting `topcorner+spacing` into `topcorner+epsilon*spacing` solves this off-by-one error.

### DIFF
--- a/cclib/method/volume.py
+++ b/cclib/method/volume.py
@@ -221,9 +221,10 @@ def wavefunction(ccdata, volume, mocoeffs):
     wavefn.data = numpy.zeros(wavefn.data.shape, "d")
 
     conversion = convertor(1, "bohr", "Angstrom")
-    x = numpy.arange(wavefn.origin[0], wavefn.topcorner[0] + wavefn.spacing[0], wavefn.spacing[0]) / conversion
-    y = numpy.arange(wavefn.origin[1], wavefn.topcorner[1] + wavefn.spacing[1], wavefn.spacing[1]) / conversion
-    z = numpy.arange(wavefn.origin[2], wavefn.topcorner[2] + wavefn.spacing[2], wavefn.spacing[2]) / conversion
+    gridendpt = wavefn.topcorner + 0.5 * wavefn.spacing
+    x = numpy.arange(wavefn.origin[0], gridendpt[0], wavefn.spacing[0]) / conversion
+    y = numpy.arange(wavefn.origin[1], gridendpt[1], wavefn.spacing[1]) / conversion
+    z = numpy.arange(wavefn.origin[2], gridendpt[2], wavefn.spacing[2]) / conversion
 
     # PyQuante & pyquante2
     for bs in range(len(bfs)):
@@ -266,9 +267,10 @@ def electrondensity_spin(ccdata, volume, mocoeffslist):
     density.data = numpy.zeros(density.data.shape, "d")
 
     conversion = convertor(1, "bohr", "Angstrom")
-    x = numpy.arange(density.origin[0], density.topcorner[0] + density.spacing[0], density.spacing[0]) / conversion
-    y = numpy.arange(density.origin[1], density.topcorner[1] + density.spacing[1], density.spacing[1]) / conversion
-    z = numpy.arange(density.origin[2], density.topcorner[2] + density.spacing[2], density.spacing[2]) / conversion
+    gridendpt = density.topcorner + 0.5 * density.spacing
+    x = numpy.arange(density.origin[0], gridendpt[0], density.spacing[0]) / conversion
+    y = numpy.arange(density.origin[1], gridendpt[1], density.spacing[1]) / conversion
+    z = numpy.arange(density.origin[2], gridendpt[2], density.spacing[2]) / conversion
 
     # For occupied orbitals
     # `mocoeff` and `gbasis` in ccdata object is ordered in a way `homos` can specify which orbital


cclib/method/volume.py
```diff
@@ -136,9 +139,16 @@ def scinotation(num):
...
 
 
-def getbfs(coords, gbasis):
-    """Convenience function for both wavefunction and density based on PyQuante Ints.py."""
-    mymol = cclib2pyquante.makepyquante(coords, [0 for _ in coords])
+def getbfs(ccmol):
+    """Convenience function for both wavefunction and density based on PyQuante Ints.py.
+    
+    Input:
+        ccmol -- ccData object
```
langner, line 146: Perhaps call this `data` or `ccdata`.

cclib/method/volume.py
```diff
@@ -136,9 +139,16 @@ def scinotation(num):
...
+        
+    Output:
+        bfs -- list of PyQuante CGBF object
```
langner, line 149: objectS?

cclib/method/volume.py
```diff
@@ -149,53 +159,95 @@ def getbfs(coords, gbasis):
...
+                    basisfunction = cgbf(atom.atuple()[1:4], exps=exponentlist, coefs=coefficientlist)
+                    basisfunction.normalize()
+                    bfs.append(basisfunction)
+                    
     return bfs
 
+# Small wrapper PyQuante & pyquante2 function that evaluates basis function on a given point
+# Used in both `wavefunction` and `electrondensity`
+def pyamp(bfs, bs, x, y, z):
+    if _found_pyquante2:
```
langner, line 197: What if both of these are false?

cclib/method/volume.py
```diff
@@ -149,53 +159,95 @@ def getbfs(coords, gbasis):
...
-        coords -- the coordinates of the atoms
-        mocoeffs -- mocoeffs for one eigenvalue
-        gbasis -- gbasis from a parser object
-        volume -- a template Volume object (will not be altered)
+
+def wavefunction(ccmol, volume, mocoeffs):
+    """Calculate the magnitude of the wavefunction at every point in a volume.
+    
+    Inputs:
+        ccmol -- ccData object
```
langner, line 208: `ccdata`?

cclib/method/volume.py
mscho527: This applies to PR #896 as well, so I've modified both of those. Thanks!

cclib/method/volume.py
mscho527: If both are false, `pyamp` wouldn't be called by `wavefunction` or `electrondensity` because each of the function's first lines -- getbfs include checking the existence of the packages and would throw Exception if any of pyquante or pyquante2 is not present. I could add the exception here as well if that would make things more explicit?

cclib/method/volume.py
mscho527: I've updated this. Thanks!

cclib/method/volume.py
```diff
@@ -221,9 +221,9 @@ def wavefunction(ccdata, volume, mocoeffs):
     wavefn.data = numpy.zeros(wavefn.data.shape, "d")
 
     conversion = convertor(1, "bohr", "Angstrom")
-    x = numpy.arange(wavefn.origin[0], wavefn.topcorner[0] + wavefn.spacing[0], wavefn.spacing[0]) / conversion
-    y = numpy.arange(wavefn.origin[1], wavefn.topcorner[1] + wavefn.spacing[1], wavefn.spacing[1]) / conversion
-    z = numpy.arange(wavefn.origin[2], wavefn.topcorner[2] + wavefn.spacing[2], wavefn.spacing[2]) / conversion
+    x = numpy.arange(wavefn.origin[0], wavefn.topcorner[0] + 0.5*wavefn.spacing[0], wavefn.spacing[0]) / conversion
```
berquist, line 224: Can you factor this back out into a constant?

cclib/method/volume.py
```diff
@@ -286,7 +286,7 @@ def electrondensity_spin(ccdata, volume, mocoeffslist):
                         data[i, j, :] = tmp
                 data *= mocoeff[bs]
                 wavefn += data
-            density.data += wavefn ** 2
+            density.data += wavefn ** 2 # lastly put them into second power
```
berquist, line 289: Is there a reason for this comment?

shivupa:
I agree with eric's requested changes. Once resolved this is LGTM

cclib/method/volume.py
mscho527: I've just factored this out as a constant. Thanks for the comment!

cclib/method/volume.py
mscho527: This looks like one of the comments that I've left in the earlier PR while reading through the code and forgot to remove. It looks like somehow this comment found its way back in :0 I've removed this. Thanks for letting me know.

berquist:
I am thinking that because of GitHub limitations with force-pushing, responding to review comments should be done in new commits, and we can just squash at the end...but then rebasing PRs that build on top of this gets more complicated...

PR #900, opened 2020-07-09
Add QTAIM method

mscho527:
_Related Issue: #886
Builds upon: #899_

This PR adds Bader's QTAIM charges to `cclib`, based on the on-grid algorithm proposed by Henkelman group in the paper doi:10.1016/j.commatsci.2005.04.010

### DIFF
--- a/cclib/method/bader.py
+++ b/cclib/method/bader.py
@@ -1,0 +1,214 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2020, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
+
+"""Calculation of Bader's QTAIM charges based on data parsed by cclib."""
+import copy
+import random
+import numpy
+import logging
+import math
+
+from cclib.method.calculationmethod import Method
+from cclib.method.volume import electrondensity_spin
+from cclib.parser.utils import convertor
+
+# Distance between two adjacent grids (sqrt[2] or sqrt[3] for uniform Cartesian grid).
+_griddist = numpy.array(
+    [
+        [[1.73205, 1.41421, 1.73205], [1.41421, 1, 1.41421], [1.73205, 1.41421, 1.73205],],
+        [[1.41421, 1, 1.41421], [1, 1, 1], [1.41421, 1, 1.41421]],
+        [[1.73205, 1.41421, 1.73205], [1.41421, 1, 1.41421], [1.73205, 1.41421, 1.73205],],
+    ],
+    dtype=float,
+)
+
+
+class MissingInputError(Exception):
+    pass
+
+
+def __cartesian_dist(pt1, pt2):
+    """ Small utility function that calculates Euclidian distance between two points
+        pt1 and pt2 are numpy arrays representing a point in Cartesian coordinates. """
+    return numpy.sqrt(numpy.einsum("ij,ij->j", pt1 - pt2, pt1 - pt2))
+
+
+class Bader(Method):
+    """Bader's QTAIM charges."""
+
+    # All of these are required for QTAIM charges.
+    required_attrs = ("homos", "mocoeffs", "nbasis", "gbasis")
+
+    def __init__(self, data, volume, progress=None, loglevel=logging.INFO, logname="Log"):
+        super(Bader, self).__init__(data, progress, loglevel, logname)
+
+        self.volume = volume
+        self.fragresults = None
+
+        if numpy.sum(self.data.coreelectrons) != 0:
+            # Pseudopotentials can cause Bader spaces to be inaccurate, as suggested by the
+            # original paper.
+            self.logger.info(
+                "It looks like pseudopotentials were used to generate this output. Please note that the Bader charges may not be accurate and may report unexpected results. Consult the original paper (doi:10.1016/j.commatsci.2005.04.010) for more information."
+            )
+
+    def __str__(self):
+        """Return a string representation of the object."""
+        return "Bader's QTAIM charges of {}".format(self.data)
+
+    def __repr__(self):
+        """Return a representation of the object."""
+        return "Bader({})".format(self.data)
+
+    def _check_required_attributes(self):
+        super(Bader, self)._check_required_attributes()
+
+    def calculate(self, indices=None, fupdate=0.05):
+        """Calculate Bader's QTAIM charges using on-grid algorithm proposed by Henkelman group
+           in doi:10.1016/j.commatsci.2005.04.010
+           
+           Cartesian, uniformly spaced grids are assumed for this function.
+           """
+
+        # First obtain charge densities on the grid
+        if len(self.data.mocoeffs) == 1:
+            self.chgdensity = electrondensity_spin(
+                self.data, self.volume, [self.data.mocoeffs[0][: self.data.homos[0]]]
+            )
+            self.chgdensity.data *= 2
+        else:
+            self.chgdensity = electrondensity_spin(
+                self.data,
+                self.volume,
+                [
+                    self.data.mocoeffs[0][: self.data.homos[0]],
+                    self.data.mocoeffs[1][: self.data.homos[1]],
+                ],
+            )
+
+        # Assign each grid point to Bader areas
+        self.fragresults = numpy.zeros(self.chgdensity.data.shape, "d")
+        next_index = 1
+
+        self.logger.info("Partitioning space into Bader areas.")
+
+        # Generator to iterate over the elements excluding the outermost positions
+        xshape, yshape, zshape = self.chgdensity.data.shape
+        indices = (
+            (x, y, z)
+            for x in range(1, xshape - 1)
+            for y in range(1, yshape - 1)
+            for z in range(1, zshape - 1)
+        )
+
+        for xindex, yindex, zindex in indices:
+            if self.fragresults[xindex, yindex, zindex] != 0:
+                # index has already been assigned for this grid point
+                continue
+            else:
+                listcoord = []
+                local_max_reached = False
+
+                while not local_max_reached:
+                    # Here, `delta_rho` corresponds to equation 2,
+                    # and `grad_rho_dot_r` corresponds to equation 1 in the aforementioned
+                    # paper (doi:10.1016/j.commatsci.2005.04.010)
+                    delta_rho = (
+                        self.chgdensity.data[
+                            xindex - 1 : xindex + 2,
+                            yindex - 1 : yindex + 2,
+                            zindex - 1 : zindex + 2,
+                        ]
+                        - self.chgdensity.data[xindex, yindex, zindex]
+                    )
+                    grad_rho_dot_r = delta_rho / _griddist
+                    maxat = numpy.where(grad_rho_dot_r == numpy.amax(grad_rho_dot_r))
+
+                    directions = list(zip(maxat[0], maxat[1], maxat[2]))
+                    next_direction = [ind - 1 for ind in directions[0]]
+
+                    if len(directions) > 1:
+                        # when one or more directions indicate max grad (of 0), prioritize
+                        # to include all points in the Bader space
+                        if directions[0] == [1, 1, 1]:
+                            next_direction = [ind - 1 for ind in direction[1]]
+
+                    listcoord.append((xindex, yindex, zindex))
+                    bader_candidate_index = self.fragresults[
+                        xindex + next_direction[0],
+                        yindex + next_direction[1],
+                        zindex + next_direction[2],
+                    ]
+
+                    if bader_candidate_index != 0:
+                        # Path arrived at a point that has already been assigned with an index
+                        bader_index = bader_candidate_index
+                        listcoord = tuple(numpy.array(listcoord).T)
+                        self.fragresults[listcoord] = bader_index
+
+                        local_max_reached = True
+
+                    elif (
+                        next_direction == [0, 0, 0]
+                        or xindex + next_direction[0] == 0
+                        or xindex + next_direction[0] == (len(self.chgdensity.data) - 1)
+                        or yindex + next_direction[1] == 0
+                        or yindex + next_direction[1] == (len(self.chgdensity.data[0]) - 1)
+                        or zindex + next_direction[2] == 0
+                        or zindex + next_direction[2] == (len(self.chgdensity.data[0][0]) - 1)
+                    ):
+                        # When next_direction is [0, 0, 0] -- local maximum
+                        # Other conditions indicate that the path is heading out to edge of
+                        # the grid. Here, assign new Bader space to avoid exiting the grid.
+                        bader_index = next_index
+                        next_index += 1
+
+                        listcoord = tuple(numpy.array(listcoord).T)
+                        self.fragresults[listcoord] = bader_index
+
+                        local_max_reached = True
+
+                    else:
+                        # Advance to the next point according to the direction of
+                        # maximum gradient
+                        xindex += next_direction[0]
+                        yindex += next_direction[1]
+                        zindex += next_direction[2]
+
+        # Now try to identify each Bader region to individual atom.
+        # Try to find an area that captures enough representation
+        self.matches = numpy.zeros_like(self.data.atomnos)
+        for pos in range(len(self.data.atomcoords[-1])):
+            gridpt = numpy.round(
+                (self.data.atomcoords[-1][pos] - self.volume.origin) / self.volume.spacing
+            )
+            xgrid = int(gridpt[0])
+            ygrid = int(gridpt[1])
+            zgrid = int(gridpt[2])
+            self.matches[pos] = self.fragresults[xgrid, ygrid, zgrid]
+
+        assert (
+            0 not in self.matches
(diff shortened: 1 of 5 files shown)


cclib/method/bader.py
```diff
@@ -0,0 +1,219 @@
...
+        super(Bader, self)._check_required_attributes()
+
+    def calculate(self, indices=None, fupdate=0.05):
+        """Calculate Bader's QTAIM charges using on-grid algorithm proposed by Henkelman group
+           in doi:10.1016/j.commatsci.2005.04.010
+           
+           Cartesian, uniformly spaced grids are assumed for this function.
+           """
+
+        def cartesian_dist(pt1, pt2):
```
berquist, line 60: I would make this a private function at the module level.

cclib/method/bader.py
```diff
@@ -0,0 +1,219 @@
...
+import numpy
+import logging
+import math
+
+from cclib.method.calculationmethod import Method
+from cclib.method.volume import electrondensity_spin
+from cclib.parser.utils import convertor
+
+# Distance between two adjacent grids (sqrt[2] or sqrt[3] for uniform Cartesian grid)
+griddist = [
```
berquist, line 19: Can you make this private?

cclib/method/bader.py
```diff
@@ -0,0 +1,219 @@
...
+
+                    if self.fragresults[xindex, yindex, zindex] != 0:
+                        # index has already been assigned for this grid point
+                        continue
+                    else:
+                        listcoord = []
+                        loopcondition = True
+
+                        while loopcondition:
+                            tmp = (
```
berquist, line 101: Can you mark this as being equation (2)? And maybe this is better called `delta_rho` or something similar, since this is a more significant quantity than implied by `tmp`.

cclib/method/bader.py
```diff
@@ -0,0 +1,219 @@
...
+                                self.chgdensity.data[
+                                    xindex - 1 : xindex + 2,
+                                    yindex - 1 : yindex + 2,
+                                    zindex - 1 : zindex + 2,
+                                ]
+                                - self.chgdensity.data[xindex, yindex, zindex]
+                            )
+                            grad = tmp / griddist
```
berquist, line 109: Can you mark this as being equation (1)? And call it something like `grad_rho_dot_r` or `grad_density_dot_r`?

cclib/method/bader.py
mscho527: I've made this function private. Thanks!

cclib/method/bader.py
mscho527: Yes -- I've just updated this variable to be private. Thanks.

cclib/method/bader.py
mscho527: Yes. I've just made the changes. Thanks!

cclib/method/bader.py
mscho527: I've just made the comments and renamed the variables. The code looks much more readable this way. Thanks!

cclib/method/bader.py
```diff
@@ -0,0 +1,223 @@
...
+                            elif (
+                                xindex + nextDirection[0] == 0
+                                or xindex + nextDirection[0] == (len(self.chgdensity.data) - 1)
+                                or yindex + nextDirection[1] == 0
+                                or yindex + nextDirection[1] == (len(self.chgdensity.data[0]) - 1)
+                                or zindex + nextDirection[2] == 0
+                                or zindex + nextDirection[2]
+                                == (len(self.chgdensity.data[0][0]) - 1)
+                            ):
+                                # Avoid exiting the grid
```
berquist, line 167: The body of this conditional is the same as the previous conditional, let's combine them.

cclib/method/bader.py
```diff
@@ -0,0 +1,223 @@
...
+                                # when one or more directions indicate max grad (of 0), prioritize
+                                # to include all points in the Bader space
+                                if list(zip(maxat[0], maxat[1], maxat[2]))[0] == (1, 1, 1,):
+                                    nextDirection = list(zip(maxat[0], maxat[1], maxat[2]))[1]
+                                    nextDirection = [ind - 1 for ind in nextDirection]
+
+                            listcoord.append((xindex, yindex, zindex))
+
+                            if (
+                                self.fragresults[
```
berquist, line 129: Can you store this outside the conditional as something like `bader_candidate_index` so you can reuse it inside the block?

cclib/method/bader.py
```diff
@@ -0,0 +1,223 @@
...
+                                # maximum gradient
+                                xindex += nextDirection[0]
+                                yindex += nextDirection[1]
+                                zindex += nextDirection[2]
+
+        # Now try to identify each Bader region to individual atom.
+        # Try to find an area that captures enough representation
+        self.matches = numpy.zeros_like(self.data.atomnos)
+        for pos in range(len(self.data.atomcoords[-1])):
+            gridpt = numpy.round(
```
berquist, line 187: I think you can replace `gridpt` with `xgrid, ygrid, zgrid` and also have `dtype=int` to avoid a separate conversion.

cclib/method/bader.py
```diff
@@ -0,0 +1,223 @@
...
+            * convertor(1, "Angstrom", "bohr") ** 3
+        )
+        self.fragcharges = numpy.zeros(len(self.data.atomcoords[-1]), "d")
+
+        for atomIndex, baderareaIndex in enumerate(self.matches):
+            # turn off all other grid points
+            vals = numpy.copy(self.chgdensity.data)
+            mask = self.fragresults == baderareaIndex
+            vals *= mask
+            self.fragcharges[atomIndex] = sum(vals.ravel()) * boxvol
```
berquist, line 221: Is there a way to rewrite this using `Volume.integrate()`? (I can't tell, kinda tired...)

cclib/method/bader.py
```diff
@@ -0,0 +1,223 @@
...
+                            # paper (doi:10.1016/j.commatsci.2005.04.010)
+                            delta_rho = (
+                                self.chgdensity.data[
+                                    xindex - 1 : xindex + 2,
+                                    yindex - 1 : yindex + 2,
+                                    zindex - 1 : zindex + 2,
+                                ]
+                                - self.chgdensity.data[xindex, yindex, zindex]
+                            )
+                            grad_rho_dot_r = delta_rho / _griddist
```
berquist, line 113: Looking at it again, I'm surprised this works since `_griddist` isn't an `ndarray`. Can you try making it one and make sure the results don't change?

cclib/method/bader.py
```diff
@@ -0,0 +1,223 @@
...
+        super(Bader, self)._check_required_attributes()
+
+    def calculate(self, indices=None, fupdate=0.05):
+        """Calculate Bader's QTAIM charges using on-grid algorithm proposed by Henkelman group
+           in doi:10.1016/j.commatsci.2005.04.010
+           
+           Cartesian, uniformly spaced grids are assumed for this function.
+           """
+
+        # First obtain charge densities on the grid
```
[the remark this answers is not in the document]
mscho527: Sure! I've added the logs already on a PR that depends on this (907) but it looks like the logger call should be present in this PR as well. Thanks.

cclib/method/bader.py
```diff
@@ -0,0 +1,223 @@
...
+                                    xindex - 1 : xindex + 2,
+                                    yindex - 1 : yindex + 2,
+                                    zindex - 1 : zindex + 2,
+                                ]
+                                - self.chgdensity.data[xindex, yindex, zindex]
+                            )
+                            grad_rho_dot_r = delta_rho / _griddist
+                            maxat = numpy.where(grad_rho_dot_r == numpy.amax(grad_rho_dot_r))
+
+                            nextDirection = list(zip(maxat[0], maxat[1], maxat[2]))[0]
```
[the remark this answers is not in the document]
mscho527: I've just stored it as `directions`. Thanks for the suggestion.

cclib/method/bader.py
mscho527: I've combined them. Thanks!

cclib/method/bader.py
mscho527: Yes -- I've just stored this and reused in the subsequent blocks. Thanks.

cclib/method/bader.py
mscho527: Yes. I've just replaced it to make it simpler and more readable. Thanks.

cclib/method/bader.py
mscho527: It is definitely possible and I've considered using `Volume.integrate()`. I think to use `Volume.integrate()`, though, n (# of atoms) number of Volume objects would have to be created each with appropriate masks to call on the method. So line 218-220 would still be there and line 221 would only be modified to a constructor for `Volume` object and call to `integrate()` method.

I think using `integrate()` method here would make it easier to follow any future changes in the method/object at an expense of a few more lines of code. I can make the necessary modifications if this fits better. Thanks!

cclib/method/bader.py
mscho527: I think this worked because division operator for Python on lists are elementwise divide as well? I've changed this into ndarray for consistency. As a side note, I forced pushed the commit by mistake -- this became some sort of a reflex :0

cclib/method/bader.py
mscho527: I've changed this part of the code back -- because indices to the array should not be numpy.ndarray, I think replacing gridpt with xgrid, ygrid, zgrid will still require separate line of int() conversion. Let me know if there is a way that is simpler.

cclib/method/bader.py
berquist: Yes, try it and see what it looks like.

cclib/method/bader.py
berquist: For some reason I can't figure out how to do this cleanly, so it's fine as it currently is.

cclib/method/bader.py
mscho527: I've modified this part of the code to use `integrate()` method. Let me know if this is good to go!

mscho527:
Rebased against master & squashed commits

cclib/method/bader.py
```diff
@@ -0,0 +1,204 @@
...
+import random
+import numpy
+import logging
+import math
+
+from cclib.method.calculationmethod import Method
+from cclib.method.volume import electrondensity_spin
+from cclib.parser.utils import convertor
+
+# Distance between two adjacent grids (sqrt[2] or sqrt[3] for uniform Cartesian grid)
```
langner, line 19: nit: please end sentences with periods.

cclib/method/bader.py
```diff
@@ -0,0 +1,204 @@
...
+    dtype=float,
+)
+
+
+class MissingInputError(Exception):
+    pass
+
+
+def __cartesian_dist(pt1, pt2):
+    """ Small utility function that calculates Euclidian distance between two points
```
langner, line 35: Same here.

cclib/method/bader.py
```diff
@@ -0,0 +1,204 @@
...
+        self.fragresults = numpy.zeros(self.chgdensity.data.shape, "d")
+        next_index = 1
+
+        self.logger.info("Partitioning space into Bader areas.")
+
+        # Exclude the outermost positions
+        for xwalker in range(1, len(self.chgdensity.data) - 1):
+            for ywalker in range(1, len(self.chgdensity.data[0]) - 1):
+                for zwalker in range(1, len(self.chgdensity.data[0][0]) - 1):
+                    (xindex, yindex, zindex) = (xwalker, ywalker, zwalker)
```
langner, line 96: Why not just use {x,y,z}index right away?

cclib/method/bader.py
```diff
@@ -0,0 +1,204 @@
...
+
+                    if self.fragresults[xindex, yindex, zindex] != 0:
+                        # index has already been assigned for this grid point
+                        continue
+                    else:
+                        listcoord = []
+                        loopcondition = True
```
langner, line 103: Could you take advantage of this variable's name to convey information about the condition of the loop? (choose a better name)

cclib/method/bader.py
```diff
@@ -0,0 +1,204 @@
...
+            )
+
+        # Assign each grid point to Bader areas
+        self.fragresults = numpy.zeros(self.chgdensity.data.shape, "d")
+        next_index = 1
+
+        self.logger.info("Partitioning space into Bader areas.")
+
+        # Exclude the outermost positions
+        for xwalker in range(1, len(self.chgdensity.data) - 1):
```
langner, line 93: Like we discussed, consider decreasing the indentation. One way to achieve this is helper functions. Another is to use a generator for the triple loop. Likely other good ideas as well, and anything that decreases the indent will help readability.

cclib/method/bader.py
```diff
@@ -0,0 +1,162 @@
...
+        """Calculate Bader's QTAIM charges using on-grid algorithm proposed by Henkelman group
+           in doi:10.1016/j.commatsci.2005.04.010
+           
+           Cartesian, uniformly spaced grids are assumed for this function.
+           """
+        
+        def cartesian_dist(pt1, pt2):
+            """ Small utility function that calculates Euclidian distance between two points
+                pt1 and pt2 are numpy arrays representing a point in Cartesian coordinates """
+            return numpy.sqrt(numpy.einsum("ij,ij->j", pt1-pt2, pt1-pt2s))
```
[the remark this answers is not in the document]
mscho527: I've fixed this. Thanks!

cclib/method/volume.py
```diff
@@ -326,7 +328,8 @@ def electrondensity_spin(ccdata, volume, mocoeffslist):
                         data[i, j, :] = tmp
                 data *= mocoeff[bs]
                 wavefn += data
-            density.data += wavefn ** 2
+            density.data += wavefn ** 2 # lastly put them into second power
+
```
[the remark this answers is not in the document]
mscho527: Yes -- I've removed this. Thanks!

test/method/testbader.py
```diff
@@ -0,0 +1,61 @@
...
+
+import numpy
+
+from cclib.method import Bader, Volume
+from cclib.parser import Psi4
+from cclib.method.calculationmethod import MissingAttributeError
+
+from numpy.testing import assert_allclose
+
+sys.path.insert(1, "..")
```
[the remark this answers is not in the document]
mscho527: I'm actually not sure about this. I've noticed that this line is included in every test script in `method` so I assumed that this is needed for Travis. Can someone let me know what the intended function of these were?

cclib/method/bader.py
```diff
@@ -0,0 +1,204 @@
...
+
+class Bader(Method):
+    """Bader's QTAIM charges."""
+
+    # All of these are required for QTAIM charges.
+    required_attrs = ("homos", "mocoeffs", "nbasis", "gbasis")
+
+    def __init__(self, data, volume, progress=None, loglevel=logging.INFO, logname="Log"):
+        super(Bader, self).__init__(data, progress, loglevel, logname)
+
```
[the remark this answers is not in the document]
mscho527: I've added a log that checks the contents of `coreelectrons` and lets the user know about this information through the log. Thanks!

cclib/method/bader.py
mscho527: I've changed this. Thanks!

cclib/method/bader.py
mscho527: I've changed this. Thanks!

cclib/method/bader.py
mscho527: I've cleaned this up while reformatting the triple loop. Thanks!

cclib/method/bader.py
mscho527: I've changed this variable to `local_max_reached`, since the loop exits when the local maximum is reached. Let me know if this can be improved further.

cclib/method/bader.py
mscho527: I've used a generator for the triple loop to reduce the indent by two. Thanks!

test/method/testbader.py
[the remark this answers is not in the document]
berquist: See 3026637b51b205ae9fcc32188d226c9d3da3a1c6.

The reason we manipulate `sys.path` at all (I think) is because the main `test` directory isn't in `PYTHONPATH`, and we need the relative `test_data` import to work.

We might be able to rip all of these out in the future, since now we're using `pytest` and not `unittest`, and I'd like to move the tests under the main `cclib` directory.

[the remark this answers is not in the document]
langner: Sounds about right. Sometimes it's hard to tell which legacy maintenance/test code can be removed. It's also somewhat unrewarding to work on it :)

cclib/method/bader.py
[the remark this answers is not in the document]
langner: One follow question here that shouldn't stop the merge: why didn't this break any test?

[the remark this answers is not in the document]
mscho527: @langner This didn't break any test because this function isn't (yet) used in the method. It is meant for use when support for non-uniform grids are added. For now, hard coded values (`_griddist`) is used instead.

test/method/testbader.py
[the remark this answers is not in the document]
mscho527: I've removed this line. I'll add an issue for other test functions. Thanks!

PR #902, opened 2020-07-15
Add a few more tests for Bader's QTAIM method

mscho527:
_Related Issue: #886 
Depends on: #900_ 

This PR adds a few more tests for Bader's QTAIM method. The aim is to add more checks into the code integrity without causing the test suite to take too long to complete by using the pre-calculated cube file.

Added tests compare
* Sum of charges assigned to HF vs. Total number of electrons in HF
* Charges assigned to carbon atoms in benzene


test/method/testbader.py
```diff
@@ -56,3 +57,31 @@ def test_val(self):
         refData = [6.9378, 0.3639, 0.3827]  # values from `bader` package
 
         assert_allclose(self.analysis.fragcharges, refData, atol=0.15)
+
+    def test_chgsum_hf(self):
+        """Does the sum of charges equate to the number of electrons for a simple molecule?"""
+
+        hfpath = os.path.dirname(os.path.realpath(__file__)) + "/hf.out"
```
berquist, line 64: Prefer `os.path.join` rather than hard-coding the separator `/` (If this happened in other PRs, I missed it.)

test/method/testbader.py
```diff
@@ -56,3 +57,31 @@ def test_val(self):
...
+        self.data = ccread(hfpath)
```
berquist, line 65: I don't think any of the new variables created in both of these methods need to be part of `self`, since they're only used in their respective methods.

test/method/testbader.py
mscho527: Thanks for letting me know of this. I've just change this to use `os.path.join`.

test/method/testbader.py
mscho527: Yes, I agree. I've changed the variables not to be part of self.

test/method/testbader.py
```diff
@@ -56,3 +57,31 @@ def test_val(self):
...
+
+    def test_symms_benzene(self):
+        """ Do the carbons in benzene ring get assigned with roughly equal charges?"""
+
+        benzenepath = os.path.join(os.path.dirname(os.path.realpath(__file__)), "benzene.out")
+        data = ccread(benzenepath)
+        vol = volume.read_from_cube(
+            os.path.join(os.path.dirname(os.path.realpath(__file__)), "benzene.cube")
+        )
+        assert_allclose(vol.origin, numpy.array([-5, -5, -5]), atol=0.5)
```
berquist, line 80: The value for each component of `vol.origin` is -5.05001036. Can you make both the reference value and the `atol` stricter?

test/method/testbader.py
```diff
@@ -56,3 +57,31 @@ def test_val(self):
...
+
+        analysis = Bader(data, vol)
+        analysis.calculate()
+
+        self.assertAlmostEqual(
+            analysis.fragcharges[0:6].max(), analysis.fragcharges[0:6].min(), delta=0.5
```
berquist, line 86: Looking at the resulting charges
```
array([4.36361172, 4.83148531, 4.66353415, 4.36361172, 4.8317697 ,
       4.66353415, 0.67225235, 0.51518601, 0.82588478, 0.67225235,
       0.51518601, 0.82588478])
```
do you know why there is such a large discrepancy?

test/method/testbader.py
mscho527: Sure -- I'll change this right away.

test/method/testbader.py
mscho527: My understanding is that the discrepancy is due to the shape of the grid (which is a cube). Now that you commented this, I guess I could check whether adjusting the grid improves the discrepancy?

berquist: I think just checking it by hand and leaving a code comment about the grid coarseness is enough (if that's the reason), since we don't want to introduce a finer grid in the tests themselves due to time constraints.

mscho527: I've just added a comment about this. Thanks!

berquist: Can you word it so that it's clear `bader` is referring to the original reference program?

mscho527: Sure -- I've just made the necessary changes. Thanks.

PR #908, opened 2020-07-16
Ensure only converged excited state properties are reported in Gaussian parser

### LINKED ISSUE #889: Gaussian 09 Opt + TDA yields excited states data at every iteration
Parsing a Gaussian 09 log file from optimising an excited state yields excited states data (etenergies etc...) from every round of the optimisation, not just the final energies.

This was surprising because, unlike atomcoords for example, etenergies and friends are rank 1 arrays, so all the excited states from every round of optimisation are together in a single array, rather than being split as for many other parsed properties.

A simple work around is to split etenergies where the energy suddenly decreases, or by the number of optimisation rounds (len(scfenergies) for example), but I'm not sure how stable a solution this is (wouldn't work if etenergies was re-ordered by increasing energy, for example).

I'm not sure if this is a bug or not, but I couldn't find this behaviour described in the docs so thought I'd post it.

I have an example log file, please let me know if it would be useful.

shivupa:
This fixes #889 [The regression test needs to be merged first.](https://github.com/cclib/cclib-data/pull/134)

### DIFF
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -922,7 +922,8 @@ class Gaussian(logfileparser.Logfile):
             # final step of an optimization.
             if not allconverged:
                 for reset_attr in ["etenergies", "etoscs", "etsyms", "etsecs", "etdips", "etveldips", "etmagdips"]:
-                    setattr(self, reset_attr, [])
+                    if hasattr(self, reset_attr):
+                        setattr(self, reset_attr, [])
 
             self.geovalues.append(newlist)
 
@@ -1974,7 +1975,7 @@ class Gaussian(logfileparser.Logfile):
         # Order is XX, YX, YY, ZX, ZY, ZZ (lower triangle).
         if line[2:23] == "Exact polarizability:":
             if not self.hp_polarizabilities:
-                if not hasattr(self, 'polarizabilities'):
+                if not  hasattr(self, 'polarizabilities'):
                     self.polarizabilities = []
                 polarizability = numpy.zeros(shape=(3, 3))
                 indices = numpy.tril_indices(3)
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -1975,7 +1975,7 @@ class Gaussian(logfileparser.Logfile):
         # Order is XX, YX, YY, ZX, ZY, ZZ (lower triangle).
         if line[2:23] == "Exact polarizability:":
             if not self.hp_polarizabilities:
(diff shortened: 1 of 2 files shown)


cclib/parser/gaussianparser.py
```diff
@@ -899,19 +899,35 @@ def extract(self, inputfile, line):
...
                     self.logger.error("Problem parsing the value for geometry optimisation: %s is not a number." % parts[2])
                 else:
                     newlist[i] = value
                 self.geotargets[i] = utils.float(parts[3])
+            # reset some parameters that are printed each iteration if the 
+            # optimization has not yet converged. For example, etenergies 
+            # (Issue #889) and similar properties are only reported for the
+            # final step of an optimization.
+            if not allconverged:
+                if hasattr(self, "etenergies"):
```
berquist, line 922: I recognize that all of these are likely to be present if `etenergies` is set (same for `etdips`), but I think all of these names should be kept in a list and you use `setattr`, it'll be cleaner.

cclib/parser/gaussianparser.py
```diff
@@ -899,19 +899,28 @@ def extract(self, inputfile, line):
...
                 else:
                     newlist[i] = value
                 self.geotargets[i] = utils.float(parts[3])
+            # reset some parameters that are printed each iteration if the 
+            # optimization has not yet converged. For example, etenergies 
+            # (Issue #889) and similar properties are only reported for the
+            # final step of an optimization.
+            if not allconverged:
+                for reset_attr in ["etenergies", "etoscs", "etsyms", "etsecs", "etdips", "etveldips", "etmagdips"]:
+                    setattr(self, reset_attr, [])
```
berquist, line 923: I think you want to check if the attribute is present first, since an empty `etsyms` list isn't the same thing as the attribute being absent (for example).

shivupa: good point resolved!

cclib/parser/gaussianparser.py
```diff
@@ -1962,7 +1973,7 @@ def natural_orbital_single_spin_parsing(inputfile, updateprogress_title):
         # Order is XX, YX, YY, ZX, ZY, ZZ (lower triangle).
         if line[2:23] == "Exact polarizability:":
             if not self.hp_polarizabilities:
-                if not hasattr(self, 'polarizabilities'):
+                if not  hasattr(self, 'polarizabilities'):
```
berquist, line 1976: Stray space?

shivupa: Fixed

PR #913, opened 2020-07-28
PySCF bridge

amandadumi:
A small PR to add a PySCF bridge and a unit. If there is anything missing from this list, please let me know!
I added a slightly more converged answer to the Psi4 test to allow for consistency between the reference values for the psi4 and the PySCF test. Consistency between these two tests is also the reason for the Cartesian basis functions being used in Pyscf even though that's not the default.

Progress.
- [x] Write the basic bridge to PySCF
- [x] Write corresponding test
- [x] Incorporate PySCF into Travis testing

### DIFF
--- a/test/bridge/testpsi4.py
+++ b/test/bridge/testpsi4.py
@@ -22,10 +22,11 @@ class Psi4Test(unittest.TestCase):
         psi4.core.set_output_file("psi4_output.dat", False)
 
         atomnos = np.array([1, 8, 1], "i")
-        a = np.array([[-1, 1, 0], [0, 0, 0], [1, 1, 0]], "f")
-        psi4mol = cclib2psi4.makepsi4(a, atomnos)
+        atomcoords = np.array([[-1, 1, 0], [0, 0, 0], [1, 1, 0]], "f")
+        psi4mol = cclib2psi4.makepsi4(atomcoords, atomnos)
+        psi4.set_options({'scf_type': 'pk'})
         en = energy("scf/6-31G**", molecule=psi4mol)
-        ref = -75.82474605514503
+        ref = -75.824754602
         assert abs(en - ref) < 1.0e-6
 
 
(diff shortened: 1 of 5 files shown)


test/bridge/testpsi4.py
```diff
@@ -25,7 +25,7 @@ def test_makepsi4(self):
         a = np.array([[-1, 1, 0], [0, 0, 0], [1, 1, 0]], "f")
```
shivupa, line 25: Since we are already modifying the test can we change `a` to `atompos` or `atomcoords`

test/bridge/testpyscf.py
```diff
@@ -0,0 +1,38 @@
...
+
+class PyscfTest(unittest.TestCase):
+    """Tests for the cclib2pyscf bridge in cclib."""
+
+    def test_makepyscf(self):
+        import pyscf
+        from pyscf import scf
+
+        atomnos = np.array([1, 8, 1], "i")
+        a = np.array([[-1, 1, 0], [0, 0, 0], [1, 1, 0]], "f")
```
shivupa, line 23: same here

shivupa:
Thanks, looks good to merge once pyscf is added to the travis

test/bridge/testpyscf.py
```diff
@@ -0,0 +1,38 @@
...
+
+        mhf = pyscfmol.HF(conv_tol=1e-6)
+        en = mhf.kernel()
+        ref = -75.824754602
+        assert abs(en - ref) < 1.0e-6
+
+
+if __name__ == "__main__":
+    a = PyscfTest()
+    a.test_makepyscf()
```
berquist, line 37: I don't think we need these two lines, the `unittest.main()` call should do this automatically.

amandadumi: Yes, you're right! I had forgot to take these out after troubleshooting locally. Thanks!

berquist:
I looked at the Travis log again and this isn't getting tested because of how pytest discovery works. You'll need to put `from .bridge.testpyscf import *` under the 3.6 version guard in `test/test_bridge.py`.

amandadumi:
Yes sorry, it seems when I was updating my branches history I lost a few of my changes without noticing. They should be back now!

PR #914, opened 2020-07-28
Add a function to read in proatom densities from chargemol's atomic_densities

mscho527:
_Related Issue: #885_ 

This PR adds a small function that reads in the pro-atomic (radial) densities from files that are formatted in `chargemol` and `horton`. These charges are used as reference charge densities for the first and second step of DDEC algorithm. Some layout of DDEC class has been added here as well.


ghutchis:
I don't have an answer yet on the source.

IIRC we can generate proatom densities using Horton (https://theochem.github.io/horton/2.1.0b1/user_postproc_aim.html#horton-atomdb-py-build-a-pro-atom-database) which seems the way to go.

mscho527:
I've changed the function and the test to read from horton's proatom database. Horton uses number of non-uniform grid for the proatomic densities, so it adds a few lines into the code (and perhaps next few PRs as well) but this should resolve the license issue. Thanks for letting me know about the license issue!

requirements.txt
```diff
@@ -6,6 +6,7 @@ six
 # methods
 scipy
 periodictable
+h5py
```
berquist, line 9: I know that the `requirements.txt` / `setup.py` / Travis (optional) dependencies are messy right now, but can you remove h5py from here for now?

Also, is it automatically installed as a Horton dependency on Travis?

timer.dat
```diff
@@ -0,0 +1,12 @@
+
+Host: homesvr
```
berquist, line 2: Thanks Psi4. Can you remove this file and add it to `.gitignore`?

cclib/method/ddec.py
```diff
@@ -0,0 +1,163 @@
...
+                # PowerRTransform is default in horton-atomdb.py.
+                gridtype, gridmin, gridmax, gridn = proatomdb[keystring].attrs["rtransform"].split()
+                gridmin = float(gridmin)
+                gridmax = float(gridmax)
+                gridn = int(gridn)
+                # Convert byte to string in Python3
+                if sys.version[0] == '3':
+                    gridtype = gridtype.decode('UTF-8')
+
+                if gridtype == "LinearRTransform":
```
berquist, line 117: Does Horton have an enum or listing of grid types in their code that can be imported? We should also have an `else` that raises an exception when the grid type is something we don't know how to handle.

timer.dat
mscho527: Oops -- sure! Thanks for letting me know!

requirements.txt
mscho527: In Python 2.7 machine, `h5py` is installed as a horton dependency. In Python 3.x machines, though, it isn't since horton is not installed there. Should h5py be moved to conda configuration then?

cclib/method/ddec.py
mscho527: There might be a list -- I'll take a look. This part of the code is implemented in cpp so I suppose that a list might reside somewhere in the python binding. I'll add the else / exception soon! Thanks.

requirements.txt
berquist: For now, yes.

cclib/method/ddec.py
mscho527: I've ended up adding a list (of supported grid types) and an assert statement because I wasn't able to find a list anywhere other than the documentation

test/method/testddec.py
```diff
@@ -0,0 +1,85 @@
...
+            self.parse()
+            delattr(self.data, missing_attribute)
+            with self.assertRaises(MissingAttributeError):
+                trialBader = DDEC6(self.data, self.volume)
+
+    def test_proatom_read(self):
+        """Are proatom densities imported correctly?"""
+
+        self.data, logfile = getdatafile(Psi4, "basicPsi4-1.2.1", ["water_mp2.out"])
+        self.volume = volume.Volume((-4, -4, -4), (4, 4, 4), (0.2, 0.2, 0.2))
```
shivupa, line 51: `self.parse()` here? Unless you were planning to do this on an even coarser grid since its not used in this test?

mscho527: Yes -- thanks for catching this! I've made the necessary changes.

mscho527:
While looking at PR915, I spotted that reading in grid radii from horton is incorrect. I'll leave another once this is fixed.

cclib/method/ddec.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+class DDEC6(Method):
+    """DDEC6 charges."""
+
+    # All of these are required for DDEC6 charges.
+    required_attrs = ("homos", "mocoeffs", "nbasis", "gbasis")
+
+    def __init__(
+        self, data, volume, proatom_path=None, progress=None, loglevel=logging.INFO, logname="Log"
```
langner, line 34: Could you perhaps add a docstring here describing the initializer arguments?

cclib/method/ddec.py
```diff
@@ -0,0 +1,170 @@
...
+        self.fragresults = None
+
+        if numpy.sum(self.data.coreelectrons) != 0:
+            # TODO: Pseudopotentials should be added back
+            pass
+
+        assert os.path.isdir(
+            proatom_path
+        ), "Directory that contains chargemol proatom densities should be added as an input."
+        # Read in reference charges
```
langner, line 48: I think this comment applies to lines up to 56 in current numbering. Perhaps add a line before the comment, and eliminate the blank lines in this block, for readability.

cclib/method/ddec.py
```diff
@@ -0,0 +1,170 @@
...
+                # Convert byte to string in Python3
+                if sys.version[0] == '3':
+                    gridtype = gridtype.decode('UTF-8')
+
+                # First verify that it is one of recognized grids
+                assert gridtype in ["LinearRTransform", "ExpRTransform", "PowerRTransform"], "Grid type not recognized."
+
+                if gridtype == "LinearRTransform":
+                    # Linear transformation. r(t) = rmin + t*(rmax - rmin)/(npoint - 1)
+                    gridcoeff = (gridmax - gridmin) / (gridn - 1)
```
langner, line 122: Are these guaranteed to be floats? If not, can be problematic for Python2.

cclib/method/ddec.py
```diff
@@ -0,0 +1,170 @@
...
+            # Use chargemol proatom densities
+            # Each shell is 5.600570811644798 angstroms apart (uniform).
+            # *scalefactor* = 10.58354497764173 in module_global_parameter.f08
+            density = numpy.loadtxt(path, skiprows=12, dtype=float)
+            radiusgrid = numpy.arange(1, len(density + 1)) * 5.600570811644798
+
+        elif os.path.isfile(horton_path):
+            # Use horton proatom densities
+            _found_h5py = find_package("h5py")
+            assert _found_h5py, ("h5py is needed to read in proatom densities from horton.")
```
langner, line 95: Do you need the temp var?

cclib/method/ddec.py
```diff
@@ -0,0 +1,170 @@
...
+
+            import h5py
+
+            with h5py.File(horton_path, "r") as proatomdb:
```
langner, line 99: It looks like there's one file in the Horton case. Do we need do read it multiple times?

mscho527: horton indeed has only one file and it would be more efficient to read in once in `__init__` and store it somewhere for future reference. Can I revisit this point after finishing up rest of the DDEC algorithms though? I am just not sure yet in which form proatomic densities will be needed in later steps. I'll add a `TODO` comment referring to this at the top of this code block. Thanks!

cclib/method/ddec.py
mscho527: No this wouldn't be necessary. I've removed this extraneous variable. Thanks.

cclib/method/ddec.py
mscho527: To my knowledge, these would be guaranteed to be floats due to lines 115-116

```
gridmin = convertor(float(gridmin), "bohr", "Angstrom")
gridmax = convertor(float(gridmax), "bohr", "Angstrom")
```

but if I missed anything, please let me know.

cclib/method/ddec.py
mscho527: I've added a comment & reflowed the code for better readability. Thanks!

cclib/method/ddec.py
mscho527: Sure -- I've added them. Thanks!

cclib/method/ddec.py
langner: SG, excllent plan.

Issue #885, opened 2020-06-19
Add method to calculate DDEC6 charges


mscho527:
_This issue is based on the GSoC proposal._

Implementing DDEC6 charges ([sourceforge.net/projects/ddec](url)), as introduced in 10.1039/c6ra04656h, will provide modern atomic charge partitioning method for `cclib`.

Seven steps outlined in the original article can be considered as milestones.

**References**
1. Original Paper (10.1039/c6ra04656h)
2. Related [Video](https://www.youtube.com/watch?v=ao9N9WVjLx0)

**Checklist**
- [x] #914 Add a function to read in proatom densities from chargemol's `atomic_densities`
- [x] #915 Set up the grid and calculate localized charge and stockholder-type charge for DDEC6 [Step 1 of algorithm]
- [x] #916 Calculate reference ion charge value [Step 2 of algorithm]
- [x] #919 Add a function that conditions reference density [Step 3 of algorithm]
- [x] #920 Calculate weights assigned to each grid point in DDEC6 [Step 4-7 of algorithm -- substeps 1~4]
- [x] #921 Add function that evaluates G_A(r_A) and H_A(r_A) for DDEC6 [Step 4-7 of algorithm -- substeps 5~7]
- [x] #922 Add more tests for DDEC6
- [x] #926 Add docs for DDEC6

**Related**
- [x] #917 Add optional weight argument for integrate method in volume **915 depends on this PR**
- [x] #927 Separate out Hirshfeld method from DDEC6 as a standalone method
- [x] #928 Add docs for Hirshfeld
- [x] #930 Clean up DDEC6

shivupa:
Herculean effort! Thanks @mscho527. Closed by #914 #915 #916 #917 #919 #920 #921 #922 #926 #927 #930 #928

PR #915, opened 2020-07-30
Set up the grid and calculate localized charge and stockholder-type charge for DDEC6

mscho527:
_Related Issue: #885 
Builds upon: #914 and #917_

This PR finishes setting up the integration grid for DDEC6 charges based on what has been read through the function added in PR 914. Also, it calculates stockholder charges and localized charges to evaluate reference charges. This amounts to the 'first' step in the seven-step-long DDEC6 algorithm as described in the original paper.

Test suite compares the results with what was obtained in `chargemol`.

### DIFF
--- a/cclib/method/ddec.py
+++ b/cclib/method/ddec.py
@@ -51,7 +51,7 @@ class DDEC6(Method):
         assert os.path.isdir(
             proatom_path
         ), "Directory that contains proatom densities should be added as an input."
-        
+
         # Read in reference charges.
         self.proatom_density = []
         self.radial_grid_r = []
@@ -70,6 +70,11 @@ class DDEC6(Method):
 
     def _check_required_attributes(self):
         super(DDEC6, self)._check_required_attributes()
+
+    def _cartesian_dist(self, pt1, pt2):
+        """ Small utility function that calculates Euclidian distance between two points
+            pt1 and pt2 are numpy arrays representing a point in Cartesian coordinates. """
+        return numpy.sqrt(numpy.dot(pt1 - pt2, pt1 - pt2))
 
     def _read_proatom(self, directory, atom_num, charge):
         """Return a list containing proatom reference densities."""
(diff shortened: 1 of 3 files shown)


cclib/method/ddec.py
```diff
@@ -0,0 +1,220 @@
...
+            localized_weight = localized_w[atomi]/localized_bigW
+            chargedensity.data *= localized_weight
+            self._localizedcharges[atomi] = self.data.atomnos[atomi] - chargedensity.integrate()
+            
+            chargedensity = copy.deepcopy(self.chgdensity)
+            stockholder_weight = stockholder_w[atomi]/stockholder_bigW
+            chargedensity.data *= stockholder_weight
+            self._stockholdercharges[atomi] = self.data.atomnos[atomi] - chargedensity.integrate()
+            
+            self.refcharges[atomi] = self._stockholdercharges[atomi] / 3.0 + 2.0 * self._localizedcharges[atomi] / 3.0
```
shivupa, line 220: can you add some parentheses to group this a bit.

mscho527: I've added some parentheses and a comment here. Thanks!

cclib/method/ddec.py
```diff
@@ -0,0 +1,243 @@
...
+        self.logger.info("Creating first reference charges.")
+        self.refcharges = numpy.zeros((atoms))
+        self._localizedcharges = numpy.zeros((atoms))
+        self._stockholdercharges = numpy.zeros((atoms))
+
+        for atomi in range(atoms):
+            chargedensity = copy.deepcopy(self.chgdensity)
+            localized_weight = localized_w[atomi] / localized_bigW
+            chargedensity.data *= localized_weight
+            self._localizedcharges[atomi] = self.data.atomnos[atomi] - chargedensity.integrate()
```
shivupa, line 233: ` self._localizedcharges[atomi] = self.data.atomnos[atomi] - (self.chgdensity.integrate()*localized_weight)` should do what you want and remove the deepcopy. In a `Volume`object, the sum of the grid data is multiplied by the spacings in x,y,z. Pulling the weighting out of the sum is fine so we can do the above. This prevents the need for a new function.

You may have already checked this out, but in case you didn't!

mscho527: I could be mistaken, but since `localized_weight` is three-dimensional (on x, y, z) and `self.chgdensity.integrate()` returns a float, the dimensions would not match? Let me know if I misunderstood your comment. Thanks!

shivupa: Oh you are right. I wasn't thinking about the shape of localized_weight. I wanted to see if there was a way around adding the function to volume, but it is necessary.

mscho527: #917 should take care of this issue. I'll modify this PR soon based on 917.

cclib/method/volume.py
```diff
@@ -229,6 +229,9 @@ def convert(x):
         with open(filename, "w") as outputfile:
             outputfile.write("\n".join(ans))
 
+    def coordinates(self, indices):
```
berquist, line 232: Docstring?

mscho527: I've added a short docstring here. Thanks!

PR #916, opened 2020-08-03
Calculate step 2 reference ion charge value of DDEC6

mscho527:
_Related Issue: #885 
Depends upon: #915_ 

This PR adds a functionality to interpolate proatomic densities for non-integer charge values. Also, it evaluates step 2 reference charges of DDEC6 algorithm.

### DIFF
--- a/cclib/method/ddec.py
+++ b/cclib/method/ddec.py
@@ -19,6 +19,8 @@ from cclib.method.volume import electrondensity_spin
 from cclib.parser.utils import convertor
 from cclib.parser.utils import find_package
 
+from typing import List
+
 
 class MissingInputError(Exception):
     pass
@@ -42,6 +44,7 @@ class DDEC6(Method):
 
         self.volume = volume
         self.fragresults = None
+        self.proatom_path = proatom_path
 
         if numpy.sum(self.data.coreelectrons) != 0:
             # TODO: Pseudopotentials should be added back
@@ -76,7 +79,12 @@ class DDEC6(Method):
             pt1 and pt2 are numpy arrays representing a point in Cartesian coordinates. """
         return numpy.sqrt(numpy.dot(pt1 - pt2, pt1 - pt2))
 
-    def _read_proatom(self, directory, atom_num, charge):
+    def _read_proatom(self, 
+                      directory,        # type = str
+                      atom_num,         # type = int
+                      charge            # type = float
+                      ):
+        # type: (...) -> numpy.ndarray, numpy.ndarray
         """Return a list containing proatom reference densities."""
         # TODO: Treat calculations with psuedopotentials
         # TODO: Modify so that proatom densities are read only once for horton
@@ -88,17 +96,36 @@ class DDEC6(Method):
         #       atoms.h5
         # File format:
         #   Starting from line 13, each line contains the charge densities for each shell
-        chargemol_path = os.path.join(
+        # If `charge` is not an integer, proatom densities have to be linearly interpolated between
+        # the densities of the ion/atom with floor(charge) and ceiling(charge)
+        charge_floor = int(math.floor(charge))
+        charge_ceil = int(math.ceil(charge))
+        
+        chargemol_path_floor = os.path.join(
             directory,
-            "c2_{:03d}_{:03d}_{:03d}_500_100.txt".format(atom_num, atom_num, atom_num - charge),
+            "c2_{:03d}_{:03d}_{:03d}_500_100.txt".format(atom_num, atom_num, atom_num - charge_floor),
+        )
+        chargemol_path_ceil = os.path.join(
+            directory,
+            "c2_{:03d}_{:03d}_{:03d}_500_100.txt".format(atom_num, atom_num, atom_num - charge_ceil),
         )
         horton_path = os.path.join(directory, "atoms.h5")
 
-        if os.path.isfile(chargemol_path):
+
+        if os.path.isfile(chargemol_path_floor) or os.path.isfile(chargemol_path_ceil):
             # Use chargemol proatom densities
             # Each shell is .05 angstroms apart (uniform).
             # *scalefactor* = 10.58354497764173 bohrs in module_global_parameter.f08
-            density = numpy.loadtxt(chargemol_path, skiprows=12, dtype=float)
+            if atom_num <= charge_floor:
+                density_floor = numpy.array([0])
+            else:
+                density_floor = numpy.loadtxt(chargemol_path_floor, skiprows=12, dtype=float)
+            if atom_num >= charge_ceil:
+                density_ceil = numpy.array([0])
+            else:
+                density_ceil = numpy.loadtxt(chargemol_path_ceil, skiprows=12, dtype=float)
+            
+            density = (charge_ceil - charge) * density_floor + (charge - charge_floor) * density_ceil
             radiusgrid = numpy.arange(1, len(density + 1)) * 0.05
 
         elif os.path.isfile(horton_path):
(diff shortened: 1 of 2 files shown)


berquist:
Follow up from our conversation about type annotations: if you add any explicit ones right now, you'll need to use the comment style (https://mypy.readthedocs.io/en/stable/python2.html) and we can convert them to 3.5 style later once 2.7 is dropped.

Here is my WIP branch: https://github.com/berquist/cclib/tree/mypy

mscho527:
> Follow up from our conversation about type annotations: if you add any explicit ones right now, you'll need to use the comment style (https://mypy.readthedocs.io/en/stable/python2.html) and we can convert them to 3.5 style later once 2.7 is dropped.
> 
> Here is my WIP branch: https://github.com/berquist/cclib/tree/mypy

Thanks! I'll add in comment-style type annotations where it is applicable.

cclib/method/ddec.py
```diff
@@ -88,17 +96,34 @@ def _read_proatom(self, directory, atom_num, charge):
...
+        if os.path.isfile(chargemol_path_floor) or os.path.isfile(chargemol_path_ceil):
             # Use chargemol proatom densities
             # Each shell is .05 angstroms apart (uniform).
             # *scalefactor* = 10.58354497764173 bohrs in module_global_parameter.f08
-            density = numpy.loadtxt(chargemol_path, skiprows=12, dtype=float)
+            if atom_num <= charge_floor:
+                density_floor = numpy.array([0])
+            else:
+                density_floor = numpy.loadtxt(chargemol_path_floor, skiprows=12, dtype=float)
+            if atom_num <= charge_ceil:
```
berquist, line 121: Without my understanding of where this comes from, isn't it more likely to be something like
```python
if atom_num >= charge_ceil:
    density_ceil = numpy.array([charge_ceil])
```
?

cclib/method/ddec.py
```diff
@@ -88,17 +96,34 @@ def _read_proatom(self, directory, atom_num, charge):
         #       atoms.h5
         # File format:
         #   Starting from line 13, each line contains the charge densities for each shell
-        chargemol_path = os.path.join(
+        charge_floor = int(math.floor(charge))
```
berquist, line 99: Can you explain where this comes from?

cclib/method/ddec.py
```diff
@@ -221,23 +280,19 @@ def calculate_refcharges(self):
...
-            chargedensity = copy.deepcopy(self.chgdensity)
-            localized_weight = localized_w[atomi] / localized_bigW
-            chargedensity.data *= localized_weight
-            self._localizedcharges[atomi] = self.data.atomnos[atomi] - chargedensity.integrate()
-
-            chargedensity = copy.deepcopy(self.chgdensity)
-            stockholder_weight = stockholder_w[atomi] / stockholder_bigW
-            chargedensity.data *= stockholder_weight
-            self._stockholdercharges[atomi] = self.data.atomnos[atomi] - chargedensity.integrate()
+            localizedcharges[atomi] = self.data.atomnos[atomi] - self.chgdensity.integrate(
```
berquist, line 288: I would put a code comment above each line referencing the exact equation number from the paper.

cclib/method/ddec.py
mscho527: From step 1 of the algorithm, non-integer reference charges will be assigned to each atom. So in step 2 of the algorithm, the proatom densities have to be linearly interpolated between the densities of the ion/atom with floor(charge) and ceiling(charge). Let me know if I missed anything. Thanks.

cclib/method/ddec.py
mscho527: Yes. Thanks for catching this!

cclib/method/ddec.py
mscho527: I've added the equation numbers above the appropriate lines. Thanks.

cclib/method/ddec.py
berquist: Can you add this as a code comment right above?

test/method/testddec.py
```diff
@@ -83,15 +83,20 @@ def test_proatom_read(self):
...
+
         analysis = DDEC6(self.data, imported_vol, os.path.dirname(os.path.realpath(__file__)))
         analysis.calculate()
-        
+
         # values from `chargemol` calculation
-        assert_allclose(analysis.refcharges, [-0.513006, 0.256231, 0.256775], rtol=.1)
+        # which is based on proatomic densities calculated with different basis set.
+        # (rtol is adjusted to account for this inevitable discrepancy)
+        assert_allclose(analysis.refcharges[0], [-0.513006, 0.256231, 0.256775], rtol=0.10)
```
langner, line 101: Were you going to add Horton values as well? And comment on where you think the chargemol discrepancy comes from?

mscho527: I was thinking of adding `horton` values after full implementation when I add other simple molecules into the test suite. It looks like I've added the comments about the discrepancy only on the Step 3 PR. I've included them here as well. Thanks!

cclib/method/ddec.py
mscho527: Yes -- I've just added it as a comment. Thanks!

cclib/method/ddec.py
```diff
@@ -212,32 +273,33 @@ def calculate_refcharges(self):
...
-            chargedensity = copy.deepcopy(self.chgdensity)
-            localized_weight = localized_w[atomi] / localized_bigW
-            chargedensity.data *= localized_weight
-            self._localizedcharges[atomi] = self.data.atomnos[atomi] - chargedensity.integrate()
-
-            chargedensity = copy.deepcopy(self.chgdensity)
-            stockholder_weight = stockholder_w[atomi] / stockholder_bigW
-            chargedensity.data *= stockholder_weight
-            self._stockholdercharges[atomi] = self.data.atomnos[atomi] - chargedensity.integrate()
+            # Equation 52 and 51 in doi: 10.1039/c6ra04656h
```
berquist, line 293: It took me a second to realize that their `Stock/Loc` superscript means Stockholder or localized (`{Stock,Loc}`), not one divided by the other...

cclib/method/ddec.py
```diff
@@ -88,17 +96,36 @@ def _read_proatom(self, directory, atom_num, charge):
...
 
-        if os.path.isfile(chargemol_path):
+
+        if os.path.isfile(chargemol_path_floor) or os.path.isfile(chargemol_path_ceil):
             # Use chargemol proatom densities
             # Each shell is .05 angstroms apart (uniform).
             # *scalefactor* = 10.58354497764173 bohrs in module_global_parameter.f08
-            density = numpy.loadtxt(chargemol_path, skiprows=12, dtype=float)
+            if atom_num <= charge_floor:
+                density_floor = numpy.array([0])
```
berquist, line 120: Coverage says this block isn't being tested. I still don't think it's right, but don't know what it should be.

mscho527: This block wouldn't be tested in the test suite because `chargemol` densities cannot be included in the code as discussed in   #914. That could be why the block is being tested?

cclib/method/ddec.py
mscho527: Yes -- the notations in the paper puzzled me for a while too!

cclib/method/ddec.py
```diff
@@ -212,32 +273,33 @@ def calculate_refcharges(self):
...
+            )
+            stockholdercharges[atomi] = self.data.atomnos[atomi] - self.chgdensity.integrate(
+                weights=(stockholder_w[atomi] / stockholder_bigW)
+            )
 
             # In DDEC6, weights of 1/3 and 2/3 are assigned for stockholder and localized charges.
-            self.refcharges[atomi] = (self._stockholdercharges[atomi] / 3.0) + (
-                self._localizedcharges[atomi] * 2.0 / 3.0
-            )
+            # (Equation 58 in doi: 10.1039/c6ra04656h)
```
shivupa, line 302: Discussed offline but this also eq 50

mscho527: Just added this into the comment. Thanks!

cclib/method/ddec.py
berquist: Yes, that's why. Just pointing out that this is dead code (for now at least).

Issue #132, opened 2014-09-15
Support ONIOM calcs in some way


langner:
This is related to #99 and #131, but different enough to warrant a separate issue. Recently, I fixed regression Gaussian03/ortho_prod_freq.log, which is an ONIOM job, so that only the real (whole) system is parsed. Eventually, we probably also want to support parsing the models systems, as fragments.

langner:
@Clyde-fare in #161 suggest creating a "fragment energies" attribute to deal with energies of fragment.

PR #918, opened 2020-08-04
Extracting APT charges and any summed atom charges / spins from Gaussian

elliotfarrar:
I have recently been using cclib to extract various properties from Gaussian .out files. Whilst doing so, I noticed that only Mulliken and Lowdin atom charges and spins can be extracted, however I also wanted to extract APT charges, and charges and spins summed into heavy atoms for each property. I tried adding extra if statements for these properties here:

https://github.com/cclib/cclib/blob/8833b3e1003ac4d42841608bb05ebb32a698a1fc/cclib/parser/gaussianparser.py#L1873-L1876

However, due to the way in which the values are parsed, some properties were missed out (charges/spins are parsed by iterating over lines until a line not containing "Sum of" is reached, however the summed properties (e.g. "Mulliken atomic charges with hydrogens summed into heavy atoms") tend not to be followed by such a line, causing issues with the next property).

https://github.com/cclib/cclib/blob/8833b3e1003ac4d42841608bb05ebb32a698a1fc/cclib/parser/gaussianparser.py#L1892

To solve this, I reworked elements of the existing code into a new function to extract charges/spins (line 1882 in PR branch):

`def extract_charges_spins(property):`

This function calculates the exact number (n) of charges/atoms in advance (lines 1908-1912), and only these lines are iterated over:

```
is_sum = 'summed' in line
        if is_sum:
                n = self.natom - self.nhydrogen
        else:
                n = self.natom
```

This uses a new "nhydrogen" attribute (number of hydrogens in the structure), which is calculated in three seperate places at the same time as natom (as a result, extraction of charges via this function is dependant on the attributes "natom" and "nhydrogen" existing, although this should be the case for any structure). 

Lines 244-251:
```
natom = 0
nhydrogen = 0
	while line.split() and not "Variables" in line and not "Leave Link" in line:
                natom += 1	              
                if line.split()[0] == "H": 
                          nhydrogen += 1
                line = inputfile.next()	 
        self.set_attribute('nhydrogen', nhydrogen)
```
Also on lines 479-487 and 585-592.

The function also takes into account that older versions of Gaussian tend to include hydrogens (with value 0) for summed charges, whilst newer versions omit them entirely (lines 1916-1917).

Using this function, the summed and non-summed charges and spins for a particular property (e.g. Mulliken) can be calculated using extract_charges_spins("mulliken") (lines 1937-1940), e.g.

```
if hasattr(self, "natom") and hasattr(self, "nhydrogen"):
            extract_charges_spins("mulliken")
            extract_charges_spins("lowdin")
            extract_charges_spins("APT")
```

I thought I would make a pull request in case this code could be considered useful, and would be glad to take any feedback as to how this might be improved to better fit with cclib, or any potential issues that would prevent it from working with cclib. I have tested this code on a variety of my personal files, as well as some files from cclib-data. If required, I would be happy to create proper tests.


elliotfarrar:
Looks like a couple of tests failed with older versions of python. I will take a look at these.

elliotfarrar:
F strings have been replaced by the .format method. Type hints have been removed. Added two additional calculations for nhydrogen, as the original calculation was not triggered in some test files.

cclib/parser/gaussianparser.py
```diff
@@ -1853,46 +1871,73 @@ def natural_orbital_single_spin_parsing(inputfile, updateprogress_title):
...
+                    charges = []
+                    spins = []
+                    # calculate how many lines need iterating over based on whether property is summed into hydrogens or not
+                    is_sum = 'summed' in line
+                    if is_sum:
+                        n = self.natom - self.nhydrogen
+                    else:
+                        n = self.natom
+                    # iterate over each line and append values to a list based on whether they are charges or spins
+                    for i in range(n):
```
berquist, line 1914: Correct me if I'm wrong, but `nhydrogen` is only used for this one loop, right? Do you need it as part of the parsed results? If not, it doesn't need to go in `cclib.parser.data`: it will be available as an attribute during the extraction step, but get cleaned up by https://github.com/cclib/cclib/blob/502d967db9072e7d1ed61f3890fe34f03d00dffb/cclib/parser/logfileparser.py#L379.

cclib/parser/gaussianparser.py
```diff
@@ -1853,46 +1871,73 @@ def natural_orbital_single_spin_parsing(inputfile, updateprogress_title):
...
+                        nline = next(inputfile)
+                        while is_sum and nline.split()[1] == "H":  # some files include hydrogens in lists of summed charges with value 0.0, so these should not be recorded
```
berquist, line 1916: Can you put this comment on the line above the `while` and make sure it doesn't exceed ~100 characters (here and in general)?

cclib/parser/gaussianparser.py
```diff
@@ -1853,46 +1871,73 @@ def natural_orbital_single_spin_parsing(inputfile, updateprogress_title):
...
-
+        # Spins may be included in the same section, e.g.
+        #
+        # Mulliken charges and spin densities:
+        #               1          2
+        #     1  O   -0.596188   0.019133
+        #     2  C    0.320624   0.000869
+        #
+        # APT and Lowdin charges are also displayed in this way
+        def extract_charges_spins(property):
```
berquist, line 1882: You should use something other than `property` as a name, since it's a Python built-in (decorator name).

cclib/parser/gaussianparser.py
```diff
@@ -1853,46 +1871,73 @@ def natural_orbital_single_spin_parsing(inputfile, updateprogress_title):
...
+            """Extracts atomic charges and spin densities into self.atomcharges and self.atomspins dictionaries
+    
+            Inputs:
+                property - property type to be extracted as a string(e.g. Mulliken, Lowdin, APT)
+            """
+
+            headers = [" atomic charges:",
```
berquist, line 1889: Not specific to this line, but a general design comment and question for everyone. Right now, the code is written like
``` python
    def extract_charges_spins(prop):
        # do discovery and parsing here
    extract_charges_spins("mulliken")
    extract_charges_spins("lowdin")
    extract_charges_spins("APT")
```
where in other places we have code that looks more like
``` python
def extract_charges_spins(self, intputfile, line, prop):
    # do parsing here
    # do discovery here
    if discovered_header_for_prop in line:
        prop = ...
        extract_charges_spins(self, inputfile, line, prop)
```
I think I like your approach more, since the latter misleads you into thinking that the method is usable outside of `extract`, which it is, but not practically, since for expected behavior to happen, `line` really needs to be in the right position before the method is called. My main confusion comes from hiding the detection (`headers` and related) inside the function. It's not clear if rewriting this bit would require removing too much from the function. What do you think?

elliotfarrar:
Thanks for the comments. I'll check these out and get back to you with tests.

cclib/parser/gaussianparser.py
elliotfarrar: Yes, you're right. It would be tidier to remove that as a parsed attribute.

cclib/parser/gaussianparser.py
elliotfarrar: I've tidied up any comments I added across the file

cclib/parser/gaussianparser.py
elliotfarrar: Good point. I've changed this to prop.

cclib/parser/gaussianparser.py
elliotfarrar: Personally, I like the extract_charges_spins() within extract(), and next to the related code, just because I find it tidier and more readable when it's all together. However, if the extract_charges_spins function was being used elsewhere, or was intended to be used outside of extract (which, as you say, in theory it can be, but in reality it won't be), then that would be a different story. Depending on the consunsus, I'm happy to change things about.

I agree with regards to the headers and detection part (I probably just got carried away trying to pack everything inside one function). In hindsight, I think it would make more sense (and be fine) if the headers/line-finding aspect came outside the function. For example, something like this might be better:

```
def extract_charges_spins():
    # parsing

props = ["mulliken", "APT", "Lowdin"]
headers = [" atomic charges:", " charges:", etc]

if hasattr(self, "natom") and hasattr(self, "nhydrogen"):
    for prop in props:
          for header in headers:
                 if '{}{}'.format(prop,header).lower() in line.lower():
                          extract_charges_spins()
```

I can rewrite that if that sounds better.

elliotfarrar:
Regarding the tests that failed: 

https://github.com/cclib/cclib/blob/502d967db9072e7d1ed61f3890fe34f03d00dffb/test/data/testSP.py#L63-L68

Because atomcharges now contains lists of summed charges (which don't include the hydrogens) we get assertion errors when comparing `len(atomcharges[property_summed])` to `natom`. I was just going to include an `if is_sum` statement and instead compare any summed charges to `natom - nhydrogen`. However, this did make me wonder whether there would be merit in including the hydrogens (with charge value 0) in the lists of summed charges, instead of omitting them.

For example, for a water molecule, where the second atom is the oxygen, we'd have:

`{'mulliken': array([x,y,x]), 'mulliken_sum': array([0,y+2x,0])}`

Instead of:

`{'mulliken': array([x,y,x]), 'mulliken_sum': array([y+2x])}`

It might seem like you're just adding redundant information, but in some cases, particularly with larger molecules, this could allow the user to see more clearly which atoms from the non-summed list match up with those from the summed list, since both lists are of the same length. But maybe it's not as concise? What do you think?

cclib/parser/gaussianparser.py
berquist: Yes, I think that's the best solution.

berquist:
> However, this did make me wonder whether there would be merit in including the hydrogens (with charge value 0) in the lists of summed charges, instead of omitting them.

+1 for doing this.

elliotfarrar:
Summed charges are now generated with 0's for hydrogens. Additionally, I have completely removed the nhydrogen attribute, and extraction of atomcharges is now dependant on the "atomnos" attribute (instead of natom + nhydrogen).

These changes raised two new issues in the CI tests. 

The first was for files including translation vectors (e.g. neoprene.out). Since the translation vector is registered as an extra atom with atomic number -2, this resulted in parsing of an extra line. This was fixed by ignoring anytime `atomno in atomnos == -2`. However, I did notice that the -2 is still reported as part of the atomnos attribute, e.g. for neoprene, which has 10 atoms, atomnos is reported as: `[ 6  1  1  6  1  6  6  1  1 17 -2]`. Is this the expected behaviour?

The second is with two files: OPT_oniom.log and oniom_IR_intensity.log. For these files, the atom numbering seems to change at random half way through the file. For example, the first set of Mulliken charges are 10 carbons (matching up with the cartesian coordinates given at the beginning of the file, as usual):

https://github.com/cclib/cclib-data/blob/f7af2f1a873061e6fb3910744b6f774cd818b328/Gaussian/Gaussian09/oniom_IR_intensity.log#L1237-L1248

```
Mulliken atomic charges:
              1
     1  C   -0.266345
     2  C    0.100570
     3  C   -0.152235
     4  C   -0.160591
     5  C   -0.000715
     6  C   -0.164977
     7  C   -0.297889
     8  C    0.175288
     9  C   -0.327703
    10  C    0.168025
```

However, for the second set, a hydrogen appears at position 9 (having checked through GaussView, the given charge here still corresponds to the carbon atom at position 9, not a hydrogen):

https://github.com/cclib/cclib-data/blob/f7af2f1a873061e6fb3910744b6f774cd818b328/Gaussian/Gaussian09/oniom_IR_intensity.log#L1928-L1939

```
 Mulliken atomic charges:
              1
     1  C   -0.253547
     2  C    0.196272
     3  C   -0.196506
     4  C   -0.194798
     5  C    0.195301
     6  C   -0.251898
     7  C    0.000000
     8  C    0.000000
     9  H    0.123413
    10  C   -0.181578
```

This also happens more in subsequent lines. Since the charges given are actually for the original atom order of atom numbering, this does not affect the extraction of non-summed charges (hence this causes no issues in the master branch), however for summed charges (where it is important to know whether a particular line is for a hydrogen or not) this causes the extraction to fail.

I'm not sure if I've missed something, but I haven't seen this in any other files. Is this a known issue? Is there a chance this file might have been editted manually?

langner:
Thanks for looking so deeply into the issues!

> The first was for files including translation vectors (e.g. neoprene.out). Since the translation vector is registered as an extra atom with atomic number -2, this resulted in parsing of an extra line. This was fixed by ignoring anytime `atomno in atomnos == -2`. However, I did notice that the -2 is still reported as part of the atomnos attribute, e.g. for neoprene, which has 10 atoms, atomnos is reported as: `[ 6 1 1 6 1 6 6 1 1 17 -2]`. Is this the expected behaviour?

This seems like a bug - created #950 for this. Feel free to filter out the -2 when working with charges, perhaps adding a comment so it's clear what's going on. Focus on getting the PR in perhaps, and if you like you can address this bug later, or we'll deal with it as time allows.

> The second is with two files: OPT_oniom.log and oniom_IR_intensity.log. For these files, the atom numbering seems to change at random half way through the file. For example, the first set of Mulliken charges are 10 carbons (matching up with the cartesian coordinates given at the beginning of the file, as usual):

ONIOM calcualtion divide the system into fragments and print result for each fragment as well as the supersystem. So this is kinda expected. Our current approach is to parse and provide only the subsystem. We would like to support the fragments as well (#132) but there has been no work in this area yet. So, again, for this specific PR, please make sure to parse the charges for the supersystem only. Let us know if you need help with that.

elliotfarrar:
Thanks for your reply, and sorry for taking so long to get back to this!
 
On the first point, the -2 is filtered out when working with charges, and I have added a comment to clarify this is what is happening (L1888-1890).
 
On the second, I was not familiar with oniom calculations or their formats, so that was abit of an oversight from me. Looking back at these examples again, I think I actually spotted a small issue with the code that was causing some oniom files to be parsed incorrectly, and in combination with the parsing of summed charges added here, causes errors (hence, the two remaining errors in the CI tests currently are expected).

I have created a new branch and PR for this (#966 ). As this issue causes errors with the parsing of summed charges, I believe completing this PR will be dependant on that one.

shivupa:
Hi Sorry about the delay #966 is in!

elliotfarrar:
The branch is merged with master and one conflict solved. Seems like all tests pass now with #966 merged into master, so I think this should be ready to go.

cclib/parser/gaussianparser.py
```diff
@@ -1909,47 +1910,109 @@ def natural_orbital_single_spin_parsing(inputfile, updateprogress_title):
...
+                            spins.append(float(nline.split()[2]))
+            else:
+                for i in self.atomnos:
+                    # Ignore translation vectors.
+                    if i == -2:
+                        pass
+                    else:
+                        nline = next(inputfile)
+                        if has_charges:
+                            charges.append(float(nline.split()[2]))
```
berquist, line 1980: Can you move the split out of the conditional, so there aren't three separate split calls?

cclib/parser/gaussianparser.py
```diff
@@ -1909,47 +1910,109 @@ def natural_orbital_single_spin_parsing(inputfile, updateprogress_title):
...
+                            spins.append(float(0))
+                    else:
+                        nline = next(inputfile)
+                        # Some older versions of Gaussian already include 
+                        # hydrogens with value 0 for summed charges or 
+                        # spins, so these should be ignored.
+                        while nline.split()[1] == "H":
+                            nline = next(inputfile)
+                        if has_charges:
+                            charges.append(float(nline.split()[2]))
```
berquist, line 1967: Same comment about split for this block.

cclib/parser/gaussianparser.py
```diff
@@ -1909,47 +1910,109 @@ def natural_orbital_single_spin_parsing(inputfile, updateprogress_title):
...
+                        if has_spin and has_charges:
```
berquist, line 1968: Is there a way to write each conditional block using `elif`?

berquist (approved):
Just a few questions.

cclib/parser/gaussianparser.py
elliotfarrar: Yeah, I've done this for both.

elliotfarrar:
I've also fixed an issue where charges were missed for some G98 and G03 files because the type of charge was not given explicity above the charges.

berquist (approved):
Sorry for dragging this out for so long!

PR #920, opened 2020-08-10
Calculate weights assigned to each grid point in DDEC6

mscho527:
_Related Issue: #885
Depends upon: #919_

This PR implements parts 1-4 of the fourth through seventh steps of DDEC6 algorithm. It uses the conditioned proatom densities and weighed average densities to perform stockholder type partition. Later PR will involve calculating quantities that update the densities based on this stockholder partition.

### DIFF
--- a/cclib/method/ddec.py
+++ b/cclib/method/ddec.py
@@ -274,24 +274,56 @@ class DDEC6(Method):
         self.logger.info("Conditioning charge densities.")
         self.condition_densities()
 
+        # STEP 4-7
+        # Optimize weights on each grid point
+        kappa = [0.0 for x in self.data.atomnos]
+        update_kappa = False
+        self.N_A = []
+        steps = 4
+        while steps < 5:  # change this to 8 after implementing kappa routine
+            self.logger.info("Optimizing charge weights (Step {}).".format(steps))
+            self.N_A.append(self._calculate_w_and_u())
+
+            # Update kappa as described in Figure S4.2 of doi: 10.1039/c6ra04656h
+            if numpy.any([x if x < -1e-5 else 0 for x in self.N_A[-1]]) and steps != 4:
+                update_kappa = True
+            if (
+                steps > 5
+                and update_kappa
+                and numpy.diff(self.N_A)[-1] < 1e-5
+                and numpy.diff(self.N_A)[-2] < 1e-5
+            ):
+                update_kappa = False
+                kappa = [0.0 for x in self.data.atomnos]
+
+            # TODO: EXIT LOOP HERE for steps = 7
+            if not update_kappa:
+                steps = steps + 1
+            else:
+                # `steps` not incremented in this case
+                # First update kappa based on equation 93
+                kappa_new = kappa - self.N_A / self.u_A
+                kappa = [x if x > 0 else 0.0 for x in kappa_new]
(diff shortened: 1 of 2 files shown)


cclib/method/ddec.py
```diff
@@ -590,9 +626,73 @@ def condition_densities(self):
...
-    def _phiai(self, ya, bigphiAI, atomi):
-        """ Evaluate phi_A^I based on equation S100
+    def calculate_w_and_u(self):
+        """ Calculate weights placed on each integration grid point
+            [STEP 4-7]
         """
+        # From equation 67, w_A(r_A) = self._cond_density
+        # From equation 8, W(r) = self.rho_cond
+        xshape, yshape, zshape = self.chgdensity.data.shape
+        atoms = len(self.data.atomnos)
```
berquist, line 636: `atoms` doesn't indicate that this is a count, prefer
```suggestion
        natoms = len(self.data.atomnos)
```

cclib/method/ddec.py
```diff
@@ -590,9 +626,73 @@ def condition_densities(self):
             # Make tau monotonic decreasing
             self.tau[atomi] = numpy.maximum.accumulate(self.tau[atomi][::-1])[::-1]
 
-    def _phiai(self, ya, bigphiAI, atomi):
-        """ Evaluate phi_A^I based on equation S100
+    def calculate_w_and_u(self):
```
berquist, line 629: I think this method should be private.

cclib/method/ddec.py
```diff
@@ -274,6 +274,38 @@ def calculate(self, indices=None, fupdate=0.05):
...
+        # Optimize weights on each grid point
+        kappa = [0.0 for x in self.data.atomnos]
+        update_kappa = False
+        self.N_A = []
+        steps = 4
+        while steps < 5:  # change this to 8 after implementing kappa routine
+            self.logger.info("Optimizing charge weights (Step {}).".format(steps))
+            self.N_A.append(self.calculate_w_and_u())
+
+            # Update kappa as described in (4) of doi: 10.1039/c6ra04656h
```
berquist, line 287: (4) doesn't look like the right reference.

cclib/method/ddec.py
mscho527: Yes. I've made the necessary changes. Thanks!

cclib/method/ddec.py
mscho527: I've made this private. Thanks!

cclib/method/ddec.py
mscho527: I've updated this to a correct reference. Thanks!

cclib/method/ddec.py
```diff
@@ -274,24 +274,56 @@ def calculate(self, indices=None, fupdate=0.05):
...
+            if (
+                steps > 5
+                and update_kappa
+                and numpy.diff(self.N_A)[-1] < 1e-5
+                and numpy.diff(self.N_A)[-2] < 1e-5
+            ):
+                update_kappa = False
+                kappa = [0.0 for x in self.data.atomnos]
+
+            # TODO: EXIT LOOP HERE for steps = 7
```
shivupa, line 299: if you change the if statement below to `if update_kappa and steps < 7:` then it will just exit the loop if steps == 7

mscho527: I was trying to wait until the full implementation of steps 4-7 (with PR921) before I add the exit conditions here. The exit condition is present in PR921 now. Thanks!

PR #922, opened 2020-08-13
Add a few more tests for DDEC6

mscho527:
wip


berquist:
Poke for rebase.

test/method/testddec.py
```diff
@@ -164,3 +165,31 @@ def test_water_charges(self):
...
+        """
+
+        h2path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "h2.out")
+        data = ccread(h2path)
+        vol = volume.Volume((-2, -2, -2), (2, 2, 2), (0.2, 0.2, 0.2))
+        analysis = DDEC6(data, vol, os.path.dirname(os.path.realpath(__file__)))
+        analysis.calculate()
+
+        self.assertAlmostEqual(numpy.sum(analysis.fragcharges), 0, delta=1e-2)
+        self.assertAlmostEqual(analysis.fragcharges[0], analysis.fragcharges[1], delta=1e-6)
```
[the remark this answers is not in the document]
berquist: Come to think of it, is it expected that these charges are both
1. non-negligible and
2. the same sign?
It should actually be the `sum` test that has a very tight threshold, and the opposite test would be comparing the absolute value of the difference to zero.

test/method/testddec.py
```diff
@@ -164,3 +165,31 @@ def test_water_charges(self):
...
+        copath = os.path.join(os.path.dirname(os.path.realpath(__file__)), "co.out")
+        data = ccread(copath)
+        vol = volume.read_from_cube(
+            os.path.join(os.path.dirname(os.path.realpath(__file__)), "co.cube")
+        )
+        analysis = DDEC6(data, vol, os.path.dirname(os.path.realpath(__file__)))
+        analysis.calculate()
+
+        self.assertAlmostEqual(numpy.sum(analysis.fragcharges), 0, delta=1e-2)
+        assert_allclose(analysis.fragcharges, [0.13221636, -0.13903595], atol=1e-3)
```
berquist, line 195: Intuitively, this answer makes more sense than the H2 answer. It's still a little concerning that there is such a deviation from a total of zero (-0.00682). What does chargemol give?

mscho527: This should most likely come from the fact that `self.charge_density.integrate()` returns 14.006876594937234. I've added a note about this. Thanks.

`chargemol` seems to be free from this because they normalize the density to total number of electrons before partitioning them (as in `module_DDEC6_valence_iterator.f08:149`). I'm not sure if this normalization step should be added given that this was not mentioned in the paper?

test/method/testddec.py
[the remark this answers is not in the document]
mscho527: The charges for H2 should just be zero. Using much denser grid (spacing of 0.1 rather than 0.2 which is the cube file included in the test) gives [0.00046066, 0.00046066].

I've modified this test so that the delta is smaller, and left a note why the values are not exactly zero in the test.

test/method/testddec.py
mscho527: As a follow-up, using finer grid (spacing of 0.05) reduces the difference:

```
>>> ddec.fragcharges
array([ 0.13149808, -0.13052562])
>>> ddec.fragcharges.sum()
0.0009724637203740727
>>> ddec.charge_density.integrate()
13.999027536279629
```

Issue #929, opened 2020-08-19
Clean Up DDEC6


shivupa:
The current functions are confusing. Restructure and rename for clarity.

shivupa:
Closed by #930

PR #930, opened 2020-08-20
Clean up DDEC6 code

mscho527:
_Related Issue: #885
Depends upon: #928_

This PR cleans up DDEC6 code by

1. Separating out duplicate code (phi optimization routine in step 3 and 4-6) into a function
2. Add more comments on the phi optimization routine
3. Modify order of class functions
4. Separate out if logic that determines update_kappa

Attempts to address issue #929

### DIFF
--- a/cclib/method/ddec.py
+++ b/cclib/method/ddec.py
@@ -113,14 +113,6 @@ class DDEC6(Stockholder):
             )
         )
 
-        # Notify user about the total charge in the density grid
-        integrated_density = self.charge_density.integrate()
-        self.logger.info(
-            "Total charge density in the grid is {}. If this does not match what is expected, using a finer grid may help.".format(
-                integrated_density
-            )
-        )
-
         # * STEP 1 *
         # Carry out step 1 of DDEC6 algorithm [Determining reference charge value]
         # Refer to equations 49-57 in doi: 10.1039/c6ra04656h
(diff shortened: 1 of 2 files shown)


cclib/method/ddec.py
```diff
@@ -121,6 +121,22 @@ def calculate(self, indices=None, fupdate=0.05):
...
+        )
+
+        # Notify user about the total charge in the density grid
+        integrated_density = self.charge_density.integrate()
+        self.logger.info(
+            "Total charge density in the grid is {}. If this does not match what is expected, using a finer grid may help.".format(
+                integrated_density
+            )
+        )
+
```
berquist, line 139: Both of these should be removed, since the check already exists on line 108. Make sure to remove the already-existing extra one too.

mscho527: I've removed these. Thanks for letting me know!

test/method/testddec.py
```diff
@@ -48,9 +48,7 @@ def testmissingrequiredattributes(self):
             vol = volume.Volume((-4, -4, -4), (4, 4, 4), (0.2, 0.2, 0.2))
             delattr(self.data, missing_attribute)
             with self.assertRaises(MissingAttributeError):
-                trialBader = DDEC6(
-                    self.data, vol, os.path.dirname(os.path.realpath(__file__))
-                )
+                trialBader = DDEC6(self.data, vol, os.path.dirname(os.path.realpath(__file__)))
```
shivupa, line 51: can you run black on this file so there are no changes in the test?

mscho527: I've re-run black on both files to make sure. It seems to me that the PR before this wasn't cleansed with black. Thanks!

berquist:
Travis isn't timing out because of the tests, but because of uploading the new archive. They're almost 6.5 GB in size. They might be causing more problems than they solve. I've created https://github.com/cclib/cclib/issues/936, which needs to be fixed before the 1.6.4 release, and in the meantime will merge this.

PR #938, opened 2020-08-30
Fix a typo in Bader method

mscho527:
This PR fixes a small typo in Bader method.

The line of code that contains the fix is executed only when one or more positions in the walker that identifies Bader spaces have gradients of exactly (0.0, 0.0, 0.0).


berquist:
@mscho527 Poke for rebase.

PR #940, opened 2020-08-30
Fix a typo in DDEC

mscho527:
This PR fixes a small typo in DDEC method.

The line of code that contains the fix is executed only when kappa is updated. The variable `cycle` had been removed in previous PRs.


berquist:
@mscho527 Poke for rebase, but only so that the CI build is on top of current master.

Issue #942, opened 2020-09-02
Q-Chem parser bug for MOs


amandadumi:
There is a bug with the Q-Chem parser when parsing MOs only if the atom symbol consists of two letters.
I think this was overlooked in testing as the only case with an atom with two symbols is the MoCl4_sp case, though it does not contain any MOs.
A small secondary point is that the [MOCl4_sp for Qchem/5.2 ](https://github.com/cclib/cclib/blob/master/data/QChem/basicQChem5.1/MoOCl4_sp.out) has an error due to a change in how Q-Chem handles ECPs.

The bug can be recreated with the [si_qchem_out.txt](https://github.com/cclib/cclib/files/5159466/si_qchem_out.txt)
Running `ccget --list si_qchem_out.txt` returns:

```bash
Attempting to read si_qchem_out.txt
[QChem si_qchem_out.txt ERROR] Encountered error when parsing.
[QChem si_qchem_out.txt ERROR] Last line read:    1  Si    s        1.00042  -0.26371   0.00000  -0.00000   0.00000   0.07477

Traceback (most recent call last):
  File "/home/amandad/.local/bin/ccget", line 33, in <module>
    sys.exit(load_entry_point('cclib', 'console_scripts', 'ccget')())
  File "/home/amandad/Documents/side_project/cclib/cclib/scripts/ccget.py", line 179, in ccget
    data = ccread(filename, **kwargs)
  File "/home/amandad/Documents/side_project/cclib/cclib/io/ccio.py", line 182, in ccread
    return log.parse()
  File "/home/amandad/Documents/side_project/cclib/cclib/parser/logfileparser.py", line 335, in parse
    self.extract(inputfile, line)
  File "/home/amandad/Documents/side_project/cclib/cclib/parser/qchemparser.py", line 1267, in extract
    mocoeffs = self.parse_matrix_aonames(inputfile, self.nbasis, self.norbdisp_alpha_aonames)
  File "/home/amandad/Documents/side_project/cclib/cclib/parser/qchemparser.py", line 333, in parse_matrix_aonames
    aoname = ''.join([name, '_', shell.upper()])
TypeError: sequence item 0: expected str instance, NoneType found
```

PR #943, opened 2020-09-02
Fixing bug with two letter chemical symbols

amandadumi:
This is to address the issue in #942 . I'm not sure what the best approach is incorporating a test for this behavior would be. Any advice would be appreciated!

### DIFF
--- a/cclib/parser/qchemparser.py
+++ b/cclib/parser/qchemparser.py
@@ -303,7 +303,7 @@ cannot be determined. Rerun without `$molecule read`."""
         line = next(inputfile)
         assert len(line.split()) == min(self.ncolsblock, ncols)
         colcounter = 0
-        split_fixed = utils.WidthSplitter((4, 3, 5, 6, 10, 10, 10, 10, 10, 10))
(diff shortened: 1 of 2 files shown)


berquist:
You should make a PR with the silicon atom output file to the cclib-data repo, and add a regression test in https://github.com/cclib/cclib/blob/master/test/regression.py. As for what goes in the test, since it's failing even on parsing the first line in the MO coefficient block, I'd probably check the exact value of a single element from that line.

langner:
> and add a regression test in https://github.com/cclib/cclib/blob/master/test/regression.py. As for what goes in the test, since it's failing even on parsing the first line in the MO coefficient block, I'd probably check the exact value of a single element from that line.

This is the last bit missing for this.

amandadumi:
This should be good to go! Let me know if there are any issues I've overlooked or if you want the test to check anything else.

langner:
Thanks! Looks good at a glance - I think cclib/cclib-data#137 needs to go in first.

langner (approved):

berquist (approved):

PR #945, opened 2020-09-08
Fix handling of open shell systems in moldenwriter and wfxwriter

shivupa:
#937 showed that the moldenwriter and wfx writer needed work to handle open shell systems. This PR properly calculated the number of electrons in the wfx writer and removes a loop over the multiplicity.

PR #948, opened 2020-09-22
Allow writing of natural orbitals to the molden file instead of MOs

shivupa:
This will need to be rebased on #945 

This adds a flag to write natural orbitals if they are present. Since the corresponding eigenvalues are occupation numbers rather than mo energies, the mo energies are just set to the occupation numbers. Let me know if this sounds inappropriate.

To Do:
Add test

### DIFF
--- a/cclib/io/moldenwriter.py
+++ b/cclib/io/moldenwriter.py
@@ -135,7 +135,7 @@ class MOLDEN(filewriter.Writer):
         return mocoeffs
 
 
-    def _mo_from_ccdata(self):
+    def _mo_from_ccdata(self, mosyms, moenergies, mocoeffs, mooccs):
         """Create [MO] section.
 
         Sym= symmetry_label_1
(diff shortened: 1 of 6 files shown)


langner:
@shivupa Did you want to squeeze this into v.1.6.4?

shivupa:
No I think it's fine to have this go to v1.6.5

berquist:
Not having looked at the Molden "spec" in a while, do you know if `[MO]` is the only MO section, or is there another dedicated section that can be used?

shivupa:
It seems like that isn't an option https://www.cmbi.ru.nl/molden/molden_format.html
Also it seems like other code write non-MOs to the MO tag https://github.com/psi4/psi4/issues/504 and associated issues.

berquist (changes requested):
Apparently I never submitted this review...

berquist:
Poke.

berquist (approved):

test/regression.py
```diff
@@ -764,13 +764,26 @@ def testnoparseGAMESS_WinGAMESS_H2O_def2SVPD_triplet_2019_06_30_R1_out(filename)
     """
     data = ccread(os.path.join(__filedir__,filename))
     writer = moldenwriter.MOLDEN(data)
+    moenergies = data.moenergies
```
berquist, line 767: This bit looks duplicated with `testmoldenwriter`.  Duplication in test code is often a good idea, but do you think filling occupations from `homos` like this is something a user would want?  It could be a utility method/function if so.  (The same question for symmetries is less clear to me.)

cclib/io/moldenwriter.py
```diff
@@ -215,15 +184,38 @@ def generate_repr(self):
...
+            for i in range(len(self.ccdata.homos)):
+                mooccs[i][0:self.ccdata.homos[i]+1] = occval
+        elif self.naturalorbitals and hasattr(self.ccdata, 'nooccnos') and hasattr(self.ccdata, "nocoeffs"):
+            moenergies = numpy.array([self.ccdata.nooccnos])
+            mocoeffs = numpy.array([self.ccdata.nocoeffs])
+            mooccs = numpy.array([self.ccdata.nooccnos])
+        elif self.naturalorbitals and hasattr(self.ccdata, 'nsooccnos') and hasattr(self.ccdata, "nsocoeffs"):
+            moenergies = self.ccdata.nsooccnos
+            mocoeffs = self.ccdata.nsocoeffs
+            mooccs = self.ccdata.nsooccnos
```
berquist, line 204: These are the same conditional.  Which is correct?

cclib/io/moldenwriter.py
```diff
@@ -215,15 +184,38 @@ def generate_repr(self):
...
+
+        mosyms = None
+        if hasattr(self.ccdata, 'mosyms') and not self.naturalorbitals:
```
berquist, line 207: Is it impossible for natural orbitals to belong to a single irrep?  Worded slightly differently, if a density matrix comes from a calculation with greater than C1 symmetry, won't its NOs also unambiguously belong to irreps within that point group?

cclib/io/moldenwriter.py
```diff
@@ -215,15 +184,38 @@ def generate_repr(self):
...
+            mosyms = self.ccdata.mosyms
+        else:
+            mosyms = numpy.full_like(moenergies, 'A', dtype=str)
 
+        # Either both [GTO] and [MO] should be present or none of them.
```
berquist, line 212: I know this comment is carried over, but how so?  You could print the basis but not MO coefficients and vice versa, then just dump what you have to MOLDEN.  (If this comment is wrong, then we can do a separate PR.)

cclib/io/moldenwriter.py
```diff
@@ -215,15 +184,38 @@ def generate_repr(self):
...
+        if hasattr(self.ccdata, 'gbasis') and mosyms is not None and moenergies is not None and mocoeffs is not None and mooccs is not None:
```
berquist, line 213: It may not shorten the line but I would do a `and all(attr is not None for attr in (mosyms, moenergies, ...)):`.

berquist (changes requested):
A bunch of questions because you're a lot more familiar with MOLDEN and natural orbitals than I am.  The only things that needs fixing is the conditional.

test/regression.py
shivupa: i was thinking of some sort of helper function just cause it was annoying for me. so maybe ill create that as an underscored function in molden writer.

cclib/io/moldenwriter.py
shivupa: natural orbitals vs natural spin orbitals

cclib/io/moldenwriter.py
shivupa: yeah definitely. some codes don't symmetrize the NOs, but as far as I could tell, we don't parses 'nosyms'

cclib/io/moldenwriter.py
shivupa: I think it would make sense to have GTO present without MO. But without the GTO present, MOs don't have a lot of meaning?

cclib/io/moldenwriter.py
shivupa: i like that thanks

cclib/io/moldenwriter.py
berquist: I am blind

cclib/io/moldenwriter.py
berquist: From a usability point, sure.  But there's no technical limitation that prevents MOs from appearing without the GTO section.

berquist (approved):

berquist (changes requested):
Something I just noticed: `_syms_coeffs_occs_energies_from_ccdata_for_moldenwriter` returns `syms, coeffs, occs, energies`.  `_mo_from_ccdata` takes `mosyms, moenergies, mocoeffs, mooccs`.  Can you make them consistent?  I'm not sure what ordering is best.

shivupa:
I put it in the order it is written to the file

PR #954, opened 2020-09-29
Speed up `Volume.electrondensity_spin` and `Volume.wavefunction`

berquist:
Being able to remove explicit indexing and passing single points for calculating basis function amplitudes on a grid brought this code from 124 to 7 seconds, now almost completely in the Bader code:
```python
from pathlib import Path
from cclib.method import Bader, volume
from cclib.io import ccread
import numpy as np

path_data = Path("/home/eric/development/cclib_berquist/test/method/hf.out")
data = ccread(str(path_data))
vol = volume.Volume((-6, -6, -6), (6, 6, 6), (0.2, 0.2, 0.2))
analysis = Bader(data, vol)
analysis.calculate()
np.testing.assert_allclose(analysis.fragcharges, [10.58407202, 0.35639819], atol=1.0e-8)
```
and the non-regression test suite time from 547 to 214 seconds.

The current breakdown for non-regression test times is:
```
55.73s call     test_method.py::BaderTest::test_symms_benzene                                                                                                                
33.80s call     test_method.py::HirshfeldTest::test_water_charges                                                                                                            
18.47s call     test_method.py::HirshfeldTest::test_chgsum_co
7.84s call     test_method.py::DDEC6Test::test_water_charges
7.72s call     test_data.py::test_all[parsers0-modules0-True-False-40-True-True]
6.30s call     test_method.py::HirshfeldTest::test_chgsum_h2
5.72s call     test_method.py::BaderTest::test_chgsum_hf
5.33s call     test_method.py::DDEC6Test::test_chgsum_co
2.77s call     test_method.py::DDEC6Test::test_chg_nh3
1.68s call     test_method.py::BaderTest::test_val
```
and everything else took less than a second each.

### DIFF
--- a/cclib/method/volume.py
+++ b/cclib/method/volume.py
@@ -62,10 +62,23 @@ if _found_pyquante:
 
         return bfs
 
-    # Small wrapper PyQuante & pyquante2 function that evaluates basis function on a given point
-    # Used in both `wavefunction` and `electrondensity`
     def pyamp(bfs, bs, points):
-        # 1D numpy array with size 1 is returned from __call__ here.
+        """Wrapper for evaluating basis functions at one or more grid points.
+
+        Parameters
+        ----------
+        bfs : list
+            List of PyQuante 1 `CGBFs`s (contracted Gaussian basis functions).
+        bs : int
+            Index into the list of CGBFs for the basis function to evaluate.
+        points : numpy.ndarray
+            An [n, 3] array of `n` Cartesian grid points on which to evaluate the basis function.
+
+        Returns
+        -------
+        out : numpy.ndarray
+            An [n, ] array of the requested basis function's value on each grid point.
+        """
         mesh_vals = numpy.zeros(len(points))
         for i in range(len(points)):
             mesh_vals[i] = bfs[bs].amp(points[i][0], points[i][1], points[i][2])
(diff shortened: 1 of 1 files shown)


langner:
Did you want to get this into v1.6.4?

berquist:
I do want this for 1.6.4. I think it's fine as-is, but was trying to speed up the Python 2 implementation. Right now, Travis takes < 12 minutes for Python 3 and 15 minutes for Python 2. I think this is because the PyQuante 1 code isn't vectorized. I monkey-patched it with the vectorized implementation taken from PyQuante 2, only to find that the DDEC6 test for H2 never terminates: the loop inside https://github.com/cclib/cclib/blob/67be86feb8d97eb123c268b86323afec008d6dc5/cclib/method/ddec.py#L699 when called from `condition_densities` because it stays stuck at zero, where in the Python 3 implementation it's just barely above (~1e-17). I've managed to trace this back to the CGBFs themselves. There are two issues:
1. PyQuante 1 uses a different Angstrom-to-bohr conversion factor than in `cclib.utils.convertor`. I fixed this by removing the need for `cclib2pyquante` like this PR already did for PyQuante 2/Python 3.
2. The contraction coefficients are slightly different. Since they're used directly from `ccdata.gbasis`, the only explanation is that the routine which normalizes the contraction coefficients is different. There's actually a comment in the PyQuante 2 code wondering if it's correct. Though, this doesn't explain why switching to a vectorized method for putting basis functions on a grid would change the result to begin with.

Unless we want to speed up CI time for a language that will eventually be removed anyway, I won't debug this further and will just leave the implementation as-is.

shivupa:
Yeah it seems like wasted effort to try to optimize the python 2 further. I can review this later today.

shivupa (approved):

langner:
> Yeah it seems like wasted effort to try to optimize the python 2 further. I can review this later today.

Agreed. At this point I'm OK with _slowing down_ Python 2 :)

PR #955, opened 2020-09-30
Print tracebacks by default when running regression tests

shivupa:
This is a draft PR to add the traceback flag to pytest as mentioned [here](https://github.com/cclib/cclib/pull/945#issuecomment-701054430) .

### DIFF
--- a/test/regression.py
+++ b/test/regression.py
@@ -3294,7 +3294,7 @@ def make_regression_from_old_unittest(test_class):
     return old_unit_test
 
 
-def test_regressions(which=[], opt_traceback=False, regdir=__regression_dir__, loglevel=logging.ERROR):
+def test_regressions(which=[], opt_traceback=True, regdir=__regression_dir__, loglevel=logging.ERROR):
 
     # Build a list of regression files that can be found. If there is a directory
     # on the third level, then treat all files within it as one job.


shivupa:
@berquist I am confused about this now. I think running pytest without the `--terse` flag gives what I would expect to happen with the `--traceback` flag. 

`python -m pytest -v  --capture=no --cov=cclib --cov-report=term --cov-report=html  test` vs `python -m pytest -v  --capture=no --cov=cclib --cov-report=term --cov-report=html  --terse test`

On another note the reason the traceback flag currently errors is because the `test_all` function in `test/test_data.py` doesn't accept a traceback flag. It does accept terse and silent flags which I think behave similarly. Regression tests do accept a traceback flag.

travis/run_pytest.bash
```diff
@@ -5,8 +5,8 @@
 
 set -euxo pipefail
 
-python -m pytest -v --capture=no --cov=cclib --cov-report=term --cov-report=html --terse test
+python -m pytest -v --traceback --capture=no --cov=cclib --cov-report=term --cov-report=html --terse test
```
berquist, line 8: I would stick this at the end as an attempt to differentiate between `pytest` flags and cclib flags.

test/conftest.py
```diff
@@ -56,6 +57,7 @@ def pytest_generate_tests(metafunc):
         metafunc.parametrize("modules", [{p: all_modules[p] for p in module_names}])
         metafunc.parametrize("terse", [metafunc.config.getoption("--terse")])
         metafunc.parametrize("silent", [metafunc.config.getoption("--silent")])
+        metafunc.parametrize("traceback", [metafunc.config.getoption("--traceback")])
```
berquist, line 60: This actually needs to go under a new block, `if metafunc.function.__name__ == "test_regressions"` and I think it'll be `opt_traceback`: https://github.com/cclib/cclib/blob/67be86feb8d97eb123c268b86323afec008d6dc5/test/regression.py#L3294
and you'll probably get the cryptic "default argument" error I was getting.

berquist:
We agreed that a better solution is to remove the option and change the default of `opt_traceback` to `True`.

berquist (approved):

PR #956, opened 2020-10-11
Extend pyscf bridge to handle MOs

amandadumi:
This pull request aims to enable the conversion of additional cclib data into PySCF formatted objects.
There are two ways I imagined going about this. The first would be to create the objects from the ground up in cclib. The second would take advantage of the Molden writer in cclib and Molden parser in PySCF. 
I've gone with the latter since the bridge is compact and the Molden parser of PySCF handles AO ordering and normalization requirements ([specifically, the Cartesian orbitals are not normalized in PySCF.](https://sunqm.github.io/pyscf/_modules/pyscf/tools/molden.html)) 

One change that was required  in the Molden writer. The addition of the SYM label of each MO as 'A' was added in the case the they are not present in the cclib data. This was to accommodate PySCF  as they use the SYM label to distinguish between orbitals. 

What are your thoughts?

### DIFF
--- a/cclib/bridge/__init__.py
+++ b/cclib/bridge/__init__.py
@@ -27,4 +27,6 @@ if find_package("ase"):
 if find_package("iodata"):
     from cclib.bridge.cclib2horton import makehorton
 
+if find_package("pyscf"):
+    from cclib.bridge.cclib2pyscf import makepyscf, makepyscf_mos
 del find_package
--- a/cclib/bridge/cclib2pyscf.py
+++ b/cclib/bridge/cclib2pyscf.py
@@ -13,6 +13,9 @@ import numpy as np
 l_sym2num = {"S": 0, "P": 1, "D": 2, "F": 3, "G": 4}
 
 
+class MissingAttributeError(Exception):
+    pass
+
 _found_pyscf = find_package("pyscf")
 if _found_pyscf:
     from pyscf import gto
@@ -26,14 +29,24 @@ def _check_pyscf(found_pyscf):
 def makepyscf(data, charge=0, mult=1):
     """Create a Pyscf Molecule."""
     _check_pyscf(_found_pyscf)
+    # if hasattr(data, "gbasis"):
+    #     basis = 
+    inputattrs = data.__dict__
+    required_attrs = {"atomcoords", "atomnos"}
+    missing = [x for x in required_attrs if not hasattr(data, x)]
+    if missing:
+        missing = " ".join(missing)
+        raise MissingAttributeError(
+            "Could not create pyscf molecule due to missing attribute: {}".format(missing)
+        )
     mol = gto.Mole(
-        atom=[
-            ["{}".format(data.atomnos[i]), data.atomcoords[-1][i]]
-            for i in range(data.natom)
+    atom=[
+        ["{}".format(data.atomnos[i]), data.atomcoords[-1][i]]
+        for i in range(data.natom)
         ],
-        unit="Angstrom",
-        charge=charge,
-        multiplicity=mult,
+    unit="Angstrom",
+    charge=charge,
+    multiplicity=mult  
     )
     inputattr = data.__dict__
     pt = PeriodicTable()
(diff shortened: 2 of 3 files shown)


shivupa:
Can you break the change to the molden writer out to a separate PR. That seems necessary and can go in before the additional cclib->pyscf stuff

amandadumi:
Yeah! That's no problem.

cclib/io/moldenwriter.py
```diff
@@ -161,14 +161,15 @@ def _mo_from_ccdata(self):
         if hasattr(self.ccdata, 'mosyms'):
             has_syms = True
             syms = self.ccdata.mosyms
+        else:
```
berquist, line 164: Can you rebase this on top of master? These changes already got merged in from your other PR.

cclib/bridge/cclib2pyscf.py
```diff
@@ -22,12 +24,55 @@ def _check_pyscf(found_pyscf):
...
     _check_pyscf(_found_pyscf)
-    mol = gto.Mole(
-        atom = [['{}'.format(atomnos[i]),atomcoords[i]] for i in range(len(atomcoords))],
-        unit="Angstrom",
-        charge=charge,
-        multiplicity=mult
-    )
-    return  mol
-  
+    # See what can be parsed:
```
berquist, line 27: Can you revert the changes to `makepyscf`? It looks like they aren't used right now.

cclib/bridge/cclib2pyscf.py
```diff
@@ -22,12 +24,55 @@ def _check_pyscf(found_pyscf):
...
+    ----
+    data: cclib data object from parsed output file
+
+    Returns:
+    ----
+    mol: a pyscf molecule object
+        a molecule object that consists of available data in Molden file.
+    pyscf_data: dictionary
+        a dictionary of additional values returned in an appropriate format for use in PySCF. 
+        If available in ccdata, the following objects are available in the dictionary, otherwise, values are None
```
berquist, line 57: Missing a listing of attributes?

cclib/bridge/cclib2pyscf.py
```diff
@@ -22,12 +24,55 @@ def _check_pyscf(found_pyscf):
...
+    temp.write(mldn_obj)
+    temp.seek(0)
+    (
+        mol,
+        pyscf_data["mo_energy"],
+        pyscf_data["mo_coeff"],
+        pyscf_data["mo_occ"],
+        pyscf_data["irrep_labels"],
+        pyscf_data["spins"],
+    ) = tools.molden.load(temp.name)
```
berquist, line 73: I enjoy this hack. I don't think we should test PySCF heavily, but can you add a small test to check the contents of `pyscf_data` members against the cclib equivalents?

berquist (changes requested):

cclib/bridge/cclib2pyscf.py
amandadumi: Do you mean to define the data types within the dictionary argument?

cclib/bridge/cclib2pyscf.py
amandadumi: yes! That's no problem, just wanted to make sure the hack was acceptable before writing the tests. Should have these done shortly. Thanks for the feedback.

cclib/bridge/cclib2pyscf.py
berquist: It was more of the docstring implying that there would be some description of the attributes available in the dictionary. But I think I understand on second read: if the right attributes are present in the `ccData` object, the dict will be populated, otherwise it'll be empty (not `None`, or are they empty lists? not clear from the pyscf code).

Actually, rather than going through the MOLDEN writer (which may be lossy), why not copy https://github.com/pyscf/pyscf/blob/7a57d2f37184ea4cd3b52a590c945197b084f66d/pyscf/tools/molden.py#L272 here?

cclib/io/moldenwriter.py
amandadumi: Rebased on top of master, no changes to `moldenwriter` in this pull request.

cclib/bridge/cclib2pyscf.py
amandadumi: That sounds like a a plan. 
I think what will make sense is to first add the ability to the `make_pyscf` function to handle setting up the basis of the molecule if `ccdat['gbasis']` is present, as the basis information is stored within the molecule object. This second function will serve to set up the  MO related internally. 

I'll just need a bit of time to figure out the cleanest way to convert the data to PySCF.

amandadumi:
Hello! I was able to get the MOs parsed and working with PySCF. However, these changes are based off the work in #975 , so please feel free to give any feedback on the MO portion, but **do not merge** until 975 has been addressed.

shivupa:
Hi sorry for the delay #975 is now in!

berquist:
I'm getting the same error locally with pyscf 1.7.6 from their conda channel.

cclib/bridge/__init__.py
```diff
@@ -27,4 +27,6 @@
 if find_package("iodata"):
     from cclib.bridge.cclib2horton import makehorton
 
+if find_package("pyscf"):
+    from cclib.bridge.cclib2pyscf import makepyscf, makepyscf_mos
```
langner, line 31: Do you think we will need to continue to support two version, or do you see these merging in the future?

amandadumi: This is a good questions. Since pyscf stores the molecule object separate from the mo_coeff object, I thought this might be a cleaner solution in cases where one may only want a molecule object. The alternative could be the makepyscf function returning either one or two things. The molecule object could be returned in the case that mocoeffs are not present in the cclib data. Otherwise the function could return both the molecule object and the mo_coeff object.

Is there a way you would prefer or maybe another way I haven't considered?

cclib/bridge/cclib2pyscf.py
```diff
@@ -54,7 +67,65 @@ def makepyscf(data, charge=0, mult=1):
...
+            mo_coeff_b = np.einsum('i,ij->ij', np.sqrt(1/s.diagonal()), mo_coeff_b)
+            mo_occ = np.zeros((2,ccdata.nmo))
+            mo_occ[0,:ccdata.homos[0]+1] = 1
+            mo_occ[1,:ccdata.homos[1]+1] = 1
+            mo_coeffs = np.array([mo_coeff_a,mo_coeff_b])
+            if hasattr(ccdata, 'mosyms'):
+                mo_syms = ccdata.mosyms
+            else:
+                mo_syms = np.full_like(ccdata.moenergies, 'A', dtype=str)
+            mo_energies = ccdata.moenergies
```
berquist, line 122: This is in eV, not hartree, so you probably want to convert here and mention the units in the docstring.

amandadumi: Oh yes, thank you! Will do.

cclib/bridge/__init__.py
berquist: I think this is fine, since I operate under the assumption that PySCF is a rapidly moving target...

berquist (approved):
Thanks for being patient! Make more PRs to update this in the even that PySCF changes...

PR #957, opened 2020-10-11
Add default MO symmetry for Molden

amandadumi:
This pull request adds a default symmetry label to the Molden writer. This was to accommodate PySCF since when loading a Molden file,  the SYM label is used to distinguish between orbitals.

This is a change needed for #956 .


shivupa (approved):

shivupa:
Minor change and tests are passing. I am going to merge

PR #963, opened 2020-10-21
Clean up scipy imports

langner:


shivupa (approved):

cclib/method/nuclear.py
```diff
@@ -21,7 +21,7 @@
 
 _found_scipy = find_package("scipy")
 if _found_scipy:
-    import scipy.constants as spc
```
berquist, line 24: I missed this PR. I realize that this change now fits with cclib conventions of not shortening imports (`import numpy as np`), but I don't agree with our convention here: it goes against import styles in the rest of the ecosystem, and especially in this case going from `spc` to `scipy.constants` makes the code harder to read. I don't think it should be changed back, but I wouldn't have made this change.

shivupa: Sorry about that, that's on me. I should have let it stay open for discussion. I was in the mindset of cclib conventions, but I get where you are coming from.

langner: Just read this while doing the release. Please consider opening an issue and doing this in bulk, so we set the convention we want and stick with it.

PR #965, opened 2020-10-26
Add Iterative Stockholder method [WIP]

mscho527:
_Work in progress_

This PR adds Iterative Stockholder method using the Stockholder abstract class that exists in cclib.

Method is explained in [https://doi.org/10.1063/1.3243863](https://doi.org/10.1063/1.3243863) and [https://doi.org/10.1063/1.2715563](https://doi.org/10.1063/1.2715563)
Testing may be done against values obtained from horton 2.1.0.

### DIFF
--- a/cclib/method/iterative_stockholder.py
+++ b/cclib/method/iterative_stockholder.py
@@ -1,0 +1,105 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2020, the cclib development team
+#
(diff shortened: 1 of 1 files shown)


langner:
@mscho527 Were you planning on finishing this up?

mscho527:
@langner Yes -- I'm planning to go over this during the Thanksgiving break :)

PR #972, opened 2020-11-05
Remove Python 2 support

berquist:
Closes #518

PR #975, opened 2020-11-10
extending pyscf bridge to basis

amandadumi:
Extending the PySCF bridge to instantiate a basis set if it is available in cclib data. 

The test checks the energy of RHF in the case the basis is provided and checks in the case of no basis being provided that the PySCF default is initialized.

### DIFF
--- a/cclib/bridge/cclib2pyscf.py
+++ b/cclib/bridge/cclib2pyscf.py
@@ -7,7 +7,11 @@
 
 """Bridge for using cclib data in PySCF (https://github.com/pyscf/pyscf)."""
 
-from cclib.parser.utils import find_package
+from cclib.parser.utils import find_package, PeriodicTable
+import numpy as np
+
+l_sym2num = {"S": 0, "P": 1, "D": 2, "F": 3, "G": 4}
+
 
 _found_pyscf = find_package("pyscf")
 if _found_pyscf:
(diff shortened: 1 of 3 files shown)


cclib/bridge/cclib2pyscf.py
```diff
@@ -8,6 +8,11 @@
 """Bridge for using cclib data in PySCF (https://github.com/pyscf/pyscf)."""
 
 from cclib.parser.utils import find_package
+import qcelemental as qe
```
langner, line 11: Does this require QCElemental? IIRC we haven't started depending on it yet. I think we will, not sure if there was a plan around that. @berquist ? @shivupa

test/bridge/testpyscf.py
```diff
@@ -8,35 +8,39 @@
 import unittest
 
 import numpy as np
-
+from ..test_data import getdatafile
```
langner, line 11: Is it possible to do this without a relative import?

cclib/bridge/cclib2pyscf.py
amandadumi: Ah yes, This was a question I had too. I saw a few pull request discussing it. I can change this to use the internal conversion from cclib.parser.utils in the mean time though if that is preffered.

test/bridge/testpyscf.py
amandadumi: Hm, I will try! The Horton test is currently also using a relative import which is what I modeled this after, but this is likely possible without it.

cclib/bridge/cclib2pyscf.py
shivupa: if we are planning to use QCElemental anyways (this was my impression) I am fine with it coming in on this PR

test/bridge/testpyscf.py
amandadumi: The import of the test_data module has been added to the init file of the test directory. With this in place, the `getdatafile` function without a relative path. Is this an acceptable solution?

cclib/bridge/cclib2pyscf.py
berquist: I stashed it away when separating out https://github.com/cclib/cclib/pull/972 from https://github.com/cclib/cclib/pull/781. Since this PR doesn't technically depend on on qcel, can you change it to use our built-in class?

amandadumi: Ah, I see. I have replaced the qcelemental call with the built-in PeriodicTable class.

test/bridge/testpyscf.py
```diff
@@ -8,35 +8,39 @@
...
         pyscfmol.cart = True
         pyscfmol.verbose = 0
         pyscfmol.build()
 
-        mhf = pyscfmol.HF(conv_tol=1e-6)
+        mhf = pyscfmol.HF(conv_tol=1e-9)
         en = mhf.kernel()
-        ref = -75.824754602
-        assert abs(en - ref) < 1.0e-6
+        assert abs(convertor(en, "hartree", "eV") - refen) < 1.0e-5
```
berquist, line 39: Can you switch this so that the conversion happens when you se `refen` above?

cclib/bridge/cclib2pyscf.py
```diff
@@ -19,15 +23,38 @@ def _check_pyscf(found_pyscf):
...
+            data.atomnos, return_index=True
+        )  # find unique atoms
+        for idx, i in enumerate(uatoms_idx):
+            curr_atom_basis = data.gbasis[i]
+            for jdx, j in enumerate(curr_atom_basis):
+                curr_l = j[0]
+                curr_e_prim = j[1]
+                new_list = [l_sym2num["{}".format(curr_l)]]
+                new_list += curr_e_prim
+                pt = PeriodicTable()
```
berquist, line 51: Can you put this at the top level, still inside the function but outside any loops?

berquist (changes requested):

test/bridge/testpyscf.py
amandadumi: This should be addressed now!

cclib/bridge/cclib2pyscf.py
amandadumi: Yes, of course. Not sure what I was thinking there. This should be addressed now.

berquist (approved):

shivupa:

Issue #974, opened 2020-11-10
excited states (TDA) from QChem


BalticPinguin:
Dear all, 
I am not sure if it is a matter of version or due to some options; but the `qchemparser` is not able to read-in my coefficients from a TDA-calculations: it expects the form (copy from the source-file)
```` 
    S(  1) --> V(  1) amplitude = -0.3105 alpha
    D( 34) --> S(  1) amplitude =  0.9322 beta
````
While I have
````
    D(    3) --> S(    1) amplitude =  0.9977 beta
````
i.e. there are four spaces between the `(` and `)` while only three are expected; so the indices used to identify the different elements don't match anymore.

I can try to prepare a pull request but wanted first to discuss the strategy used for fixing it: my idea would be to work with regex instead of indices to find the elements; but you probably avoid this with a good reason?

Kind regards,
Hubert

shivupa:
Thanks @BalticPinguin 
What version of QChem was this, and do you have a log file we can use as a regression test? 

As to the regex, I tend to avoid them when possible since they are harder to read, but in this case it makes more sense to use one. @berquist might have more thoughts.

BalticPinguin:
It is version 5.3.1 (sorry, I forgot to write it right away).
The file I have as a sample is this one: 

[trip.txt](https://github.com/cclib/cclib/files/5518671/trip.txt)

langner:
I believe this now fixed. Please reopen if not.

PR #976, opened 2020-11-17
tentative fix for issue #974

srtlg:
I needed a fix for issue #974, too.  Maybe this is acceptable?


berquist (changes requested):
Yes, I think this is also much nicer than the original solution. Since we already have the regression files you submitted, and we're going to go right to Q-Chem 5.3 (https://github.com/cclib/cclib/issues/977), can you remove the 5.2 files here?

BalticPinguin:
Obviously I was too slow writing my PR (which is mostly ready, but uses a less handy approach). :-)

berquist:
The error has nothing to do with your PR, but has been fixed in `master`. Can you rebase?

langner:
We're going to start releasing 1.7 - would be nice to get this in; would need to be in the next few days.

langner:
> Yes, I think this is also much nicer than the original solution. Since we already have the regression files you submitted, and we're going to go right to Q-Chem 5.3 (#977), can you remove the 5.2 files here?

This was addressed. No way for me to mark as resolved, but I think it's fine to merge in.

PR #978, opened 2020-11-18
Fix quoting issue

langner:


berquist (approved):

langner:
Oops, force pushed something by accident to my repo - second try in #979

PR #983, opened 2020-11-21
Switch from Travis CI to GitHub Actions

shivupa:
This is an attempt at adding github actions


langner:
How do we now this works?

shivupa:
Sorry I should have labeled this as WIP. I am still working on figuring out if everything is going to work right/if we can activate the actions for this PR

shivupa:
(Testing locally with [act](https://github.com/nektos/act))

ghutchis:
There should be workflow results at https://github.com/shivupa/cclib/actions 
(e.g., https://github.com/ghutchis/avogadrolibs/actions)

Try changing to:

`on: [push, pull_request]`

berquist:
For now, I would put everything you're sticking in `github_actions` into `.github` instead. We will need to (probably later)
- extract out the non-GitHub Actions specific parts and
- move the general parts to a non-hidden location with a better name. Having the test runners there doesn't make sense, since those are the scripts we use when testing locally as well, in the event that developers can't/don't want to use `act`.

shivupa:
Thanks will do. Is this holding up the release? I am trying to troubleshoot still

langner:
> Thanks will do. Is this holding up the release? I am trying to troubleshoot still

Nope, we can do this for the next one (1.7.1) I'm just swamped somewhat and went on staycation last week.

BTW, really happy to see this moving along.

berquist:
If you see https://github.com/nektos/act/issues/418#issuecomment-727261137, I made https://hub.docker.com/repository/docker/berquist/ubuntu-builder that we can temporarily use locally via `act -P ubuntu-latest=berquist/ubuntu-builder`.

shivupa:
Build is green, but the doc building isn't triggering. Should run according to https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows

langner:
#1008 should help with the new ASE bridge failure

langner:
Also created PR for your branch @shivupa - shivupa/cclib#4

berquist:
I think this is finally ready for review. Codecov is working on the branch, and RTD is going to keep failing until its configuration file is in `master`.

PR #985, opened 2020-12-03
Switch from periodictable to MolSSI/QCElemental

berquist:
The QCElemental part of https://github.com/cclib/cclib/pull/781/.

### DIFF
--- a/cclib/method/nuclear.py
+++ b/cclib/method/nuclear.py
@@ -14,51 +14,20 @@ from cclib.method.calculationmethod import Method
 from cclib.parser.utils import PeriodicTable, convertor, find_package
 
 import numpy as np
+import qcelemental as qcel
+from qcelemental import constants
 
-_found_periodictable = find_package("periodictable")
-if _found_periodictable:
-    import periodictable as pt
-
-_found_scipy = find_package("scipy")
-if _found_scipy:
-    import scipy.constants
-
-
-def _check_periodictable(found_periodictable: bool) -> None:
-    if not _found_periodictable:
-        raise ImportError("You must install `periodictable` to use this function")
-
-
-def _check_scipy(found_scipy: bool) -> None:
-    if not _found_scipy:
-        raise ImportError("You must install `scipy` to use this function")
-
-
-def get_most_abundant_isotope(element: "pt.Element") -> "pt.Isotope":
-    """Given a `periodictable` element, return the most abundant
-    isotope.
-    """
-    most_abundant_isotope = element.isotopes[0]
-    abundance = 0
-    for iso in element:
-        if iso.abundance > abundance:
-            most_abundant_isotope = iso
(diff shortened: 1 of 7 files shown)


berquist:
CI is failing again because of https://github.com/cclib/cclib/pull/781#issuecomment-552129231

shivupa (approved):

berquist:
Looking over the coverage report, I think this actually needs, at a minimum, more testing and comments/docs about which set of atomic masses and physical constants is being used.

I also wonder if we *need* this.  Originally I think there were two goals:
1. Remove dependence on SciPy for physical constants
2. Switch to a library updated with seemingly more frequency, maintained by chemists, that is more likely to contain atomic constants of direct use in the future (such as nuclear quadrupole moments for spectroscopies)

Point 1 is moot since we depend on SciPy for other things now, like rotation matrices.  For point 2, `periodictable` currently works perfectly fine and has many more packages depending on it (https://libraries.io/pypi/periodictable/dependents; basically everything in https://libraries.io/pypi/qcelemental/dependents is in their own ecosystem).

shivupa:
> Looking over the coverage report, I think this actually needs, at a minimum, more testing and comments/docs about which set of atomic masses and physical constants is being used.
> 
> I also wonder if we _need_ this. Originally I think there were two goals:
> 
>     1. Remove dependence on SciPy for physical constants
> 
>     2. Switch to a library updated with seemingly more frequency, maintained by chemists, that is more likely to contain atomic constants of direct use in the future (such as nuclear quadrupole moments for spectroscopies)
> 
> 
> Point 1 is moot since we depend on SciPy for other things now, like rotation matrices. For point 2, `periodictable` currently works perfectly fine and has many more packages depending on it (https://libraries.io/pypi/periodictable/dependents; basically everything in https://libraries.io/pypi/qcelemental/dependents is in their own ecosystem).

I actually didn't merge because I wanted to check the consistency between the current constants and the `qcelemental` constants. `qcelemental` and `scipy.constants` both seem to use NIST CODATA 2018 values so I am fine with either (as you said scipy wouldn't add any more dependencies..) Sorry I should have mentioned why I didn't merge yet.

berquist:
No, I wasn't waiting, for some reason I was assuming that we started operating like my last job (submitter merges)...

I am working on a COM test.

langner:
I'm going to move this to v1.7.2 now.

langner:
Bumped to 1.8.1 for now, but is this something we still wanted to go through with?

berquist:
I'd rather switch this to be optional at best.  It turns out we can't depend on the continued maintenance of any MolSSI library, but it might be ok to have a bridge for generating QCElemental molecules, use their "QCSchema" writer, etc.

berquist:
I'm going to open a new PR with just the nuclear method updates and tests.

PR #992, opened 2020-12-24
Fix GAMESS parser when there are >100 scf iterations

simonaxelrod:
Hi there,
I noticed a bug when there are more than 100 scf iterations in GAMESS. Please let me know how my fix looks.
Thanks,
Simon

### DIFF
--- a/cclib/parser/gamessparser.py
+++ b/cclib/parser/gamessparser.py
@@ -693,13 +693,13 @@ class GAMESS(logfileparser.Logfile):
                     # will look like 10099, 101100, 102101, etc., with no spaces between
                     # the numbers. We can check if this is the case by seeing if the first
                     # number on the line exceeds 10000.
-                    gt_100_its = int(line.split()[0]) > 10000
+                    base_split = line.split()
+                    gt_100_its = int(base_split[0]) > 10000
                     if gt_100_its:
-                        base_split = line.split()
                         first_two = [base_split[0][:3], base_split[0][3:]]
                         split = [*first_two, *base_split[1:]]
                     else:
-                        split = line.split()
+                        split = base_split
                     values.append([float(split[self.scf_valcol])])
                 try:
                     line = next(inputfile)
--- a/cclib/parser/gamessparser.py
+++ b/cclib/parser/gamessparser.py
@@ -689,13 +689,25 @@ class GAMESS(logfileparser.Logfile):
                 except ValueError:
                     pass
                 else:
-                    values.append([float(line.split()[self.scf_valcol])])
+                    # if there were 100 iterations or more, the first part of the line
+                    # will look like 10099, 101100, 102101, etc., with no spaces between
+                    # the numbers. We can check if this is the case by seeing if the first
+                    # number on the line exceeds 10000.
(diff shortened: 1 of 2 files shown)


cclib/parser/gamessparser.py
```diff
@@ -689,14 +689,26 @@ def extract(self, inputfile, line):
                 except ValueError:
                     pass
                 else:
-                    values.append([float(line.split()[self.scf_valcol])])
+                    # if there were 100 iterations or more, the first part of the line
+                    # will look like 10099, 101100, 102101, etc., with no spaces between
+                    # the numbers. We can check if this is the case by seeing if the first
+                    # number on the line exceeds 10000.
+                    gt_100_its = int(line.split()[0]) > 10000
```
berquist, line 696: Can you store the result of `line.split()` so it can be used here and reused again on lines 698 and 702?

berquist (changes requested):
Hello Simon, thanks for the contribution! I don't see any obvious problems, though some of the naming is confusing. If you could submit an example logfile that has this problem to https://github.com/cclib/cclib-data/ via a PR, we can give some naming suggestions.

cclib/parser/gamessparser.py
simonaxelrod: Done!

simonaxelrod:
Made the changes, and also made a PR request for cclib-data here https://github.com/cclib/cclib-data/pull/147. Let me know how this looks

berquist (changes requested):
Having seen the output in https://github.com/cclib/cclib-data/pull/147, this makes more sense to me now. I was going to say that this could be avoided with using explicit indexing, but because of the need for `scf_valcol`, I can't see that solution being any shorter or easier to understand. So this looks good.

I've confirmed that the regression didn't parse before this patch and now it does. Could you do one more thing and add a test to https://github.com/cclib/cclib/blob/master/test/regression.py for the file you've added to cclib-data? Something like a check for how many SCF iterations it has is fine.

simonaxelrod:
Great, thank you! I've also added a regression test which checks the number of scf iterations

berquist (approved):
Excellent! The regression test looks good, I've tested that it works. I'm going to hold off merging until we decide on what to do about our free Travis CI credits that have run out...

cclib/parser/gamessparser.py
```diff
@@ -689,14 +689,26 @@ def extract(self, inputfile, line):
...
+                    # the numbers. We can check if this is the case by seeing if the first
+                    # number on the line exceeds 10000.
+                    base_split = line.split()
+                    gt_100_its = int(base_split[0]) > 10000
+                    if gt_100_its:
+                        first_two = [base_split[0][:3], base_split[0][3:]]
+                        split = [*first_two, *base_split[1:]]
+                    else:
+                        split = base_split
+                    values.append([float(split[self.scf_valcol])])
```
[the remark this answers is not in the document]
shivupa: THis would get rid of the if statement

shivupa:
Sorry we implemented this because it was holding up some CI testing things. Thanks for submitting the PR!

simonaxelrod:
No problem, thanks for finishing it up!

PR #995, opened 2021-01-05
Add zero-point vibrational energy to most parsers

berquist:
In at least one case entropy/pressure/temperature has been added too, since they appear before the ZPVE in the output.


shivupa (approved):

shivupa:

PR #1000, opened 2021-01-06
Update DOI in ANNOUNCE, and add Shiv and Geoffrey to Zenodo ref

langner:


shivupa:
2021 I think right?

shivupa (approved):

langner:
> 2021 I think right?

Well, technically we did the release commit last year, so I'll leave it at 2020.

berquist (approved):

berquist:
Number 1000!

PR #1001, opened 2021-01-11
Expanded Turbomole Functionality

oliver-s-lee:
For Turbomole:
 - Added support for some new attributes
 - Expanded support for some existing attributes
 - New test data


shivupa:
Hi! First sorry for the delay on this. We were working out a continuous integration migration which took a very long time. 

Seeing this many `@skipForParser('Turbomole', 'The parser is still being developed so we skip this test')` removed makes me very happy. At first glance this looks like a great contribution. 

However this is a huge PR however and we need to work out a plan to get this in. I am thinking the most reasonable approach is:
1. A PR containing only the job output files (no changes to the code)
2. A PR adding the basis parsing + any applicable associated tests
3. A PR adding the electrostatic moments parsing + any applicable associated tests
4. A PR adding the MO parsing + any applicable associated tests
5. Gradients + tests
6. Geom opt
7. MO parsing <- this adds an attribute so this might require discussion
8. MP 
9. Excited state (this is pretty big so if you can see ways to break this further feel free)
I think that is the bulk of it.

I apologize that this places additional work on you to break things up, but this way we can provide faster reviews to get things merged in. Does this sound reasonable?

oliver-s-lee:
Hi there,

Sure, your plan makes sense. I'll work on splitting the commits and will submit new PRs with smaller chunks.
I'm afraid I'm not totally familiar with the git process, does it make sense to leave this PR open or should I close it?

shivupa:
This one can be closed in favor of the smaller chunk PRs.

PR #1005, opened 2021-01-28
Parse SCF energy, Grad, Hess, Atom masses and transition (ETran) properties from fchk

jcerezochem:
Parses SCF energy, Grad, Hess, Atom masses and transition (ETran) properties.
Built upon the fchk extract interface by @ATenderholt,


cclib/parser/fchkparser.py
```diff
@@ -171,6 +171,113 @@ def extract(self, inputfile, line):
...
+                           -1:'Unknown'}
+            etsyms = []
+            for i in etspin:
+                if i in spin_labels:
+                    etsyms.append(spin_labels[i])
+                else:
+                    etsyms.append(spin_labels[-1])
+
+            # This must be completed with actual symm irrep, but "ETran symm" section seems
+            # not to be given nothing but zeroes regardles the symm irrep
```
shivupa, line 229: This comment is confusing. As I understand it: "After this 'Etran sym' appears (and would need to be parsed), but at least in Gaussian this contains only zeroes regardless of the irrep."

cclib/parser/fchkparser.py
```diff
@@ -171,6 +171,113 @@ def extract(self, inputfile, line):
...
+
+            self.set_attribute('etsyms', etsyms)
+
+        if line[0:18] == 'ETran state values':
+            # This section is organized as follows:
+            # ·First the properties of each excited state (up to self.net):
+            # E, {muNx,muNy,muNz,muvelNx,muvelNy,muvelNz,mmagNx,mmagNy,mmagNz,unkX,unkY,unkZ,unkX,unkY,unkZ}_N=1,self.net
+            # ·The comes 48 items (only if Freq is requested)
```
shivupa, line 237: Then*

shivupa (changes requested):

shivupa:
Thanks for this! I'll have to check how we are testing the FCHK parser and these attributes in other places. It would be nice to have a test to go with this.

cclib/parser/fchkparser.py
jcerezochem: Ok, agree. I wanted to note that the extracted values will not coincide with those from Gaussian log files, as they lack the actual irrep label.

jcerezochem:
I've modified the comments on the code. Regarding the test, I guess I should add a fchk file from TDDFT Freq calculation to:
cclib/data/Gaussian/basicGaussian16/

Right?

shivupa:
Hi I had a few more edits which I put on this branch: https://github.com/shivupa/cclib/tree/1005edits
I removed the new attribute. Once the current attributes have been parsed, the new attribute can be derived from the current attributes. (e.g. `etsyms` should be of size `net` right?) I'd be open to discussion of adding an additional attribute, but I'd prefer not to if possible.

Additionally it seems like the tests we are running for the FChk parser are being run from `cclib/data/FChk/basicQChem5.2`. If you'd like I could run a TD test in QChem and add it as a FChk test?

jcerezochem:
Thanks for the edits. I have made some further changes:

1. `net` has to be computed on each if-block it is being used if it is not an attribute (otherwise it gives a "used before assignment" error). It is not a problem, as you said it can be obtained from `len(self.etsyms)`, but we need to skip the `assert` test at "ETran spin" section
2. `netroot` is a useful attribute, since e.g. the gradient&Hessian (if present) correspond to such root. By the way, this attribute is also defined in gaussianparser.py

Regarding the test, I don't know if QChem stores exactly the same info and in the same way (labels), but this should be checked. If you can try, it'd be great.

shivupa:
> Thanks for the edits. I have made some further changes:
> 
>     1. `net` has to be computed on each if-block it is being used if it is not an attribute (otherwise it gives a "used before assignment" error). It is not a problem, as you said it can be obtained from `len(self.etsyms)`, but we need to skip the `assert` test at "ETran spin" section
Makes sense thanks.
>     2. `netroot` is a useful attribute, since e.g. the gradient&Hessian (if present) correspond to such root. By the way, this attribute is also defined in gaussianparser.py
OK this is good. I see it in the gaussian parser as well. Sorry for removing that.
> Regarding the test, I don't know if QChem stores exactly the same info and in the same way (labels), but this should be checked. If you can try, it'd be great.

I added a QChem test that didn't work out since QChem doesn't dump the same things to the output file. See [here](https://github.com/shivupa/cclib/tree/1005test). I replaced this with the gaussian test and opened a PR on your fork. Please check that out.

jcerezochem:
I've merged your changes and the parser works with the new test data (I did not execute the test, simply use your test fchk files). So if I understood correctly, no additional changes required for this PR, right?

I could also check that QChem is creating more understandable sections after TD: 
`Number of Excited States`
`Excitation Energies`
`Oscillator Strengths`
But I guess this should be handled in a separated PR. What QChem may be dumping similarly are Hessian and Gradient, but I did not find any example with such quantities.

shivupa:
Hi sorry for the delay, we had some larger changes coming into the test framework that just wrapped up. Can you rebase this on master? It should be ready to merge after the test runs. 
To rebase, 
add the upstream repo
`git remote add upstream https://github.com/cclib/cclib.git`
pull upstream master
` git rebase -i upstream/master`

jcerezochem:
Hi. No problem! (also sorry for my delay). Thanks for the detailed instructions. I rebased picking all commits, let me know if this was not ok, and they should be squashed instead.

shivupa (approved):

shivupa:

PR #1006, opened 2021-01-28
Add cpu/wall time to QChem metadata

amandadumi:
Extending the parser to parse the CPU/wall time from QChem output files. 

I needed to pull this out for my own research, though maybe it could be useful for others as well. Would parsing this information fall within cclib scope? Additionally,  if this is something you would parse, is metadata the place this should reside?

I could extend this to other parsers as well.

### DIFF
--- a/test/data/testSP.py
+++ b/test/data/testSP.py
@@ -347,6 +347,25 @@ class GenericSPTest(unittest.TestCase):
             packaging.version.parse(self.data.metadata["package_version"]),
             packaging.version.Version
         )
+    @skipForParser('ADF', 'reading cpu/wall time is not implemented for this parser')
+    @skipForParser('DALTON', 'reading cpu/wall time is not implemented for this parser') 
+    @skipForParser('FChk', 'reading cpu/wall time is not implemented for this parser') 
+    @skipForParser('GAMESS', 'reading cpu/wall time is not implemented for this parser') 
+    @skipForParser('GAMESSUK', 'reading cpu/wall time is not implemented for this parser') 
+    @skipForParser('GAMESSUS', 'reading cpu/wall time is not implemented for this parser') 
+    @skipForParser('Gaussian', 'reading cpu/wall time is not implemented for this parser') 
+    @skipForParser('Jaguar', 'reading cpu/wall time is not implemented for this parser') 
+    @skipForParser('Molcas', ' reading cpu/wall time is not implemented for this parser') 
+    @skipForParser('Molpro', 'reading cpu/wall time is not implemented for this parser') 
+    @skipForParser('NWChem', 'reading cpu/wall time is not implemented for this parser') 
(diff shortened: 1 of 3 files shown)


ghutchis:
From what I know, this could present problems with Gaussian, but it seems like a generally useful thing for many people (e.g., what's the trade-off between different methods).

amandadumi:
Oh yes, good point.  I was also presuming Gaussian would be off the table for this.

shivupa:
I like this and I think this would be good. We could leave any code that may have a problem with this unimplemented.
Three things:
Could you update this file to reflect the changes:
https://github.com/cclib/cclib/blob/d93d71fd51cf925dfc2a38d1d0ee5a1225b0cb11/doc/sphinx/data_notes.rst#metadata

Could you open a issue with a checklist for the codes that have yet to be implemented so subsequent PRs can be linked to that.

Could a test like [this](https://github.com/cclib/cclib/blob/bac01d2a38e236c2e717ae523442a0b0183282f5/test/data/testSP.py#L299) be added (and skipped for unimplemented parsers)?

amandadumi:
Thanks for the feedback. I've opened an issue to track this progress (#1007 ). I updated the documentation, and I've attempted to add a test. This isn't yet being recognized by pytest though.  I will keep working on that .

Though, I was curious what the best format is to store this data, it is currently a string. @shivupa brought up the idea of storing this as a [timedelta](https://stackoverflow.com/a/12352624). This has the benefit of being uniform between codes, but the drawback of having a slight learning curve for the user and more processing on our end.

amandadumi:
I went with the timedeltas as it will be easier to have a uniform usable datatype across various code outputs. 

I believe the test is correct based on the linked example from @shivupa above. Though when calling pytest locally using `python -m pytest -v --capture=no --cov=cclib --cov-report=term --cov-report=html --terse test`, none of the metadata tests are being checked according to the coverage report. Should I be running the tests differently?

test/data/testSP.py
```diff
@@ -347,6 +347,25 @@ def testmetadata_package_version(self):
             packaging.version.parse(self.data.metadata["package_version"]),
             packaging.version.Version
         )
+    
+    @skipForParser('ADF', 'reading input file contents and name is not implemented')
```
shivupa, line 351: Can you update the reason to all of these to something like 'reading cpu/wall times is not implemented for this parser.'

shivupa: Also can you add ORCA to this list. 

```
python -m pytest -v --capture=no --cov=cclib --cov-report=term --cov-report=xml:coverage_unit.xml test/test_data.py
```
If I run that locally with your branch the test is showing up and passing for QChem as expected.

amandadumi: These should both be addressed now. Thanks for checking on the tests, I'll trouble shoot this locally then.

shivupa (approved):
I like this. Also I think this sets a good precedence for metadata parsing. I don't think an error parsing metadata should cause an error.

PR #1008, opened 2021-01-31
Fix ASE bridge at version 3.21.1

langner:


langner:
Perhaps we don't want to merge this one, since I created a PR against Shiv's branch at shivupa/cclib#4

PR #1020, opened 2021-03-24
Turbomole: New log files for testing

oliver-s-lee:
New Turbomole log files for Turbomole v 7.5 or 7.4 for SP, Opt, MP2, MP3, CCSD, CC2 and ADC(2) with various excited states.

Additionally, the existing test data for Turbomole v5.9 dvb_sp_symm and dvb_un_sp has been modified to match the output from an SP calculation, rather than an opt as was the case, to prevent confusion with testers further down the line.


test/testdata
```diff
@@ -308,7 +308,7 @@ SPun      Psi4        GenericSPunTest       basicPsi4-1.3.1       dvb_sp_uhf.out
 SPun      Psi4        GenericSPunTest       basicPsi4-1.3.1       dvb_sp_uks.out
 SPun      QChem       GenericSPunTest       basicQChem4.2         dvb_sp_un.out
 SPun      QChem       GenericSPunTest       basicQChem5.1         dvb_sp_un.out
-SPun      Turbomole   GenericSPunTest       basicTurbomole5.9     dvb_un_sp/basis    dvb_un_sp/control    dvb_un_sp/job.last    dvb_un_sp/coord    dvb_un_sp/alpha    dvb_un_sp/beta    dvb_un_sp/gradient
+SPun      Turbomole   GenericSPunTest       basicTurbomole5.9     dvb_un_sp/basis    dvb_un_sp/control    dvb_un_sp/dscf.out    dvb_un_sp/coord    dvb_un_sp/alpha    dvb_un_sp/beta    dvb_un_sp/gradient
```
shivupa, line 311: Was this just a renaming of the corresponding file or did you have to rerun the job?

shivupa (approved):

test/testdata
oliver-s-lee: It's the same file but edited to remove the optimisation output, leaving just the output from dscf, and then renamed to a more likely name for a single point output file if that makes sense

shivupa:
Thank you merging!

PR #1022, opened 2021-04-01
Turbomole meta basis

oliver-s-lee:
Adds basis set (metadata) parsing for Turbomole


shivupa:
(Rebased on master)

shivupa (approved):

cclib/parser/turbomoleparser.py
```diff
@@ -123,6 +123,34 @@ def extract(self, inputfile, line):
...
+            line = next(inputfile)
+            line = next(inputfile)
+            basis_sets = []
+            while "---------------------------------------------------------------------------" not in line:
+                basis_sets.append(" ".join(line.split()[4:-1]))
+                line = next(inputfile)
+
+            # Turbomole gives us the basis set for each atom, but we're only interested if the same basis set is used throughout (for now).
+            if len(set(basis_sets)) == 1:
+                self.metadata["basis_set"] = list(set(basis_sets))[0]
```
berquist, line 150: Since there is no guarantee of set ordering, this will pick a random basis set name. Maybe if the size of the set it greater than one, log a warning? This should at least be made repeatable by using a sorted list.

cclib/parser/turbomoleparser.py
```diff
@@ -123,6 +123,34 @@ def extract(self, inputfile, line):
...
+        #    type   atoms  prim   cont   basis
+        #    ---------------------------------------------------------------------------
+        #     o        1     15      5   sto-3g hondo  [2s1p|6s3p]
+        #     h        2      3      1   sto-3g hondo  [1s|3s]
+        #    ---------------------------------------------------------------------------
+        if "type   atoms  prim   cont   basis" in line:
+            line = next(inputfile)
+            line = next(inputfile)
+            basis_sets = []
+            while "---------------------------------------------------------------------------" not in line:
```
berquist, line 144: Since the barrier is only dashes, you can do a thing we do in other places that looks like
```python
while set(line.strip()) != {"-"}:
```

test/data/testSP.py
```diff
@@ -447,6 +446,12 @@ class TurbomoleSPTest(GenericSPTest):
     """Customized restricted single point unittest"""
 
     num_scf_criteria = 2
+    
+    def testmetadata_basis_set(self):
+        """Does metadata have expected keys and values?"""
+        # One of our test cases used sto-3g hondo
+        valid_basis = self.data.metadata["basis_set"].lower() == "sto-3g" or self.data.metadata["basis_set"].lower() == "sto-3g hondo"
```
berquist, line 453:
```suggestion
        valid_basis = self.data.metadata["basis_set"].lower() in ("sto-3g", "sto-3g hondo")
```
though in the future, we should map `sto-3g hondo` to `sto-3g`, since AFAIK there is only one version of STO-3G.

cclib/parser/turbomoleparser.py
```diff
@@ -109,7 +109,7 @@ def extract(self, inputfile, line):
             nmo = int(line.split('=')[1])
             self.set_attribute('nbasis', nmo)
             self.set_attribute('nmo', nmo)
-
+
```
berquist, line 112: Stray space?

berquist (changes requested):
The only thing I think needs to be resolved is dealing with the case where multiple basis sets are present.

cclib/parser/turbomoleparser.py
shivupa: Sorry as I understand it if there is more than 1 basis set then won't this just not get set? 
To me this seems like the behavior we would want. If it couldn't be parsed correctly we just don't set it.

oliver-s-lee: Yes that is correct, at present the basis set will not be set if more than one is given.
Not sure what is done elsewhere, but I guess the alternative would be to set the most 'popular' basis set (the one used for the most atoms) instead?

cclib/parser/turbomoleparser.py
oliver-s-lee: oops

cclib/parser/turbomoleparser.py
oliver-s-lee: cool, thanks for the tip

cclib/parser/turbomoleparser.py
berquist: Oh sorry, my eyes completely glazed over the `if len(set(basis_sets)) == 1:` line. I like that solution (the current one) more than the "most popular" solution.

berquist (approved):

PR #1023, opened 2021-04-01
Turbomole meta misc

oliver-s-lee:
Miscellaneous metadata parsing for Turbomole, including:
- Success flag
- DFT/HF method
- DFT functional


shivupa:
(Rebased)

shivupa (approved):

cclib/parser/turbomoleparser.py
```diff
@@ -48,6 +48,9 @@ class Turbomole(logfileparser.Logfile):
 
     def __init__(self, *args, **kwargs):
         super(Turbomole, self).__init__(logname="Turbomole", *args, **kwargs)
+        
+        # Flag for whether this calc is DFT.
+        self.DFT = False
```
berquist, line 53: Some people hate this, but this is where I like a variant of Hungarian notation, or like how Lisp does it:
```suggestion
        self.is_DFT = False
```
What do you think?

cclib/parser/turbomoleparser.py
```diff
@@ -109,19 +112,60 @@ def extract(self, inputfile, line):
             nmo = int(line.split('=')[1])
             self.set_attribute('nbasis', nmo)
             self.set_attribute('nmo', nmo)
-
+            
+        # The DFT functional.
+        # This information is printed by dscf but not in an easily parsible format, so we'll take it from the control file instead...
+        # Additionally, turbomole stores functional names in lower case. This looks odd, so we'll convert to uppercase (?)
```
berquist, line 118: This is a fine first approximation, then you discover wB97 and related...but I think that's an edge case not worth worrying about now.

cclib/parser/turbomoleparser.py
```diff
@@ -109,19 +112,60 @@ def extract(self, inputfile, line):
...
+        # The DFT functional.
+        # This information is printed by dscf but not in an easily parsible format, so we'll take it from the control file instead...
+        # Additionally, turbomole stores functional names in lower case. This looks odd, so we'll convert to uppercase (?)
+        if line[3:13] == "functional":
+            self.metadata['functional'] = line.split()[1].upper()
+        
+        # The DFT functional.
+        # This information is printed by dscf but not in an easily parsable format, so we'll take it from the control file instead...
+        # Additionally, turbomole stores functional names in lower case. This looks odd, so we'll convert to uppercase (?)
+        if line[3:13] == "functional":
```
berquist, line 125: It looks like the above block is a duplicate of this one?

cclib/parser/turbomoleparser.py
```diff
@@ -109,19 +112,60 @@ def extract(self, inputfile, line):
...
+        
         # Extract the version number and optionally the build number.
         searchstr = ": TURBOMOLE"
         index = line.find(searchstr)
         if index > -1:
             line = line[index + len(searchstr):]
             tokens = line.split()
-            package_version = tokens[0][1:].replace("-", ".")
+            # This line could look like any of the following:
+            # - TURBOMOLE rev. V7.4.1 (2bfdd732)
```
berquist, line 141: It might be cleaner at this point to extract the `VX.Y.Z` and the revision using regexes.

cclib/parser/turbomoleparser.py
```diff
@@ -109,19 +112,60 @@ def extract(self, inputfile, line):
...
+            # - TURBOMOLE V7.2 ( 21471 ) 11 Oct 2017 at 17:04:51
+            # - TURBOMOLE V5-9-0 29 Nov 2006 at 22:06:41
+            
+            # Determine where the version string is.
+            if tokens[0] == "rev.":
+                version_index = 1
```
berquist, line 147: Do you know of a case where this is tested?

berquist (changes requested):
The most important thing is that you've accidentally duplicated the check for the density functional.

cclib/parser/turbomoleparser.py
oliver-s-lee: I'm not against this personally. I based this naming of the 'convention' (if that's the right term) used in the Gaussian Parser, which is not to include the 'is_' prefix. eg (coupledcuster, oniom, bomd etc):
```
def before_parsing(self):

        # Extract only well-formed numbers in scientific notation.
        self.re_scinot = re.compile(r'(\w*)=\s*(-?\d\.\d{2}D[+-]\d{2})')
        # Extract only well-formed numbers in traditional
        # floating-point format.
        self.re_float = re.compile(r'(\w*-?\w*)=\s*(-?\d+\.\d{10,})')

        # Flag for identifying Coupled Cluster runs.
        self.coupledcluster = False

        # Fragment number for counterpoise or fragment guess calculations
        # (normally zero).
        self.counterpoise = 0

        # Flag for identifying ONIOM calculations.
        self.oniom = False

        # Flag for identifying BOMD calculations.
        # These calculations have a back-integration algorithm so that not all
        # geometries should be kept.
        # We also add a "time" attribute to the parser.
        self.BOMD = False

        # Do we have high-precision polarizabilities printed from a
        # dedicated `polar` job? If so, avoid duplicate parsing.
        self.hp_polarizabilities = False
```
Happy to change tho?

cclib/parser/turbomoleparser.py
oliver-s-lee: Done, regex is not my forte I'm afraid so there may be a better solution...

cclib/parser/turbomoleparser.py
oliver-s-lee: Sadly not, we're missing the correct test data for the version of Turbomole that prints the version info like this...

oliver-s-lee: Added

cclib/parser/turbomoleparser.py
berquist: Yes, let's switch. The Gaussian parser is one of the crustier ones.

berquist (approved):

PR #1025, opened 2021-04-01
Turbomole mos

oliver-s-lee:
Adds support for parsing homos, mosyms and moenergies

Please note this PR also adds the custom Turbomole attribute moirreps which is required for parsing other data (eg, excited states)

### DIFF
--- a/cclib/parser/turbomoleparser.py
+++ b/cclib/parser/turbomoleparser.py
@@ -327,17 +327,24 @@ class Turbomole(logfileparser.Logfile):
             while line.strip().startswith('#') and not line.find('eigenvalue') > 0:
                 line = next(inputfile)
 
+            moirreps = []
             moenergies = []
             mocoeffs = []
             mosyms = []
 
             while not line.strip().startswith('$'):
                 number, sym = line.split()[:2]
+                number = int(number)
+                sym = self.normalisesym(sym)
+                
                 info = re.match(r".*eigenvalue=(?P<moenergy>[0-9D\.+-]{20})\s+nsaos=(?P<count>\d+).*", line)
                 eigenvalue = utils.float(info.group('moenergy'))
                 orbital_energy = utils.convertor(eigenvalue, 'hartree', 'eV')
+                
                 moenergies.append(orbital_energy)
-                mosyms.append(self.normalisesym(sym))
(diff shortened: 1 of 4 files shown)


shivupa:
(rebased)

cclib/parser/turbomoleparser.py
```diff
@@ -108,8 +108,8 @@ def extract(self, inputfile, line):
         if line[3:11] == "nbf(AO)=":
             nmo = int(line.split('=')[1])
             self.set_attribute('nbasis', nmo)
-            self.set_attribute('nmo', nmo)
-
+            #self.set_attribute('nmo', nmo)
```
shivupa, line 111: Should this be deleted?

shivupa (changes requested):

cclib/parser/turbomoleparser.py
oliver-s-lee: Yes it is no longer required; there is an alternative method now for setting nmo

cclib/parser/turbomoleparser.py
```diff
@@ -122,7 +121,71 @@ def extract(self, inputfile, line):
...
+                self.set_attribute("homos", [self.determine_homo(self.mosyms[0], orbitals)])
+            
+            if "alpha:" in line:
+                orbitals, line = self.parse_dscf_orbitals(inputfile, line)
+                if len(orbitals) > 0:
+                    homo = self.determine_homo(self.mosyms[0], orbitals)
+                    if not hasattr(self, "homos"):
+                        self.set_attribute('homos', [homo])
+                    else:
+                        self.homos[0] = homo
```
berquist, line 175: Strange, do you know why this isn't covered?

berquist (approved):
Looks good, just one question.

cclib/parser/turbomoleparser.py
oliver-s-lee: The logic for this branch is only triggered if we've already parsed HOMOs before (from a previous opt step), but all our test data for unrestricted HF is single point only.

berquist: That makes sense. In that case, I'm ok with this.

berquist (approved):

PR #1027, opened 2021-04-01
Turbomole mp

oliver-s-lee:
Adds support for parsing Møller–Plesset energies for Turbomole.


berquist (changes requested):
All looks good but a question about the removal of the coupled cluster part.

oliver-s-lee:
The removed CC part is an artefact of how this PR was split from the original big PR I think. In the re-written version MP and CC are parsed separately, while currently in master they're parsed together. When the PR was split I simply segregated the MP and CC sections, forgetting the old CC parsing method would be (temporarily) lost. If it's a problem I can re-add the old CC parsing section?

The rewritten CC bit is in #1028, which incidentally depends on this PR (it also contains the new MP bit, can't remember why I did it like that sorry)...

berquist (approved):
I think that's ok, since it's temporary; it answers my question.

PR #1029, opened 2021-04-01
Turbomole es td

oliver-s-lee:
Adds support for parsing TD-DFT excited states for Turbomole

### DIFF
--- a/cclib/parser/turbomoleparser.py
+++ b/cclib/parser/turbomoleparser.py
@@ -972,12 +972,14 @@ class Turbomole(logfileparser.Logfile):
             while "Electric transition dipole moment (length rep.):" not in line:
                 line = next(inputfile)
             line = next(inputfile)
+            
+            # Unlike PDM, we leave TDMs in units of au (based on the implementation for Gaussian). 
             line = next(inputfile)
-            tdm_x = utils.convertor(float(line.split()[1]), "ebohr", "Debye")
+            tdm_x = float(line.split()[1])
             line = next(inputfile)
-            tdm_y = utils.convertor(float(line.split()[1]), "ebohr", "Debye")
+            tdm_y = float(line.split()[1])
             line = next(inputfile)
-            tdm_z = utils.convertor(float(line.split()[1]), "ebohr", "Debye")
+            tdm_z = float(line.split()[1])
             
             self.append_attribute("etdips", [tdm_x, tdm_y, tdm_z])
             
--- a/cclib/parser/turbomoleparser.py
+++ b/cclib/parser/turbomoleparser.py
@@ -4,6 +4,7 @@
 #
 # This file is part of cclib (http://cclib.github.io) and is distributed under
 # the terms of the BSD 3-Clause License.
+import collections
 
 """Parser for Turbomole output files."""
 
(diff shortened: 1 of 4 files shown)


cclib/parser/turbomoleparser.py
```diff
@@ -819,12 +823,160 @@ def extract(self, inputfile, line):
...
+            # The irrep of the state is a few lines back.
+            symm_parts = self.last_lines[-5].split()
+            # We don't always have the multiplicity available.
+            if len(symm_parts) < 4:
+                # No mult.
+                mult = "???"
+            else:
+                mult = symm_parts[1].capitalize()
+            
+            symmetry = "{}-{}".format(mult, symm_parts[-2].capitalize())
```
berquist, line 882: Is http://cclib.github.io/data_notes.html#etsyms a free-for-all? I'm assuming yes and that this is within requirements.

cclib/parser/turbomoleparser.py
```diff
@@ -819,12 +823,160 @@ def extract(self, inputfile, line):
...
+                        end_AB = 0
+                        
+                    # And end orbital.
+                    end_MO_irrep = (int(parts[3]), self.normalisesym(parts[4]))
+                    end_MO = self.moirreps[end_AB].index(end_MO_irrep)
+                    
+                    # Finally, get our coefficient.
+                    # Sadly, Turbomole only prints |coeff.|^2*100, so we can't determine whether
+                    # the coefficient is negative or positive...
+                    coeff = (utils.float(parts[-1]) /100) ** 0.5
```
berquist, line 943: I'm not sure this is a good idea, because then the norm over all coefficients will be wrong. It might be better to use `np.nan` instead. Is it still useful as-is?  A logger warning should be printed if this is kept.

berquist (approved):

berquist (changes requested):
Sorry, didn't mean to accept right away.

cclib/parser/turbomoleparser.py
oliver-s-lee: I believe so yes, but I'm no expert. The format used here matches that of the Gaussian parser...

cclib/parser/turbomoleparser.py
oliver-s-lee: That's a good point which I hadn't considered :/

Yes it is still useful like this, because the square of the coefficient gives the probability which is (for me personally) more useful than the coefficient anyway. Obviously the probability is independent of the sign of the coefficient, and as far as I know there's no specific cclib attribute for transition probability, so hence the implementation here.

It seems strange to me that Turbomole wouldn't provide the coefficients somewhere, but I can't see it. If there is some other way in which the sign of the coefficient could be determined I guess that would be best, but I don't know how that could be done...

berquist: I can't find anything in the manual about setting the print level for the TDDFT program.

I am ok with this as long as you add a `logger.warning` call just before setting `etsecs` that the signs of the coefficients are not available, but the probabilities are correct.

oliver-s-lee: I've added the warning, might want to tweak the wording?

berquist (approved):
The wording for the warning is fine. Once you rebase, I'll merge.

berquist:
Sorry for the wait, I completely missed this.

PR #1030, opened 2021-04-01
Turbomole es ricc2

oliver-s-lee:
Adds support for parsing excited states with Turbomole's ricc2 module (methods CC2 and ADC(2))


berquist:
@oliver-s-lee Poke for rebase.

cclib/parser/turbomoleparser.py
```diff
@@ -983,6 +983,150 @@ def extract(self, inputfile, line):
...
+        #  | a   |   1   |   3   |    0.3864541 |   10.51595 |  84816.880 |  95.05 |   4.95 |
+        #  | a   |   1   |   4   |    0.3938325 |   10.71673 |  86436.241 |  94.09 |   5.91 |
+        #  | a   |   1   |   5   |    0.3938325 |   10.71673 |  86436.241 |  94.09 |   5.91 |
+        #  | a   |   1   |   6   |    0.4122171 |   11.21700 |  90471.202 |  93.32 |   6.68 |
+        #  | a   |   1   |   7   |    0.4351537 |   11.84114 |  95505.195 |  94.16 |   5.84 |
+        #  | a   |   1   |   8   |    0.4454969 |   12.12259 |  97775.278 |  94.26 |   5.74 |
+        #  | a   |   1   |   9   |    0.4454969 |   12.12259 |  97775.273 |  94.26 |   5.74 |
+        #  | a   |   1   |  10   |    0.5036295 |   13.70446 | 110533.909 |  93.51 |   6.49 |
+        #  +================================================================================+
+        # TODO: Perhaps need a more general way to look for lines like this?
```
berquist, line 1003: The only thing I can think of right now would be a regex with `(?:(CC2|ADC\(2\))) excitations energies\s+|` (untested :) ) but I don't think it's worth making that change yet.

cclib/parser/turbomoleparser.py
```diff
@@ -983,6 +983,150 @@ def extract(self, inputfile, line):
...
+        if "| sym | multi | state |          CC2 excitation energies       |  %t1   |  %t2   |" in line \
+        or "| sym | multi | state |          ADC(2) excitation energies    |  %t1   |  %t2   |" in line:
+            line = next(inputfile)
```
berquist, line 1006: We don't have a written suggestion for this yet, but can you replace these with a `self.skip_lines` call, then a single `line = next(inputfile)`? I've written https://github.com/cclib/cclib/issues/555#issuecomment-882054020 so that we'll add it to docs. You can grep for some examples in the `cclib/parser` folder.

cclib/parser/turbomoleparser.py
```diff
@@ -983,6 +983,150 @@ def extract(self, inputfile, line):
...
+                    mult = "Quartet"
+                elif parts[2] == "-":
+                    mult = "???"
+                else:
+                    mult = parts[2]
+                    
+                symmetry = "{}-{}".format(mult, parts[1].capitalize())
+                self.append_attribute("etsyms", symmetry)
+                    
+                #energy = utils.convertor(utils.float(parts[4]), "hartree", "wavenumber")
```
berquist, line 1036: Stray comment?

cclib/parser/turbomoleparser.py
```diff
@@ -983,6 +983,150 @@ def extract(self, inputfile, line):
...
+        #      |    7 a        7 (a)   |   12 a       12 (a)   |  -0.49335      24.3   |
+        #      |    7 a        7 (a)   |    9 a        9 (a)   |   0.42257      17.9   |
+        #      |    7 a        7 (b)   |    9 a        9 (b)   |  -0.42257      17.9   |
+        #      |    7 a        7 (b)   |   13 a       13 (b)   |   0.20077       4.0   |
+        #      |    7 a        7 (a)   |   13 a       13 (a)   |  -0.20077       4.0   |
+        #      |    7 a        7 (a)   |   15 a       15 (a)   |   0.11607       1.3   |
+        #      |    7 a        7 (b)   |   15 a       15 (b)   |  -0.11607       1.3   |
+        #      +=======================+=======================+=======================+
+        if "| occ. orb.  index spin | vir. orb.  index spin |  coeff/|amp|     %    |" in line:
+            line = next(inputfile)
```
berquist, line 1073: Same comment here about `skip_lines`.

berquist (changes requested):

cclib/parser/turbomoleparser.py
oliver-s-lee: Like this?

berquist: Yes, exactly.

berquist (approved):

cclib/parser/turbomoleparser.py
oliver-s-lee: Cool got it.

Just out of interest, would the following not also work (replacing both lines)?:
`line = self.skip_lines(inputfile, ["divider", "units", "divider", "value"])[-1]`
Or is this style not prefered?

berquist: Huh, I guess that could actually work. This might not be a bad idea, there's just no precedent for it. I didn't even remember that `skip_lines` returns the skipped lines. What does everyone else think? @langner @shivupa

shivupa: Yup that does seem like it should work as well. We don't use `skip_lines` like that but I don't see a problem with it. It somewhat feels confusing cause it reads like we are skipping the last line, but then we end up using it. 

> I didn't even remember that skip_lines returns the skipped lines. 

Me neither. I'm curious if the return value is used anywhere.

PR #1031, opened 2021-04-01
Turbomole misc

oliver-s-lee:
Miscellaneous updates for Turbomole.

This is the final PR from this set (for now).

PR #1028, opened 2021-04-01
Turbomole cc

oliver-s-lee:
Adds support for parsing coupled cluster energies for Turbomole

PR #1026, opened 2021-04-01
Turbomole opt

oliver-s-lee:
Adds support for parsing optimisation related data for Turbomole

PR #1037, opened 2021-04-15
Turbomole scf fix

oliver-s-lee:
'Minimal' Turbomole calculations (eg, to generate cubes from a previous calculation) do not calculate SCF energies which break the Turbomple parser. This fixes that.


berquist:
I think a lot the contents of this PR for the actual parser are changes that were merged in already via the smaller PRs. I don't feel comfortable right now doing the merge/rebase against current master myself. Can you take a look?

berquist:
Ah, I also see that this contains changes from #1027, #1031, #1030, #1029, #1028, and #1026. I think I understand how you made these branches now. They need to be merged in a specific order. Should it be numerical order?

shivupa (approved):

berquist (approved):

PR #1039, opened 2021-04-20
Elapsed and CPU time in gaussian

elliotfarrar:
I added a few lines of code to extract the total elapsed and total cpu job times to the metadata attribute for Gaussian jobs. I couldn't see any existing code or issue for this.

If there is no time in a file, a default value of 0 is given for `metadata["elapsedtime"]` and/or `metadata["cputime"]`, but this could also be changed to n/a or something else.


cclib/parser/gaussianparser.py
```diff
@@ -2099,5 +2103,15 @@ def natural_orbital_single_spin_parsing(inputfile, updateprogress_title):
                     self.optdone = []
                 self.optdone.append(len(self.optstatus) - 1)
 
+        # Extract total elapsed and CPU job times in seconds
+        if line[:14] == ' Elapsed time:':
+            days, hours, minutes, seconds = float(line.split()[2]), float(line.split()[4]), float(line.split()[6]), float(line.split()[8])
```
berquist, line 2108: Can you change this to only `line.split()` once into a separate variable, then use that?

cclib/parser/gaussianparser.py
```diff
@@ -2099,5 +2103,15 @@ def natural_orbital_single_spin_parsing(inputfile, updateprogress_title):
...
+            elapsed = seconds + 60*minutes + 60*60*hours + 60*60*24*days
+            self.metadata["elapsedtime"] += elapsed
+        if line[:14] == ' Job cpu time:':
+            days, hours, minutes, seconds = float(line.split()[3]), float(line.split()[5]), float(line.split()[7]), float(line.split()[9])
```
berquist, line 2112: Same comment here.

berquist (changes requested):
Can you remove the empty Turbomole files? Weird...

fdroessler:
@berquist the problem actually seems to be this commit: https://github.com/cclib/cclib/commit/1956e646f6e366dc608a598a56be0287a5c8f8cb
here someone introduced files that are named the same but have different casing:
CCfE0-1--1---5
     *
CCFE0-1--1---5

This seems to break the repo for anyone using an OS that does not support filenames that only differ in case. In my case and I think @elliotfarrar case that is OSX. It's a pretty nasty thing ... I can make a PR to fix it and then we can rebase this one before merging it.

elliotfarrar:
Thanks for solving the problem with the turbomole files @fdroessler.

@berquist I've tidied up the code a bit and added tests for the two properties for one file from each Gaussian version.

berquist (changes requested):
Sorry for not giving a proper review before. The code and tests look good as-is, but looking at #1006:
1. `metadata` is a bit of a free for all, but we do need to standardize on key names. `elapsedtime` is the same as `wall_time`, right?
2. You mention picking something other than the default of zero, and I think a better approach is not have the key present at all if the field isn't there. It puts more responsibility on the user (they will need to check if the key is present before try to use the value or use zero), but makes it explicit what could or couldn't be parsed.
3. It's just a convenience thing, but I also like the idea of parsing with `datetime.timedelta`. (In the future, for programs that print starting times, we could also parse that information into an absolute time, which might be good for provenance.)
4. There is a [new test for times in `metadata`](https://github.com/cclib/cclib/pull/1006/files#diff-23c7024ea043d6e9decf5080e242971581660ff56a05ebc9eadaa534cade84b3R365) (just checking if keys are present) that you should enable for Gaussian.

elliotfarrar:
Thanks for taking a look @berquist. I've made some changes in response to your comments.

> 1. `metadata` is a bit of a free for all, but we do need to standardize on key names. `elapsedtime` is the same as `wall_time`, right?

I used the names given by Gaussian but didn't realise there was already some protocol for this. I've updated the names to wall_time and cpu_time, as they are for QChem.

> 2. You mention picking something other than the default of zero, and I think a better approach is not have the key present at all if the field isn't there. It puts more responsibility on the user (they will need to check if the key is present before try to use the value or use zero), but makes it explicit what could or couldn't be parsed.

If there is no time available in the output file we will now get empty lists in metadata, like: ` 'wall_time': [], 'cpu_time': []`, which I believe is the same as the QChem parsing from #1006. Is this sufficient or better for the properties to be left out of metadata entirely if they are not there (would that clash with the check[ here](https://github.com/cclib/cclib/blob/90acaa1c68aa5addd2839812a3b02d0720321746/test/data/testSP.py#L364))?

> 3. It's just a convenience thing, but I also like the idea of parsing with `datetime.timedelta`. (In the future, for programs that print starting times, we could also parse that information into an absolute time, which might be good for provenance.)

I've updated to parse with datetime.timedelta and in general matched the code as closely as possible to #1006 for consistency.

> 4. There is a [new test for times in `metadata`](https://github.com/cclib/cclib/pull/1006/files#diff-23c7024ea043d6e9decf5080e242971581660ff56a05ebc9eadaa534cade84b3R365) (just checking if keys are present) that you should enable for Gaussian.

I've deleted the line skipping Gaussian for this.

I also previously added regression tests that checked for specific values for cpu_time and wall_time and have updated these to check against the new datetime.timedelta format.

berquist:
> better for the properties to be left out of metadata entirely if they are not there (would that clash with the check here)?

It does clash, since the test only checks for the key. A better test is to also check that the first element of each list is also a `datetime.timedelta` instance.

elliotfarrar:
I've changed it so the key (cpu_time or wall_time) is now only defined if the field exists in the first place. This is done by moving the `self.metadata['wall_time'] =[]` line from `before_parsing` into where the extraction takes place, but making sure that the metadata key is not overwritten if it already exists. There might be a better way to do it.

The test in `testSP.py` now checks that the first element of metadata["key"] is a datetime.timedelta instance, if that key exists in metadata (rather than only checking that metadata["key"] exists for both keys). I also changed the QChem code to work the same way so that this didn't conflict with the new test.

berquist:
I've created https://github.com/the-grayson-group/cclib/pull/11 to fix the failing unit tests. Can you take a look?

elliotfarrar:
> I've created [the-grayson-group#11](https://github.com/the-grayson-group/cclib/pull/11) to fix the failing unit tests. Can you take a look?

This looks good. I've fixed the regression tests I added and tests are passing now.

test/data/testSP.py
```diff
@@ -359,10 +359,15 @@ def testmetadata_package_version(self):
     @skipForParser('Psi4', 'reading cpu/wall time is not implemented for this parser') 
     @skipForParser('Turbomole', 'reading cpu/wall time is not implemented for this parser') 
     def testmetadata_times(self):
-        """Does metadata have expected keys and values?"""
-        self.assertIn("cpu_time", self.data.metadata)
-        self.assertIn("wall_time", self.data.metadata)
-
+        """Does metadata have expected keys and values of correct types?"""
+        if "wall_time" in self.data.metadata:
+            assert self.data.metadata["wall_time"]
```
shivupa, line 364: This test doesn't seem like it will ensure that `wall_time` is in `metadata`. Shouldn't this (or a similar line to what was removed) appear outside of the if statement?

shivupa: Same for `cpu_time`

elliotfarrar: This was changed to cover files where either or both of the properties can't be parsed. In which case `wall_time` or `cpu_time` does not appear in metadata at all, rather than being present but as an empty list for example.

shivupa: Thanks that makes sense

shivupa (approved):

berquist (approved):
Great, thanks for all your hard work!

PR #1041, opened 2021-05-11
Remove duplicate files that cause name collisions on certain OS

fdroessler:
In https://github.com/cclib/cclib/commit/1956e646f6e366dc608a598a56be0287a5c8f8cb a couple of duplicate files were introduced into the data directory. They are named the same but differ in casing.

 data/Turbomole/basicTurbomole7.4/CO_cc2_TD/CCfE0-1--1---1
 data/Turbomole/basicTurbomole7.4/CO_cc2_TD/CCFE0-1--1---1

Notice the F and f in the filename

This is ok for some operating systems such as linux but not for others such as OSX. This has caused issues and confusion for other PRs e.g. #1039 

As fas as I can tell, the files that I am removing here are not used anywhere anyway and are most likely a result of an accident. Otherwise maybe @oliver-s-lee can comment on this too in case these file are used anywhere I missed.


berquist (approved):
Thank you for your detective work!

PR #1042, opened 2021-05-17
Parse intrinsic reaction coordinate calculations from Q-Chem

### LINKED ISSUE #1040: Parse intrinsic reaction coordinate calculations from Q-Chem
Similar to #842, but will require some filtering of `atomcoords` because, like #283, there are some geometry steps printed which are not actually a part of the final IRC.

berquist:
Closes #1040 

Some things that need to be resolved:
- `ircstatus` being separate from `optstatus` is more flexible in the event that a multi-step job is run, but the `{converged,new,unknown,unconverged}_geometries` methods only work with `optstatus`, and their docstrings imply that they could be used for IRC calculations. The best solution is only use `optstatus` and split multi-part jobs into multiple `ccData` instances.
- `ircvalues` is a grabbag. Most of the stuff probably isn't important.
- The fix for printing an extra set of `atomcoords` (the input ones appear again at the very end) fails for when a job appears after the IRC one.


cclib/parser/qchemparser.py
```diff
@@ -1457,6 +1472,77 @@ def extract(self, inputfile, line):
...
+            if "Reaction path following" in line:
+                # When this section appears, the immediately preceding
+                # energy/geometry corresponds to the converged step along the
+                # reaction path, and the other steps are corrector/predictor
+                # steps.
+                tokens = next(inputfile).split()
+                assert len(tokens) == 10
+                assert self.rpath_direction is not None
+                coord_idx = len(self.atomcoords) - 1
+                ircvalues_step = {
```
shivupa, line 1502: is there some guarantee of what fields will be populated in this dict for different IRC calculations? It seems like we are creating another metadata situation where the fields of this are not well defined. This might be necessary though

berquist: Right now, I think the only thing that's important is some indication of which direction we're going (up or down the TS eigenmode). Other programs might only allow you to go in one direction, but I need a way of differentiating between the two since Q-Chem tries both.

langner:
Is this ready for v1.7.1?

berquist:
I think all that remains is to make sure the tests are good and add some docs.

berquist:
I'm a bit stuck conceptually for how this implementation is supposed to work.

For an IRC calculation, there are two levels of convergence:
- Convergence of the whole reaction path down to the product or reactant state
- Convergence of a single step along the reaction path

I *think* we want to set it so that `OPT_DONE` corresponds to each step along the reaction path.  Is this how the other parsers use `optstatus`?

ATenderholt:
I think Gaussian was the initial implementation, so I've been refreshing my memory from there. Also looks like it's changed a bit since I last worked on it.

It does look like `optstatus` is set for every coordinate to indicate its state in a given optimization step (`OPT_NEW`), uncoverged (`OPT_UNCONVERGED`), or converged (`OPT_DONE`).

```
>>> from cclib.parser import ccopen
>>> parser = ccopen("data/Gaussian/basicGaussian16/dvb_scan_relaxed.log")
>>> data = parser.parse()
>>> len(data.optstatus)
61
>>> len(data.atomcoords)
61
>>> len(data.scfenergies)
61
>>> data.optdone
True
>>> data.optstatus
array([5, 1, 0, 0, 0, 4, 1, 0, 0, 0, 4, 1, 0, 0, 0, 0, 4, 1, 0, 0, 0, 0,
       0, 0, 4, 1, 0, 0, 0, 4, 1, 0, 0, 0, 4, 1, 0, 0, 4, 1, 0, 0, 4, 1,
       0, 0, 0, 4, 1, 0, 0, 0, 4, 1, 0, 0, 4, 1, 0, 0, 4], dtype=int32)

```

I'm a bit surprised the first value is 5 (`OPT_DONE` & `OPT_NEW`), but maybe it just started from an optimized structure.

PR #1046, opened 2021-06-13
QChem: keep the last vibrational analysis section, not the first

### LINKED ISSUE #871: Parsing frequency jobs in linked calculations
When parsing IR and Raman spectra from Q-Chem output files the current cclib combines all jobs from the output and create a "combined" spectrum. If one makes a TS job, it is often that calculation starts and finishes with frequency jobs.
I will be absolutely great, if there will be an opportunity to check the output file for frequency jobs and parse from only the last one.

berquist:
Closes #871


berquist:
https://github.com/cclib/cclib-data/pull/153 should go in before this PR.

shivupa (approved):

PR #1050, opened 2021-07-14
CI custom docker image

shivupa:
Hopefully running test from a custom image will result in a faster CI.


shivupa:
I think something is wrong with the pytest install in the image, but this will save roughly 4 minutes in the CI (mainly all from skipping a conda install).

shivupa:
Update to this. 

Good News: I have now automated the generation of images from the git repo https://github.com/shivupa/cclib-ci/actions

Bad News: The generated images are not working

shivupa:
The images build on push to master not on a schedule as I haven't worked out what exactly is wrong with the image.

berquist:
It's good that this is working, but the size of the final image is ~5.5 GB.  This means that downloading the image takes almost two and a half minutes, which is too long.  I am trying to shrink the image size further by using the distroless trick from https://uwekorn.com/2021/03/01/deploying-conda-environments-in-docker-how-to-do-it-right.html.

berquist:
Maybe this is ready now and deal with image bloat later. The two things that bother me are:
- We can't test on 3.8 and 3.9 because bridge dependencies break, though we discussed picking a Docker image based on which files are changed (having one with bridge dependencies and one without)
- It's easy to break our CI by committing a broken image to https://github.com/shivupa/cclib-ci.

shivupa:
This seems ready to go

PR #1054, opened 2021-07-22
Parse rotational constants and molecular mass for Gaussian logs

amarkpayne:
# Purpose
Currently, cclib has the ability to parse rotational constants and molecular mass from MOPAC log files (I think this functionality was added to cclib long ago by RMG developer Greg Magoon to support his work on QMTP in RMG). This PR extends this ability to Gaussian log files, which is something we use cclib for in RMG. 

# Testing
I have tested locally that the added code works by testing it on Gaussian log files for various semi-empirical calculations. It is the same suite of tests that we use to test the MOPAC parsing ability already implemented in cclib

# Other considerations
This is my first time contributing to cclib, so let me know if I missed any steps in your usual development workflow. I tried to follow the guidelines outlined in your online documentation.

### DIFF
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -2271,5 +2271,12 @@ class Gaussian(logfileparser.Logfile):
             else:
                 self.rotcons.append([utils.float(splits[i])/29.9792458 for i in (-3, -2, -1)])
 
+        # Extract Molecular Mass (in amu)
+        # Example:
+        # Molecular mass:   128.06260 amu.
+        if line[:16] == ' Molecular mass:':
+            splits = line.split()
+            self.molmass = utils.float(splits[2])
+
         if line[:31] == ' Normal termination of Gaussian':
             self.metadata['success'] = True
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -2264,12 +2264,7 @@ class Gaussian(logfileparser.Logfile):
         # Note: rotational constant will be converted to wavenumber units (1/cm) to standardize across parsers
         if line[:28] == ' Rotational constants (GHZ):':
             splits = line[28:].split()
-
-            # Determine if the molecule is linear and only has two constants
-            if '*' in line:  # linear molecule
-                self.append_attribute("rotconsts", [0.0]+[float(splits[i]) for i in (-2, -1)])
-            else:
-                self.append_attribute("rotconsts", [float(splits[i]) for i in (-3, -2, -1)])
+            self.append_attribute("rotconsts", [float(splits[i]) for i in (-3, -2, -1)])
 
         # Extract Molecular Mass (in amu)
         # Example:
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -2261,15 +2261,15 @@ class Gaussian(logfileparser.Logfile):
         # Rotational constants (GHZ): ************ 12.73690 12.73690
         # Note: rotational constant will be converted to wavenumber units (1/cm) to standardize across parsers
         if line[:27] == ' Rotational constants (GHZ)':
-            if not hasattr(self, "rotcons"):
-                self.rotcons = []
+            if not hasattr(self, "rotconsts"):
+                self.rotconsts = []
             splits = line.split()
 
             # Determine if the molecule is linear and only has two constants
             if '*' in line:  # linear molecule
-                self.rotcons.append([0.0]+[float(splits[i]) for i in (-2, -1)])
+                self.rotconsts.append([0.0]+[float(splits[i]) for i in (-2, -1)])
             else:
-                self.rotcons.append([float(splits[i]) for i in (-3, -2, -1)])
+                self.rotconsts.append([float(splits[i]) for i in (-3, -2, -1)])
 
         # Extract Molecular Mass (in amu)
         # Example:
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -2262,8 +2262,8 @@ class Gaussian(logfileparser.Logfile):
         # OR for linear molecules:
         # Rotational constants (GHZ): ************ 12.73690 12.73690
         # Note: rotational constant will be converted to wavenumber units (1/cm) to standardize across parsers
-        if line[:27] == ' Rotational constants (GHZ)':
-            splits = line.split()
+        if line[:28] == ' Rotational constants (GHZ):':
+            splits = line[28:].split()
 
             # Determine if the molecule is linear and only has two constants
             if '*' in line:  # linear molecule
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -2267,9 +2267,9 @@ class Gaussian(logfileparser.Logfile):
 
             # Determine if the molecule is linear and only has two constants
             if '*' in line:  # linear molecule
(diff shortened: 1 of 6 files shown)


cclib/parser/gaussianparser.py
```diff
@@ -2212,5 +2212,29 @@ def extract_charges_spins(line,prop):
...
+        # Rotational constants (GHZ): ************ 12.73690 12.73690
+        # Note: rotational constant will be converted to wavenumber units (1/cm) to standardize across parsers
+        if line[:27] == ' Rotational constants (GHZ)':
+            if not hasattr(self, "rotcons"):
+                self.rotcons = []
+            splits = line.split()
+
+            # Determine if the molecule is linear and only has two constants
+            if '*' in line:  # linear molecule
+                self.rotcons.append([0.0]+[utils.float(splits[i])/29.9792458 for i in (-2, -1)])
```
shivupa, line 2228: Could you use `utils.convertor` here see https://github.com/cclib/cclib/blob/07e7c8e185a684ed7d811f2e96810ff38b0489d0/cclib/parser/utils.py#L77

shivupa (changes requested):
This looks good so far. Just a note about the convertor. I will check on adding a test that covers this

berquist:
`rotcons` isn't actually an attribute: it currently only exists as a calculation method. It needs to be added as a proper attribute. We also need to decide on what the standard units will be and potentially add any conversion factors to the converter (though Shiv already hinted at that).

shivupa:
Per https://github.com/cclib/cclib/issues/256 it seems like we were hoping users would calculate rotational constants from our [method](https://github.com/cclib/cclib/blob/07e7c8e185a684ed7d811f2e96810ff38b0489d0/cclib/method/nuclear.py#L176) 

Would this work with the expected use case you had in mind @amarkpayne

shivupa:
haha @berquist beat me to it

berquist:
Well, I could see wanting to parse what's there rather than recomputing it. If you're doing this for a database (anything more than 100 logfiles, just to pick a number), the MOI tensor calculation and diagonalization will probably become significant.

cclib/parser/gaussianparser.py
amarkpayne: Absolutely! I'll push a fixup commit with this shortly

amarkpayne:
In our use case the user would be parsing hundreds of files, so speed would be preferable here. Let me also look at your method though

shivupa:
I agree then it would make sense to make this a parsed attribute. 
The steps needed to do that would be to
1. Add an [attribute](https://github.com/cclib/cclib/blob/937fdbf036b5002c47021e613077bc64902138fa/cclib/parser/data.py#L158) and [description](https://github.com/cclib/cclib/blob/937fdbf036b5002c47021e613077bc64902138fa/cclib/parser/data.py#L76) to the ccdata object
2. Use `self.set_attribute('rotcons', parsed_rotcons)` to set the rotational constants that you parsed. A similar call for the masses.
3. A test for the `dvb_sp.out` would since this parses the line
`dvb_sp.out: Rotational constants (GHZ):           4.6266363           0.6849065           0.5965900`
So we could add a test in this [file](https://github.com/cclib/cclib/blob/07e7c8e185a684ed7d811f2e96810ff38b0489d0/test/data/testSP.py#L119) where the values of the parsed data is tested

Let us know if you get stuck at any point and we can help you

amarkpayne:
So not to muddy the discussion but I actually was able to try out the `Nuclear.rotational_constants` method really quick (it wasn't too bad to code up, sorry I was unaware of its existence given that the MOPAC parser does things differently) and it seems to work just fine. In the unit tests we run, we only parse 4 logs, but as far as I can tell calculating these numbers did not significantly increase the time compared to everything else we are doing, and we get the same answers (at least within the precision we set in our unit tests).

TLDR it looks like either approach will work for us, so let me know what you prefer. Also let me know if molmass is fine as is as an attribute

ghutchis:
Sorry to jump in, but it's also going to depend on the size of the molecule. The MOI and diagonalization is negligible for v. small molecules, but if you get to 100s of atoms, times 100s of log files, you'll want to parse it if available. For a database in the group, I'd definitely prefer to parse for formats where available.

shivupa:
> TLDR it looks like either approach will work for us, so let me know what you prefer. Also let me know if molmass is fine as is as an attribute

It is up to you. In the short term you could proceed by using the `Nuclear.rotational_constants` method, and we could work on getting this in. If you didn't have the bandwidth to do this, we could take care of putting this in. Now that we know this would be of interest, it makes sense to work on it. Let us know what works for you.

I think we might not need `molmass`,  since we could just sum `atommasses` (an existing attribute) whenever we need `molmass`?

>Sorry to jump in, but it's also going to depend on the size of the molecule. The MOI and diagonalization is negligible for v. small molecules, but if you get to 100s of atoms, times 100s of log files, you'll want to parse it if available. For a database in the group, I'd definitely prefer to parse for formats where available.

Agreed!

berquist:
@amarkpayne Poke. If you use `utils.convertor` to remove the hard-coded constants, then rebase (the test failures are not from your PR), I'd be happy to merge this.

berquist:
Since Mark hasn't responded, I've decided to try finishing this off. It's also parsed for MOPAC, but as wavenumbers, so it needs a conversion. Right now that part of MOPAC isn't tested. I figured that should be in a follow-up PR.

test/data/testGeoOpt.py
```diff
@@ -180,6 +180,24 @@ def testoptstatus(self):
             self.assertEqual(self.data.optstatus[i], self.data.OPT_UNKNOWN)
         self.assertEqual(self.data.optstatus[-1], self.data.OPT_DONE)
 
+    @skipForParser('ADF', 'Not implemented yes')
```
amandadumi, line 183: Looks like this block of skips just needs the comment to be changed from 'yes' -> 'yet'

cclib/parser/gaussianparser.py
```diff
@@ -2254,5 +2256,27 @@ def extract_charges_spins(line,prop):
...
+        # Rotational constants (GHZ):           3.13081     1.24272     0.88960
+        # OR for linear molecules:
+        # Rotational constants (GHZ): ************ 12.73690 12.73690
+        # Note: rotational constant will be converted to wavenumber units (1/cm) to standardize across parsers
+        if line[:28] == ' Rotational constants (GHZ):':
+            splits = line[28:].split()
+
+            # Determine if the molecule is linear and only has two constants
+            if '*' in line:  # linear molecule
+                self.append_attribute("rotconsts", [0.0]+[float(splits[i]) for i in (-2, -1)])
```
amandadumi, line 2270: Is there a way with our current test to check this? I was looking at the 'cclib/data/Gaussian/basicGaussian16/CO_TD_delta.log'  and saw that their rotational constants have 0.0000 instead of the * that this parser seems to be looking for:
' Rotational constants (GHZ):           0.0000000          58.2014785          58.2014785'

amandadumi (changes requested):
Thank you for getting to this and  sorry I wasn't able to yet. But this all looks great, just two questions!

cclib/parser/gaussianparser.py
berquist: I am not seeing it for HF or CO2 either, using 03, 09, or 16; they all print zero.  I see it for Psi4, and maybe others, so it might have been a holdover from looking at something else.

berquist: Ok, I've removed that branch in the conditional.

amandadumi (approved):
This looks good to me! Thank you.

berquist:
poke

shivupa (approved):

Issue #131, opened 2014-09-15
Support fragment calculations in some way


langner:
Calculations based on fragments are getting more popular. Currently in cclib our approach is to parse the supermolecule or largest/complete system in a job. Since we mightwant  to change the API for other reasons (#99), we should try at the same time to accommodate fragments in some way.

Related: #116

Clyde-fare:
For fragments or ONIOM what about modifying the parsing code in logfileparser such that we have something like:

        frag_ids = self.get_frag_ids()
        self.frag_parsers = {frag_id: copy.deepcopy(self) for frag_id in frag_ids}

        for line in inputfile:

            self.updateprogress(inputfile, "Unsupported information", cupdate)
            
            if self.frag_id:
                current_parser = self.frag_parsers[self.frag_id]
                current_parser.extract(inputfile, line)
                self.frag_id = current_parser.frag_id
            else:
                self.extract(inputfile, line)

I.e. we have separate parser objects associated with each fragment (and the master associated with all non-fragment information) the parsers then have sections that  look for lines indicating the start or end a particular fragment section and update their fragment_id attribute to indicate to either empty if we are leaving the fragment info or to the relevant id if we are changing to a new fragment.

Then at we would extract the data from the fragment parsing objects to ccdata objects and attach them to the master ccdata object.

langner:
@Clyde-fare I'm not sure I understand what you mean. What is the point of copying the parser for each fragment, if it is the same? 

(linking to #161 for reference.)

Clyde-fare:
I was thinking that following an initial scan determining how many fragments are involved

A parser for each fragment would be created by copying the main parser.

Then for each iteration of the master loop present in logfileparser a fields frag_id would be used to determine which fragment parser should be used  to parse the next inputline.

I had a go at getting it to work see my [branch](https://github.com/Clyde-fare/cclib/tree/fragment_parsing) but haven't quite got it working yet. 

I though it would be a reasonable idea because data corresponding to the fragments would be naturally separated inside fragments parser and the process should work well if the output appearing in the log file that corresponds to a particular fragment has an identical layout to equivalent output in a normal calculation.

However having tried it out I think there's probably a simpler solution. My main problems were:

* It complicates the master loop present in logfileparser.
* Copying the parsers has to be taken care of as the Logfile object they contain is not copyable.
* Sychronisation between the fragment parser objects is not entirely straightforward. 

[...]

PR #1057, opened 2021-07-27
NWChem Excited states, Vibrational Frequencies and Thermochimistry

BenoitDamota:
Hello,

This is my first contribution, I hope I'm doing it right.

The goal of this PR is to provide code to parse NWChem files with Excited states, Vibrational Frequencies or Thermochimistry calculations. Everything works for my NWChem files. 

The support is not full. At the moment, I raise Exception in the code for ET transitions that are not only alpha (all my calculations are like that). What is the best practices ?

I also have other pieces of code for NWChem, because official cclib does not work on my NWChem files (verbosity problem of my files ?). If you are interested, I can also share this in another branch.

All advices are welcome.
Best regards


shivupa:
Hi Thanks for contributing!
To answer some of your questions:
>The support is not full. At the moment, I raise Exception in the code for ET transitions that are not only alpha (all my calculations are like that). What is the best practices ?The support is not full. At the moment, I raise Exception in the code for ET transitions that are not only alpha (all my calculations are like that). What is the best practices ?

It is ok if the support isn't spin unrestricted. What we usually do if something can't be parsed (rather than raising an exception) is ensure that the attribute does not exist. So in the case of alpha/beta ET transitions being present, the corresponding attributes should not be set. This lets one parse a file with these present for other attributes if desired.

>I also have other pieces of code for NWChem, because official cclib does not work on my NWChem files (verbosity problem of my files ?). If you are interested, I can also share this in another branch.

Any bug fixes (and files to replicate these bugs) are greatly appreciated 😄

BenoitDamota:
Hi,

The last commit improve the ET transitions reading for (spin unrestricted) and fix a bug when no title are provided (in section "NWChem Geometry Optimization"). 

It also add a detection of the number of atoms, that is a temporay hack for my files. All other methods implemented in the parser failed to read natom that is mandatory for many other properties. My next commit will focus on this aspect. 

Don't hesitate to make advices on already submitted code

BenoitDamota:
I identify another blocking point. At several lines of nwchemparser.py, there is :
```python
if hasattr(self, 'linesearch') and self.linesearch:
    return
```
This comes from the reading of steps information :
```python
        #          --------
        #          Step   0
        #          --------
        #
        # and the summary lines that describe the main SCF cycle for the frsit step look like this:
        #
        #@ Step       Energy      Delta E   Gmax     Grms     Xrms     Xmax   Walltime
        #@ ---- ---------------- -------- -------- -------- -------- -------- --------
        #@    0    -379.76896249  0.0D+00  0.04567  0.01110  0.00000  0.00000      4.2
        #                                                       ok       ok
        #
        # However, for subsequent step the format is a bit different:
        #
        #  Step       Energy      Delta E   Gmax     Grms     Xrms     Xmax   Walltime
        #  ---- ---------------- -------- -------- -------- -------- -------- --------
        #@    2    -379.77794602 -7.4D-05  0.00118  0.00023  0.00440  0.01818     14.8
        #                                              ok
        #
        # There is also a summary of the line search (which we don't use now), like this:
        #
        # Line search:
        #     step= 1.00 grad=-1.8D-05 hess= 8.9D-06 energy=   -379.777955 mode=accept
        # new step= 1.00                   predicted energy=   -379.777955
        #
        if line[10:14] == "Step":
            self.geostep = int(line.split()[-1])
            self.skip_line(inputfile, 'dashes')
            self.linesearch = False
        if line[0] == "@" and line.split()[1] == "Step":
            at_and_dashes = next(inputfile)
            line = next(inputfile)
            assert int(line.split()[1]) == self.geostep == 0
            gmax = float(line.split()[4])
            grms = float(line.split()[5])
            xrms = float(line.split()[6])
            xmax = float(line.split()[7])
            if not hasattr(self, 'geovalues'):
                self.geovalues = []
            self.geovalues.append([gmax, grms, xmax, xrms])
            self.linesearch = True
        if line[2:6] == "Step":
            self.skip_line(inputfile, 'dashes')
            line = next(inputfile)
            assert int(line.split()[1]) == self.geostep
            if self.linesearch:
                #print(line)
                return
            gmax = float(line.split()[4])
            grms = float(line.split()[5])
            xrms = float(line.split()[6])
            xmax = float(line.split()[7])
            if not hasattr(self, 'geovalues'):
                self.geovalues = []
            self.geovalues.append([gmax, grms, xmax, xrms])
            self.linesearch = True
```
The problem is that after convergence (or failure), the linesearch variable is always True and prevents to read information of the end of the file and next files. 
To further explain my NWChem workflow , I perform 4 sequential tasks in that order: OPT, PROPERTY (single point to get the Wavefunction), FREQ and TD (singlet). I got millions of molecules like that. The problem occurs only for the task named property since I develop FREQ and TD parsing without testing linesearch value. 

I see several solutions:
1.  remove the tests of linesearch where it is problematic, but I have no idea of the implications and I bet this code was here for a reason
2. complete the tests to parse the information only if they are missing
3. set linesearch to false few lines after the convergence or failure message.

shivupa:
Do you have an example logfile you could share so I could troubleshoot some things to decide how to proceed?  "millions of molecules" is a use case I'd be happy to see cclib used in so I'm eager to help fix these parser issues.

cclib/parser/nwchemparser.py
```diff
@@ -1081,6 +1093,140 @@ def extract(self, inputfile, line):
...
+            self.set_attribute('entropy', utils.convertor(1e-3 * float(line.split()[3]),"kcal/mol","hartree"))
+        
+        # extract vibrational frequencies (in cm-1)
+        if line.strip() == "Normal Eigenvalue ||           Projected Infra Red Intensities":
+            if not hasattr(self, 'vibfreqs'):
+                self.vibfreqs = []
+            if not hasattr(self, 'vibirs'):
+                self.vibirs = []
+            line = next(inputfile) # units
+            line = next(inputfile) # dashes
```
shivupa, line 1132: `self.skiplines` here

cclib/parser/nwchemparser.py
```diff
@@ -1081,6 +1093,140 @@ def extract(self, inputfile, line):
...
+            line = next(inputfile) # first line of data
+            while (line[:-1] != " ----------------------------------------------------------------------------"):
+                self.vibfreqs.append(float(line.split()[1]))
+                self.vibirs.append(float(line.split()[5]))
```
shivupa, line 1136: here (and everywhere you use `float` to convert) could you use `utils.float()` https://github.com/cclib/cclib/blob/e5f66bd48f3f0a78138e7044753789cb1a403106/cclib/parser/utils.py#L62

shivupa (changes requested):
I think this looks pretty good overall. I'd like to add tests. In particular the unit tests `testTD.py` and `testvib.py`. The corresponding files would have to be added to `cclib/data/NWChem/basicNWChemX.X/` and added to https://github.com/cclib/cclib/blob/master/test/testdata. [Here](https://github.com/cclib/cclib/blob/e5f66bd48f3f0a78138e7044753789cb1a403106/test/testdata#L327) and [here](https://github.com/cclib/cclib/blob/e5f66bd48f3f0a78138e7044753789cb1a403106/test/testdata#L358).

It would be nice to include one of the files you provide as a regression test.

BenoitDamota:
Here is an example : 
The inputs [000010042.nw.tar.gz](https://github.com/cclib/cclib/files/6894713/000010042.nw.tar.gz)
The output [000010042.zip](https://github.com/cclib/cclib/files/6894709/000010042.zip)

Yes, millions of molecules computed with NWChem like this 😉 
I'm the computer scientist of the quchempedia@home BOINC project

BenoitDamota:
I can provide this concatenation of the 4 files (OPT, SP, FREQ and TD) for regression test.

[000010042.zip](https://github.com/cclib/cclib/files/6894820/000010042.zip)

cclib/parser/nwchemparser.py
BenoitDamota: Corrected on my working wopy

cclib/parser/nwchemparser.py
BenoitDamota: Can I correct all occurences ? even those that are not mine ?

shivupa: Please do! Much appreciated

shivupa:
> I'm the computer scientist of the quchempedia@home BOINC project

Thanks for the file and now the millions of molecules makes sense!^ 

I can update this PR with a regression test and unit test later today/sometime tomorrow unless you get to this first.

cclib/parser/nwchemparser.py
BenoitDamota: it should be ok

BenoitDamota:
> > I'm the computer scientist of the quchempedia@home BOINC project
> 
> Thanks for the file and now the millions of molecules makes sense!^
> 
> I can update this PR with a regression test and unit test later today/sometime tomorrow unless you get to this first.

I'm not in a hurry, I've been working on this for several years. 
I'm not sure what else is expected of me, especially for tests. It's not my first PR, I discover

In any case, thank you for the help and involvement in moving this project forward.

shivupa:
I opened https://github.com/BenoitDamota/cclib/pull/1/files which should cover the vibrational stuff in this PR. I will add a test for the TDDFT shortly. 
If it looks good please merge it. Note the change in the zpe. In the other parsers only the zpe correction is stored so I updated this to be consistent.

shivupa:
Updated the PR with the TDDFT test

BenoitDamota:
I finally merge the pull request... Thank you for your help
I'm not sure to understand the zpe/zpve modification but i doesn't change the results of my pipeline. So, it should be ok for the chemist.

What about the mentioned issue :

> I identify another blocking point. At several lines of nwchemparser.py, there is :
> 
> ```python
> if hasattr(self, 'linesearch') and self.linesearch:
>     return
> ```
> 
> This comes from the reading of steps information :
> 
> ```python
>         #          --------
>         #          Step   0
>         #          --------
>         #
>         # and the summary lines that describe the main SCF cycle for the frsit step look like this:
>         #
>         #@ Step       Energy      Delta E   Gmax     Grms     Xrms     Xmax   Walltime
>         #@ ---- ---------------- -------- -------- -------- -------- -------- --------
>         #@    0    -379.76896249  0.0D+00  0.04567  0.01110  0.00000  0.00000      4.2
>         #                                                       ok       ok
>         #
>         # However, for subsequent step the format is a bit different:
>         #
>         #  Step       Energy      Delta E   Gmax     Grms     Xrms     Xmax   Walltime
>         #  ---- ---------------- -------- -------- -------- -------- -------- --------
>         #@    2    -379.77794602 -7.4D-05  0.00118  0.00023  0.00440  0.01818     14.8
>         #                                              ok
>         #
>         # There is also a summary of the line search (which we don't use now), like this:
>         #
>         # Line search:
>         #     step= 1.00 grad=-1.8D-05 hess= 8.9D-06 energy=   -379.777955 mode=accept
>         # new step= 1.00                   predicted energy=   -379.777955
>         #
>         if line[10:14] == "Step":
>             self.geostep = int(line.split()[-1])
>             self.skip_line(inputfile, 'dashes')
>             self.linesearch = False
>         if line[0] == "@" and line.split()[1] == "Step":
>             at_and_dashes = next(inputfile)
>             line = next(inputfile)
>             assert int(line.split()[1]) == self.geostep == 0
>             gmax = float(line.split()[4])
>             grms = float(line.split()[5])
>             xrms = float(line.split()[6])
>             xmax = float(line.split()[7])
>             if not hasattr(self, 'geovalues'):
>                 self.geovalues = []
>             self.geovalues.append([gmax, grms, xmax, xrms])
>             self.linesearch = True
>         if line[2:6] == "Step":
>             self.skip_line(inputfile, 'dashes')
>             line = next(inputfile)
>             assert int(line.split()[1]) == self.geostep
>             if self.linesearch:
>                 #print(line)
>                 return
>             gmax = float(line.split()[4])
>             grms = float(line.split()[5])
>             xrms = float(line.split()[6])
>             xmax = float(line.split()[7])
>             if not hasattr(self, 'geovalues'):
>                 self.geovalues = []
>             self.geovalues.append([gmax, grms, xmax, xrms])
>             self.linesearch = True
> ```
> 
> The problem is that after convergence (or failure), the linesearch variable is always True and prevents to read information of the end of the file and next files.
> To further explain my NWChem workflow , I perform 4 sequential tasks in that order: OPT, PROPERTY (single point to get the Wavefunction), FREQ and TD (singlet). I got millions of molecules like that. The problem occurs only for the task named property since I develop FREQ and TD parsing without testing linesearch value.
> 
> I see several solutions:
> 
> 1. remove the tests of linesearch where it is problematic, but I have no idea of the implications and I bet this code was here for a reason
> 2. complete the tests to parse the information only if they are missing
> 3. set linesearch to false few lines after the convergence or failure message.

BenoitDamota:
if it is more relevant and you prefer, we can manage it in another PL or issue?

shivupa:
I opened https://github.com/BenoitDamota/cclib/pull/2 and https://github.com/cclib/cclib-data/pull/156 . The cclib-data PR should be merged first that will have the required regression test file.

@berquist if you could review this that would be appreciated. 

It should be noted that this relates to #131  and #657, here we have a multistep job -> a single ccData object.

BenoitDamota:
> I opened [BenoitDamota#2](https://github.com/BenoitDamota/cclib/pull/2) and [cclib/cclib-data#156](https://github.com/cclib/cclib-data/pull/156) . The cclib-data PR should be merged first that will have the required regression test file.

I test on my files and merge the PR [BenoitDamota#2](https://github.com/BenoitDamota/cclib/pull/2)

shivupa (approved):

cclib/parser/nwchemparser.py
```diff
@@ -93,17 +104,24 @@ def extract(self, inputfile, line):
             self.set_attribute('natom', natom)
 
         if line.strip() == "NWChem Geometry Optimization":
-            self.skip_lines(inputfile, ['d', 'b', 'b', 'b', 'b', 'title', 'b', 'b'])
-            line = next(inputfile)
+            try:
```
berquist, line 107: Rather than be dogmatic and say "we must always use `skip_lines` when trying to reach some new text" or something similar, I'd prefer whatever needs to be rewritten so that the `assert` isn't duplicated.

cclib/parser/nwchemparser.py
```diff
@@ -267,7 +285,7 @@ def extract(self, inputfile, line):
         if line.strip() == "General Information":
 
             if hasattr(self, 'linesearch') and self.linesearch:
-                return
+                self.logger.warning("This file might have multiple jobs or contain a geometry optimization. Attributes might be overwritten multiple times.")
```
berquist, line 288: Can you break this string (and the other times it appears) over each sentence?

cclib/parser/nwchemparser.py
```diff
@@ -480,10 +498,10 @@ def extract(self, inputfile, line):
             at_and_dashes = next(inputfile)
             line = next(inputfile)
             assert int(line.split()[1]) == self.geostep == 0
-            gmax = float(line.split()[4])
```
berquist, line 483: Since this code is being touched, can you replace all the repeated `line.split()` calls (starting on old line 482/new line 500) with a `tokens = line.split()` and use `tokens` everywhere?

cclib/parser/nwchemparser.py
```diff
@@ -495,10 +513,10 @@ def extract(self, inputfile, line):
             if self.linesearch:
                 #print(line)
                 return
-            gmax = float(line.split()[4])
```
berquist, line 498: Same comment here.

cclib/parser/nwchemparser.py
```diff
@@ -1077,10 +1095,146 @@ def extract(self, inputfile, line):
...
+        # Total Entropy                    =  176.764 cal/mol-K
+        # - Translational                =   44.169 cal/mol-K (mol. weight = 448.1245)
+        # - Rotational                   =   37.018 cal/mol-K (symmetry #  =        1)
+        # - Vibrational                  =   95.577 cal/mol-K
+        # Cv (constant volume heat capacity) =  103.675 cal/mol-K
+        # - Translational                  =    2.979 cal/mol-K
+        # - Rotational                     =    2.979 cal/mol-K
+        # - Vibrational                    =   97.716 cal/mol-K
+        if line[1:12] == "Temperature":
+            self.set_attribute('temperature', utils.float(line.split()[2][:-1]))
```
berquist, line 1119: A lot of these lines are long.  Can you put the longer `set_attribute` calls on multiple lines following PEP8?

cclib/parser/nwchemparser.py
```diff
@@ -1077,10 +1095,146 @@ def extract(self, inputfile, line):
...
+        if line[1:32] == "Zero-Point correction to Energy" and hasattr(self, 'scfenergies'):
+            self.set_attribute('zpve', utils.float(line.split()[8]))
+        if line[1:29] == "Thermal correction to Energy" and hasattr(self, 'scfenergies'):
+            self.set_attribute('electronic_thermal_energy', utils.float(line.split()[8]) + utils.convertor(self.scfenergies[-1], "eV", "hartree"))
+        if line[1:14] == "Total Entropy":
+            self.set_attribute('entropy', utils.convertor(1e-3 * utils.float(line.split()[3]),"kcal/mol","hartree"))
+
+        # extract vibrational frequencies (in cm-1)
+        if line.strip() == "Normal Eigenvalue ||           Projected Infra Red Intensities":
+            if not hasattr(self, 'vibfreqs'):
```
berquist, line 1133: Rather than this and the `self.vibfreqs.append` later, prefer `self.append_attribute("vibfreqs", ...)`.  Same goes for `vibirs`.

cclib/parser/nwchemparser.py
```diff
@@ -1077,10 +1095,146 @@ def extract(self, inputfile, line):
...
+        if line[:6] == "  Root":
+            et_num = int(line.split()[1])
+            if not hasattr(self, "etenergies"):
+                self.etenergies = []
+                self.etoscs = []
+                self.etsyms = []
+                self.etsecs = []
+
+            self.etenergies.append(utils.convertor(utils.float(line.split()[-2]), "eV", "wavenumber"))
+            self.etsyms.append(str.join(" ", line.split()[2:-4]))
```
berquist, line 1192: Same comments about `append_attributes`.

berquist (changes requested):
`float` is being replaced (everywhere?) by `utils.float`.  Because of its definition https://github.com/cclib/cclib/blob/b7340921b4df38869606fbf7bffdb43e586db786/cclib/parser/utils.py#L62 I think it will noticeably slow things down.  Can you confirm that making this replacement is necessary everywhere?  If it isn't, I'd avoid it.  I recognize that we don't have any docs to provide guidance about this.

BenoitDamota:
I'm not sure I understand how the pull request ends. Do I need to make any other modifications?

shivupa:
No this should be done shortly, and nothing more is required on your end.
The path forward:

- The example job broke the CI since we don't support multiple jobs in 1 file. This needed to be removed from testing for now. Relevant PR that needs to be merged prior to this PR https://github.com/cclib/cclib-data/pull/159
- This should then pass CI and be approved+merged.
- As this stands this won't work fully for your use case (without preprocessing the files to split the outputs into separate files). The change you proposed won't work as a permanent solution unfortunately. You'd have to decide, but perhaps working from a branch with the changes you mentioned, might be necessary for immediate use.
- The permanent solution is being addressed in https://github.com/cclib/cclib/pull/1076 . This is a somewhat large change.

Appreciate your patience.

berquist (approved):

BenoitDamota:
indeed, it was fast! Thanks guys, I'll look at the PR for the files with multiple steps. If I can help...

PR #1062, opened 2021-08-13
pin h5py to fix CI

shivupa:
PySCF wheels require an older version of h5py. This is necessary for now.


berquist (approved):

berquist:
Looks like this will be unnecessary in the near future: https://github.com/pyscf/pyscf/issues/1003#issuecomment-871619435

PR #1065, opened 2021-08-14
Support ORCA 5

ATenderholt:
Looks like formatting changes were introduced in ORCA 5.


cclib/parser/orcaparser.py
```diff
@@ -1429,17 +1433,40 @@ def energy_intensity(line):
...
         if line[:11] == "IR SPECTRUM":
-            self.skip_lines(inputfile, ['d', 'b', 'header', 'd'])
+            if self.metadata['package_version'][0] == '4':
+                self.skip_lines(inputfile, ['d', 'b', 'header', 'd'])
+                regex = r'\s+(?P<num>\d+):\s+(?P<frequency>\d+\.\d+)\s+(?P<intensity>\d+\.\d+)'
+            elif self.metadata['package_version'][0] == '5':
+                self.skip_lines(inputfile, ['d', 'b', 'header', 'units', 'd'])
+                regex = r'\s+(?P<num>\d+):\s+(?P<frequency>\d+\.\d+)\s+(?P<eps>\d+\.\d+)\s+(?P<intensity>\d+\.\d+)'
+            else:
+                raise Exception('Unsupported version: ' + self.metadata['package_version'])
```
ATenderholt, line 1457: Is there a better approach besides checking for 4 vs 5 in version numbers?

test/data/testvib.py
```diff
@@ -412,7 +412,7 @@ class GaussianRamanTest(GenericRamanTest):
 class OrcaRamanTest(GenericRamanTest):
     """Customized Raman unittest"""
 
-    max_raman_intensity = 1048
+    max_raman_intensity = 1045
```
ATenderholt, line 415: Orca 5 has a lower max raman intensity of 1037.

cclib/parser/orcaparser.py
berquist: Nope, this is actually one of the advantages of having the package version available IMO.

ATenderholt: It's good that package version was available, but if there were (many) more than just these two cases, it seems like another approach would be better.

shivupa (approved):

cclib/parser/orcaparser.py
shivupa: if package version is not parsed (if future versions change the format) then this would result in a KeyError right? Can we check if `package_version` is present and if not just take the newer regex.

shivupa (approved):

shivupa:
Just had a minor change.

ATenderholt:
@shivupa Good call on handling cases where `package_version` might not exist. I've pushed an update.

shivupa:
Thanks! Looks great. If there are no comments by the end of the day we can merge this

ATenderholt:
Looks like failure on code coverage due to the extra if-statement. Which is odd since it didn't complain about the similar one before.

Is there a way to add a test case for this? Or should I just ignore it since it doesn't seem to affect overall coverage (still at 89.47%)?

shivupa:
I'm fine with leaving it. I'll merge now.

PR #1070, opened 2021-09-12
Switch to atomic units with future flag

berquist:
#89 can be closed once this is no longer hidden behind `--future`.


berquist:
Closing in favor of making a PR that merges into `master` and does away with hiding behind the `future` flag.

PR #1072, opened 2021-09-19
QChem: add 5.4 unit tests

berquist:
Part of #977, this should go in before adding the lines to `testdata` because a number of things are broken.

This also adds the formatted checkpoint file associated with each calculation.


shivupa (approved):

shivupa:
> a number of things are broken.

Meaning in the parsing of these files things fail? i am good with putting it and working through the failures

berquist:
> > a number of things are broken.
> 
> Meaning in the parsing of these files things fail? i am good with putting it and working through the failures

Yes. For some reason extra (?) MO coefficient matrices at the end of SCF are getting printed, and the logic for figuring out which column we're on is broken.  I feel like I've seen that before, maybe it can be fixed once and for all with the next PR.

PR #1076, opened 2021-09-26
Multiple jobs in one output file

### LINKED ISSUE #657: Support multistep jobs generally

shivupa:
This is a (?proposal) WIP PR to allow for the parsing of multiple jobs in a single file. 

For a simple example:
```
# to form temp:
# cat dvb_sp_hf.out &>> temp
# cat dvb_sp_hf.out &>> temp

import cclib
print("Single Job Single File")
a = cclib.ccopen("dvb_sp_hf.out")
print(a.parse())

print("Multiple Jobs Single File")
b = cclib.ccopen("temp")
print(b.parse())
```
The single job still creates a single ccData object and the multiple job file creates a list of ccData objects:
```
Single Job Single File
<cclib.parser.data.ccData_optdone_bool object at 0x7f61d99cb6a0>
Multiple Jobs Single File
[<cclib.parser.data.ccData_optdone_bool object at 0x7f61db8f85e0>, <cclib.parser.data.ccData_optdone_bool object at 0x7f61db8f8610>]
```

Things that I know still need to be fixed:

- [ ] ccframe
- [ ] ccwrite
- [ ] False positives? (BOMD, Geometry scans, others)  

Closes #657

### DIFF
--- a/cclib/parser/gamessparser.py
+++ b/cclib/parser/gamessparser.py
@@ -1588,7 +1588,6 @@ class GAMESS(logfileparser.Logfile):
             self.set_attribute('entropy', utils.convertor(float(thermoValues[6])/1000.0,"kcal/mol","hartree"))
 
 
-        if line[:30] == ' ddikick.x: exited gracefully.'\
-                or line[:41] == ' EXECUTION OF FIREFLY TERMINATED NORMALLY'\
+        if line[:41] == ' EXECUTION OF FIREFLY TERMINATED NORMALLY'\
                 or line[:40] == ' EXECUTION OF GAMESS TERMINATED NORMALLY':
             self.metadata['success'] = True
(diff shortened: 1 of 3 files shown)


berquist:
Can you put this behind the `--future` flag, which we can then remove/make default for 2.x?

berquist:
Poke for rebase

berquist:
I changed the milestone to the 2.0 alpha but we're only going to implement this for the new parser combinators, not the old/existing infrastructure.

PR #1085, opened 2021-12-21
Orca: fix for oscillator strength

felixplasser:
Orca prints two values for the oscillator strengths, the `ELECTRIC DIPOLE MOMENTS` and `VELOCITY DIPOLE MOMENTS`.

At the moment, the `VELOCITY DIPOLE MOMENTS` are parsed by cclib. But, actually, the `ELECTRIC DIPOLE MOMENTS`, i.e. the length representation, are the thing one usually talks about. Also the comment on line 1102 suggests that cclib wants to actually parse the `ELECTRIC DIPOLE MOMENTS`. So, I would suggest changing this line in the parser to make sure that the `ELECTRIC DIPOLE MOMENTS` are read in. Is that ok?

In addition, I am including a small fix for calculations including SOC.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -1117,7 +1117,7 @@ Dispersion correction           -0.016199959
             self.extend_attribute('etsyms', etsyms)
 
         # Parse the various absorption spectra for TDDFT and ROCIS.
-        if 'ABSORPTION SPECTRUM' in line or 'ELECTRIC DIPOLE' in line:
+        if ('ABSORPTION SPECTRUM' in line and 'ELECTRIC DIPOLE' in line) and not 'SOC' in line:
             # CASSCF has an anomalous printing of ABSORPTION SPECTRUM.
             if line[:-1] == 'ABSORPTION SPECTRUM':
                 return
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -1117,7 +1117,7 @@ Dispersion correction           -0.016199959
             self.extend_attribute('etsyms', etsyms)
 
         # Parse the various absorption spectra for TDDFT and ROCIS.
-        if ('ABSORPTION SPECTRUM' in line and 'ELECTRIC DIPOLE' in line) and not 'SOC' in line:
+        if 'ABSORPTION SPECTRUM' in line or 'ELECTRIC DIPOLE' in line:
             # CASSCF has an anomalous printing of ABSORPTION SPECTRUM.
             if line[:-1] == 'ABSORPTION SPECTRUM':
                 return
@@ -1126,9 +1126,11 @@ Dispersion correction           -0.016199959
 
             # Standard header, occasionally changes
             header = ['d', 'header', 'header', 'd']
+            energy_intensity = None
 
-            def energy_intensity(line):
-                """ TDDFT and related methods standard method of output
+            if line == "ABSORPTION SPECTRUM VIA TRANSITION ELECTRIC DIPOLE MOMENTS":
+                def energy_intensity(line):
+                    """ TDDFT and related methods standard method of output
 -----------------------------------------------------------------------------
          ABSORPTION SPECTRUM VIA TRANSITION ELECTRIC DIPOLE MOMENTS
 -----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


schneiderfelipe:
Hi @felixplasser the relevant parts of the failing tests are

https://github.com/cclib/cclib/blob/5a88163858f97f7419007e2e152045ae26be7eec/test/data/testTD.py#L151

and

https://github.com/cclib/cclib/blob/5a88163858f97f7419007e2e152045ae26be7eec/test/data/testTD.py#L182-L191

Would you mind taking a look at those values?

@shivupa I'm not sure whether they were cross-validated between different packages (the first value shows up in other tests such as [here](https://github.com/cclib/cclib/blob/5a88163858f97f7419007e2e152045ae26be7eec/test/data/testTD.py#L154-L163), for QChem).

shivupa:
This seems reasonable I just want to check some test failures. They are all probably because the tests are checking for the "wrong" number I just want to compare to the other codes.

shivupa:
> 



> @shivupa I'm not sure whether they were cross-validated between different packages (the first value shows up in other tests such as [here](https://github.com/cclib/cclib/blob/5a88163858f97f7419007e2e152045ae26be7eec/test/data/testTD.py#L154-L163), for QChem).

Generally we haven't been cross validating between packages, but I was hoping they would show up in another parser. Thanks for checking!

felixplasser:
Hi, thanks for looking at this!

Yes, all the orca `testoscs` would fail after this because they are checking for the "velocity" values rather than the normal dipole values.

I am just trying to see how I can actually run the tests. Then I'll get back.

berquist:
> I am just trying to see how I can actually run the tests. Then I'll get back.

The GHA workflow is running
```bash
bash ./.github/scripts/run_pytest.bash
```
but you may prefer
```bash
bash ./.github/scripts/run_unittest.bash
```
which doesn't run coverage checking, so it's a little faster.  To only run ORCA unit tests, you can do (also from the top level directory)
```bash
python -m test.test_data --terse ORCA
python -m test.regression --traceback ORCA
```

felixplasser:
I am still not quite sure how to run the tests but things look consistent with what @schneiderfelipe  said

If I look at `basicORCA4.2/dvb_td.out`, then I find

~~~~
         ABSORPTION SPECTRUM VIA TRANSITION ELECTRIC DIPOLE MOMENTS
-----------------------------------------------------------------------------
State   Energy    Wavelength  fosc         T2        TX        TY        TZ  
        (cm-1)      (nm)                 (au**2)    (au)      (au)      (au) 
-----------------------------------------------------------------------------
   1   43167.6    231.7   0.005629700   0.04293  -0.15576  -0.13665  -0.00000
   2   46157.0    216.7   1.165369790   8.31192   2.83128   0.54387   0.00000
...
         ABSORPTION SPECTRUM VIA TRANSITION VELOCITY DIPOLE MOMENTS
-----------------------------------------------------------------------------
State   Energy    Wavelength   fosc        P2         PX        PY        PZ  
        (cm-1)      (nm)                 (au**2)     (au)      (au)      (au) 
-----------------------------------------------------------------------------
   1   43167.6    231.7   0.000205421   0.00006   0.00770  -0.00115  -0.00000
   2   46157.0    216.7   0.086136273   0.02717  -0.16446  -0.01115  -0.00000
~~~~

The original test requires a value of 0.09, which is consistent with "VELOCITY". But a value of 1.165 would be consistent with "ELECTRIC DIPOLE" according to the change in the code I made.

felixplasser:
The Q-Chem job in the test suite does not seem to have a consistent geometry. But if I run a TDA-B3LYP/STO-3G job with the geometry from the Orca file in Q-Chem, then I get the following:

~~~~
 Excited state   7: excitation energy (eV) =    5.7231
 Total energy for state  7:                  -382.09795419 au
    Multiplicity: Singlet
    Trans. Mom.: -2.8817 X   0.0552 Y   0.0000 Z
    Strength   :     1.1648182819
~~~~

So, it seems like also Q-Chem agrees with my version of the oscillator strengths

felixplasser:
I updated the unit tests and everything works now if I run

~~~~
python -m test.test_data --terse ORCA
~~~~

But the strange thing about the `td_rocis` test is that it produces 8 different spectra. So, one would have to make really sure that one ends up with the required values.

berquist:
> But the strange thing about the `td_rocis` test is that it produces 8 different spectra. So, one would have to make really sure that one ends up with the required values.

Yes, that was a conscious decision by Jonathon when he wrote all that code, but it is supposed to be behind a sort of "unstable" attribute (`transprop`, not `etoscs`) until we figure out how to present multiple spectra properly across all programs and job types. We would conceptually encounter a similar problem doing TDA followed by TDDFT in Q-Chem, but I think there we go with the usual package-level convention of taking the last section that appears.

felixplasser:
> > But the strange thing about the `td_rocis` test is that it produces 8 different spectra. So, one would have to make really sure that one ends up with the required values.
> 
> Yes, that was a conscious decision by Jonathon when he wrote all that code, but it is supposed to be behind a sort of "unstable" attribute (`transprop`, not `etoscs`) until we figure out how to present multiple spectra properly across all programs and job types. We would conceptually encounter a similar problem doing TDA followed by TDDFT in Q-Chem, but I think there we go with the usual package-level convention of taking the last section that appears.

I agree, in most cases it is perfectly fine to just take the last instance of the oscillator strengths.

danielhollas:
@felixplasser thanks for catching this and for the fix, I was banging my over this for quite some time.

@shivupa are there any blockers for merging this to master? I need to install cclib in a Docker container in an automated fashion and don't want to introduce a dependency on a forked repo. Thanks!

shivupa:
Hi! The tests are failing. I will fix the tests and clean up this tonight. Thanks for your patience!

shivupa:
>In addition, I am including a small fix for calculations including SOC.
The reason for this fix wasn't apparently to me. If there is an offending output could this be provided and we can make a fix/regression test?

I rebased and cleaned up this PR. There was an issue in the logic so there is some restructuring. This should now behave as expected. Let me know if this works for everyone.

berquist (approved):

langner (approved):

amandadumi:
Thanks so much for the contribution!

PR #1086, opened 2021-12-28
Pytest filter through GitHub actions

amandadumi:
Introduces filtering for tests. 

The core set of tests for parsers run for every pull request, while the method related tests will only run when there is a change within the methods folder.

### DIFF
--- a/.github/workflows/cclib_pytest.yml
+++ b/.github/workflows/cclib_pytest.yml
@@ -25,10 +25,21 @@ jobs:
       - name: Install cclib
         run: |
           python -m pip install .
-      - name: Test with pytest
+      - name: Filter code
+        uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            methods:
+              - 'cclib/cclib/method/**'
+      - name: Test core code with pytest
         run: |
           env | sort
           bash .github/scripts/run_pytest.bash
+      - name: Test method code with pytest
+        if: steps.filter.outputs.method == 'true' 
+        run: |
+          bash .github/scripts/run_pytest_methods.bash
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
         with:
--- a/.github/scripts/run_pytest.bash
+++ b/.github/scripts/run_pytest.bash
@@ -5,7 +5,7 @@
 
 set -euxo pipefail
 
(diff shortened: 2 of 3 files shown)


shivupa:
LGTM we will just have to think about codecov

amandadumi:
Agreed. With pytest-cov, it would be easy to collect coverage for the methods only when the methods tests are triggered. Additionally, certain directories could be omitted when when only the core tests are run. However, Im not sure you can bring these together. Still looking into this.

berquist (approved):
So when regressions are run, I *think* the coverage results are added to the existing ones on disk from the unit tests. If that's true, then it might be as "simple" as saving the previous coverage results somewhere and restoring them before the run.

amandadumi:
Updating here, I think the coverage issue can be handled through https://docs.codecov.com/docs/carryforward-flags. With these in place, it will carry forward testing data from a previous commit if certain flags are used. 

I am still working for figure out how to get this to work with github actions. I am hoping it is as easy as just having a coverage.yml file in the .github folder and small changes to the cclib_pytest.yml file.

.github/codecov.yml
```diff
@@ -0,0 +1,35 @@
...
+flags:
+  method:
+    paths:
+      - cclib/method/
+    carryforward: true
+  unit:
+    paths:
+      - cclib/io/
+      - cclib/parser/
+      - cclib/parser/
```
berquist, line 33: duplicate line?

amandadumi:
I've reverted this commit to just include the changes necessary to filter with pytest, i.e. nothing is included to address codecov changes. I will open a separate pull request for this.

.github/workflows/cclib_pytest.yml
```diff
@@ -25,10 +25,22 @@ jobs:
...
+            methods:
+              - 'cclib/cclib/method/**'
+      - name: Test core code with pytest
         run: |
           env | sort
           bash .github/scripts/run_pytest.bash
+      - name: Test method code with pytest
+        if: steps.filter.outputs.method == 'true' 
+        run: |
+          env | sort
```
berquist, line 42: You can probably get rid of this `env | sort` since it will just print the same stuff as the one above.

berquist (approved):

PR #1094, opened 2022-01-24
Fix deprecation warning on getargspec

arosen93:
I have a vendetta against `DeprecationWarning`'s, so here is a quick fix for the `inspect.getargspec()` command used in `cclib.parser.logfileparser.`:
```
...\cclib\parser\logfileparser.py:276: DeprecationWarning: inspect.getargspec() is deprecated since Python 3.0, use inspect.signature() or inspect.getfullargspec()
    if len(inspect.getargspec(self.extract)[0]) != 3:

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```


berquist (approved):
Thanks Andrew!

PR #1095, opened 2022-01-25
coverage actions v2

amandadumi:
As described [here](https://github.com/codecov/codecov-action) version 1 of coverage-action is being sunset and we should switch to v2. 

I'm not familiar enough with cclib codecov set up yet, so I am not sure if the TOKEN is uploaded as described in the note on their page:
"Note: This assumes that you've set your Codecov token inside Settings > Secrets as CODECOV_TOKEN"


shivupa (approved):

shivupa:
Seems like nothing breaks switching to 2. LGTM

PR #1101, opened 2022-02-04
introducing carryforward flags

amandadumi:
This will introduce carry forward flags to allow conditional testing work well with codecov.  

Since we can create a separate reports methods,  the carry forward flags will let us pull an old report if a new report isn't generated.  More documentation can be found here: https://docs.codecov.com/docs/carryforward-flags

This commit isn't incorporating the filtering yet, but just the flags to enable it. One more commit will be needed in which `.github/workflows/cclib_pytest.yml` will change what reports are uploaded.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -13,14 +13,6 @@ coverage:
         flags:
           - unit
 
-comment:
-  layout: "reach, diff, flags, files"
-  behavior: default
-  require_changes: false  
-  require_base: yes
-  require_head: yes       
-  branches: null
-
 flags:
   method:
     paths:
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,34 @@
+
+coverage:
+  status:
+    project:
+      default:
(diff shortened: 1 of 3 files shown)


berquist:
> One more commit will be needed
In this PR, or will it be separate?

amandadumi:
I was planning on a separate PR. I was hoping this PR could trigger the creation of the coverage_methods.xml, however I might have done this in the wrong order since the pytest filter was merged first so the methods test may not be run. I'll have to check the coverage report here closely.

langner:
From discussion: we should try adding the cron job before this so the filtered out tests are carried through.

amandadumi:
I believe this is set up correctly. I am setting the flag for methods to be false so we can generate a coverage-method.xml, which can be "carried forward" when methods haven't changed for a pull request. Once this is merged, one more pull request will be needed to turn this flag off. (I think)

shivupa (approved):

berquist (approved):

Issue #1110, opened 2022-02-10
Missing Coupled Cluster energies in Gaussian C.01


oliver-s-lee:
As the title suggests, ccenergies are not available for CCD or CCSD calculations (although 'CCSD' can still be found in metadata['methods']) in newer versions of Gaussian 16.

The problem appears to be a change in capitalisation in this line of the .log file compared to earlier Gaussian versions:
Gaussian A.03 (test data):
`            Population analysis using the SCF density.`
Gaussian C.01 (D in Density is now capitalised):
`            Population analysis using the SCF Density.`

This is significant because this line is used as a trigger to save ccenergies in the Gaussian parser:
```
        if line[12:53] == "Population analysis using the SCF density":
            if hasattr(self, "ccenergy"):
                if not hasattr(self, "ccenergies"):
                    self.ccenergies = []
                self.ccenergies.append(utils.convertor(self.ccenergy, "hartree", "eV"))
                del self.ccenergy
```

This can be fixed by checking for both strings, but might fail again in the future if Gaussian changes something else random in the line. Also, if population analysis is not performed (eg, pop=none), this line won't appear at all, so probably isn't a great choice of trigger (see below example without pop analysis).

As far as I can see, there's no reason not to process ccenergies in after_parsing() (maybe there is and I don't see it?), so I would propose to move it there?

Relevant section of the log file without pop analysis:
```
**********************
 DD1Dir will call FoFMem   1 times, MxPair=        20
 NAB=    10 NAA=     0 NBB=     0.
 A-vector RMS=  3.2708823D-08 Max=  7.5784157D-08 conv= 1.00D-05.
 RLE energy=       -0.0530492069
 DE(Corr)= -0.53049212E-01 E(CORR)=     -75.017378003     Delta=-3.16D-08
 NORM(A)=   0.10148087D+01
 CI/CC converged in    8 iterations to DelEn=-3.16D-08 Conv= 1.00D-07 ErrA1= 3.27D-08 Conv= 1.00D-05
 Wavefunction amplitudes converged. E(Corr)=     -75.017378003    
 Largest amplitude= 8.18D-02
 Discarding MO integrals.
 Leave Link  913 at Thu Feb 10 17:43:27 2022, MaxMem=   104857600 cpu:               0.1 elap:               0.1
 (Enter /gpfs1/apps/software/Gaussian16/AVX//g16/l9999.exe)
 Unable to Open any file for archive entry.
[...]

PR #1111, opened 2022-02-10
Gaussian C.01 CC Fix

oliver-s-lee:
To address #1110

### DIFF
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -162,6 +162,12 @@ class Gaussian(logfileparser.Logfile):
                 rotation = utils.get_rotation(numpy.array(inputcoord), numpy.array(atomcoord))
                 grads_std.append(rotation.apply(grad))
             self.set_attribute('grads', numpy.array(grads_std))
+        
+        if hasattr(self, "ccenergy"):
+            if not hasattr(self, "ccenergies"):
+                self.ccenergies = []
+            self.ccenergies.append(utils.convertor(self.ccenergy, "hartree", "eV"))
+            del self.ccenergy
 
     def extract(self, inputfile, line):
         """Extract information from the file object inputfile."""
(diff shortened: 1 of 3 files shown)


cclib/parser/gaussianparser.py
```diff
@@ -162,6 +162,12 @@ def after_parsing(self):
                 rotation = utils.get_rotation(numpy.array(inputcoord), numpy.array(atomcoord))
                 grads_std.append(rotation.apply(grad))
             self.set_attribute('grads', numpy.array(grads_std))
+        
+        if hasattr(self, "ccenergy"):
+            if not hasattr(self, "ccenergies"):
+                self.ccenergies = []
+            self.ccenergies.append(utils.convertor(self.ccenergy, "hartree", "eV"))
```
berquist, line 169: These three lines can be replaced with
```suggestion
            self.append_attribute("ccenergies", utils.convertor(self.ccenergy, "hartree", "eV"))
```

data/Gaussian/basicGaussian16/water_ccd.log
```diff
@@ -1,10 +1,11 @@
- Entering Gaussian System, Link 0=g16_main
```
berquist, line 1: Rather than update the unit test files, can you instead add these files to https://github.com/cclib/cclib-data/, then add tests to https://github.com/cclib/cclib/blob/master/test/regression.py that check for `ccenergies` for that file?  If you need help understanding the test format in that file, let us  know.

berquist (changes requested):

data/Gaussian/basicGaussian16/water_ccd.log
amandadumi: The test file has been merged into cclib-data which should allow the regression test to run once added.

amandadumi:
This looks good, thanks for contributing this fix!

PR #1115, opened 2022-02-27
Update README.md with the Google Groups link

schneiderfelipe:
This updates a link to the old mailing list. Kind of a follow-up of #1107.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,4 +10,4 @@
 
 cclib is a Python library that provides parsers for output files of computational chemistry packages. It also provides a platform for computational chemists to implement algorithms in a platform-independent way.
 
-For more information, go to [https://cclib.github.io](https://cclib.github.io). There is a mailing list for questions at cclib-users@lists.sourceforge.net.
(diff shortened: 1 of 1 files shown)


berquist:
The format is fine, but the link doesn't work for me for some reason. The working link is https://groups.google.com/u/1/g/cclib. I think (but don't know) that `u/1` is actually dynamic and refers to which Google account I'm signed into, so shouldn't be a part of the permanent link, meaning that isn't right either. Can others confirm/deny?

mscho527:
I can confirm that the link Felipe used on the PR works fine for me. I think you are correct in that the `u/1` part comes up when multiple accounts are signed in. I have three Google accounts signed in and the link on the PR shows the Google Group page with the default Google account (which is not what I use for cclib mailing list). Maybe your default account (if it is an organization account) has different policies for Google Groups?

berquist:
Yes, that's exactly the problem.

berquist (approved):

Issue #1122, opened 2022-03-28
Should we have some sort of code style?


schneiderfelipe:
I notice there's not much consistency across the codebase in some cases (e.g., `'strings'` and `"strings"`, etc.). **Any** code style is good enough (I really like [black](https://github.com/psf/black)). This could be enforced in a post-commit or post-merge hook.

arosen93:
I also would love to see `black .` applied to the repo, but I will `black` just about anything...

mscho527:
I think people were asked to run `black -l 100` on any file they changed in the commit. As far as I remember, there was a discussion on whether we want to run `black` on the entire repo some time ago, but we decided not to because it would make it harder to track changes through git history. Please correct me if I am not recalling the discussion correctly!

berquist:
> I think people were asked to run `black -l 100` on any file they changed in the commit. As far as I remember, there was a discussion on whether we want to run `black` on the entire repo some time ago, but we decided not to because it would make it harder to track changes through git history. Please correct me if I am not recalling the discussion correctly!

This is correct on all points. I'm not sure it would be as bad as some renames or moves: I needed multiple revision checkouts to see the original history of `io`.

One more concrete (and common) concern is that it will mess with all open PRs, so we should try and push them through before running a formatter. Rebasing them after formatting the codebase will be a pain.

I use black as part of https://pre-commit.com/ (but not in GHA) for all my other projects. I think we should do whatever makes contributing to the code and modifying it easier.

shivupa:
What Minsik said was indeed true. I have some new thoughts about this.

I agree with making it easier to contribute. Personally I think precommit hooks are a little tedious on developers since it requires them to set it up, but I am open to discussion on that point.

I do think just running `black -l 100` on the source and adding a [github action](https://black.readthedocs.io/en/stable/integrations/github_actions.html) to check if future contributions abide by the constraints sounds like a good option.

berquist:
I agree with not requiring pre-commit hooks, at least to start off with; I usually just leave the hook uninstalled and call `pre-commit run -a` occasionally which runs black/linters/mypy/... with the specification preconfigured in a file (`.pre-commit-config.yaml`) so I don't need to remember the exact command to run for each tool:

``` yaml
---
fail_fast: true
default_language_version:
  python: python3.10
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: "v4.3.0"
    hooks:
      - id: trailing-whitespace
[...]

PR #1123, opened 2022-03-31
Use black

schneiderfelipe:
This is so that we can take a look at how the repository will look like if we run [`black -l 100`](https://github.com/psf/black) over it. This PR would solve #1122. I also sorted import statements with [`isort`](https://github.com/PyCQA/isort) and rerun `black -l 100`. I separated it in three different commits so that we can see the differences.

A bird's eye view on the changes using [`cloc`](https://github.com/AlDanial/cloc):

**original**:

```
$ cloc .
...
--------------------------------------------------------------------------------
Language                      files          blank        comment           code
--------------------------------------------------------------------------------
...
Python                          271           7601          10591          25507
...
```

**this PR**:

```
$ cloc .
...
--------------------------------------------------------------------------------
Language                      files          blank        comment           code
--------------------------------------------------------------------------------
...
Python                          271           7662          10572          26166
...
```

EDIT: we could enforce or require some [pre-commit hook](https://pre-commit.com/) to maintain this in the future.


schneiderfelipe:
`isort` was indeed the problem.

berquist:
I've added a preliminary `pyproject.toml` for just black and isort (because we shouldn't use black without isort) and a plausible pre-commit config, though I haven't figured out how to get it to also exclude data files outside of `data` (with something like `\.(adfin|adfout|arc|dal|dat|com|cube|fchk|gjf|in|inp|log|nw|mop|out|txt)$ `).

cclib/parser/data.py
```diff
@@ -105,87 +105,91 @@ class ccData:
...
+        "transprop": Attribute(dict, "electronic transitions", "transitions"),
+        "vibanharms": Attribute(numpy.ndarray, "anharmonicity constants", "vibrations"),
+        "vibdisps": Attribute(numpy.ndarray, "displacement", "vibrations"),
+        "vibfreqs": Attribute(numpy.ndarray, "frequencies", "vibrations"),
+        "vibfconsts": Attribute(numpy.ndarray, "force constants", "vibrations"),
+        "vibirs": Attribute(numpy.ndarray, "IR", "vibrations:intensities"),
+        "vibramans": Attribute(numpy.ndarray, "raman", "vibrations:intensities"),
+        "vibrmasses": Attribute(numpy.ndarray, "reduced masses", "vibrations"),
+        "vibsyms": Attribute(list, "vibration symmetry", "vibrations"),
+        "zpve": Attribute(float, "zero-point correction", "properties:energies"),
```
berquist, line 182: Same comment about alignment here.

berquist (commented):
I didn't look at all the changes, but of course most of them are positive and minor.  We will have to make decisions about some explicitly-aligned code.

berquist:
Closing because this was just to see how things would change, and we're going to apply these changes once (most of) the PR queue is cleared out.

Issue #1077, opened 2021-10-06
Replace pyquante with something


langner:
Top two candidates:
- [probably preferred] gau2grid (https://github.com/dgasmith/gau2grid)
- pyscf

cryos:
I spent quite a bit of time trying to get `gau2grid` to to electron density and molecular orbitals - I never got it to produce anything useful, and it seems like it might be too low level. If you do I would be very interested, but I gave up after multiple discussions with the main author of the code.

PR #1126, opened 2022-04-08
carryforward flag switch

amandadumi:
This  turns the method filter flag to "true". This completes the carryforward initial implementation, ensuring  coverage can remain accurate as we filter tests. Allows the methods tests to only run when a file in the methods folder has changed #1086 , but ensure the previous methods coverage report is carried forward and uploaded to codecov if it hasn't been regenerated for the current pull request ( #1101 and this pull request).


amandadumi:
This did not work as I expected. I'll look into it!

shivupa:
What is the status on this?

amandadumi:
Hi! This is on the back burner until I address #1077 is addressed.

PR #1136, opened 2022-06-05
Add method for computing CM5 atomic charges

### LINKED ISSUE #676: Method for calculating CM5 charges
I recently came across [this repository](https://github.com/hokru/cm5charges) for calculating CM5 charges from ORCA outputs. If we have access to atomic radii through one of the periodic table packages, it can be implemented as a cclib method.

berquist:
Closes #676. This replaces #852.

As one of the requirements, it also adds Hirshfeld charge reading from ORCA.

### DIFF
--- a/cclib/method/__init__.py
+++ b/cclib/method/__init__.py
@@ -10,6 +10,7 @@
 from cclib.method.bader import Bader
 from cclib.method.bickelhaupt import Bickelhaupt
 from cclib.method.cda import CDA
+from cclib.method.cm5 import CM5
 from cclib.method.cspa import CSPA
 from cclib.method.ddec import DDEC6
 from cclib.method.density import Density
--- a/cclib/method/cm5.py
+++ b/cclib/method/cm5.py
@@ -30,7 +30,7 @@ class CM5(Method):
     Parameters
     ----------
     data : ccData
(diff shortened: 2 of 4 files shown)


berquist:
Once everything is confirmed working I'll break this up into a separate PR that's first just parsing Hirshfeld and/or CM5 atomic charges.

berquist:
Notes to self:
- This still needs at least a test against parsed CM5.
- Methods for computing dipole and quadrupole moments from CM5 charges were originally available.  Investigate if we have a general interface for doing this with charges and our existing moments functionality.  For CM5 it looks like there was also a scaling factor.

berquist:
The scaling factor of 1.20 comes from https://doi.org/10.1021/acs.jctc.5b00414, which supersedes an earlier recommendation of 1.27 from https://doi.org/10.1021/ct500016d.

berquist:
Implementing the scaling factor into the moments code is unnecessary, since it works out that you can just multiply the dipole/quadrupole after the fact and get the right answer.  This means there is nothing to do regarding the scaling factor.

shivupa (changes requested):
Minor thing

shivupa (approved):

PR #1137, opened 2022-06-06
Parse Hirshfeld and CM5 analyses from ORCA, Gaussian, and Q-Chem

berquist:
First part of #1136.


mscho527 (approved):
This looks great! One thing minor -- I think this wouldn't be an issue in most cases, but out of precaution, we probably want to make it clear in docs (when those are added) that the `somemethod_sum` lists are atomic charges with hydrogen summed into heavy atom. (one might, I suppose, mistakenly think that it is sum of all atomic charges -- a float approximately the total charge on the molecule)

berquist:
> This looks great! One thing minor -- I think this wouldn't be an issue in most cases, but out of precaution, we probably want to make it clear in docs (when those are added) that the `somemethod_sum` lists are atomic charges with hydrogen summed into heavy atom. (one might, I suppose, mistakenly think that it is sum of all atomic charges -- a float approximately the total charge on the molecule)

Done; let me know if it is clear enough.

berquist:
poke

berquist:
Regressions are failing because older examples don't have Hirshfeld charges. Rather than skip all these and possibly bloat SP/SPun files with a bunch of different atomic charge types, maybe we should have a new unit test type.

shivupa (approved):

PR #1138, opened 2022-06-15
Turbomole MP2 Opt Fix

oliver-s-lee:
In some cases, such as MP2 optimisations run with ricc2, the list of mpenergies would contain the same energy twice for each step, resulting in mpenergies being twice as long as expected. This is a fix for this problem, along with a regression test.


berquist (approved):
Confirmed good with https://github.com/cclib/cclib-data/pull/161. Thanks!

PR #1141, opened 2022-07-04
Gaussian fixes and additions

jvalegre:
@berquist this is a summary of fixes/additions for the Gaussian parser.

1. Fixed issues:
Atom masses don't get duplicated in OPT + FREQ jobs
ONIOM energy location with while corrected
Natural charges from NBO in open-shell calculations don't get only the values from 1 spin state (i.e. beta e only)
ZPVE, enthalpy and free energy in eV

2. Addition of parameters (as included in the data.py file):
-- As properties --
**Molecular mass** as properties
**S2 before annhilation final value** as properties
**S2 after annhilation final value** as properties
**ONIOM final energy** as properties:energy
**Wyberg bond order matrix** from NBO as properties:NBO
**G4 energy** as properties:energy
**ONIOM final energy** as properties:energy
**TD-DFT final energy** as properties:energy
**NMR isotropic values** as properties:NMR
**NMR anisotropic values** as properties:NMR
**NMR eigenvalues** as properties:NMR
**Rotational T** as properties:rotational
**Rotational constants** as properties:rotational
**Symmetry number** as properties:rotational
**Full point group** as properties:rotational

-- As atoms --
**Natural spins** from NBO as atoms (override Mulliken values)

-- As optimization --
**S2 before annhilation during OPT** as optimization
**S2 after annhilation during OPT** as optimization
**ONIOM energies during OPT** as optimization:ONIOM
**TD-DFT energy during OPT** as optimization:TD
**Number of times the calc converged** as optimization

-- As metadata --
**Functional** as metadata
**Basis set** as metadata
**Grid type** as metadata
**Calculation type** as metadata
**Dispersion model** as metadata
**Solvation model** as metadata
**Keywords line** as metadata
**Processors used** as metadata
**Memory** used as metadata
**Calculation date** as metadata
**Program and version** as metadata


berquist:
@LawrenceSpear Is there anything specific that you are interested in? It may be worth a new, separate issue.

jvalegre:
Hi @LawrenceSpear @berquist  - just to catch up, did the PR pass all the tests and the metadata params will be implemented?

oliver-s-lee:
Hi @jvalegre, thanks for the PR and sorry for not getting to it sooner, there's some really cool stuff in here!

However, this is a bit of a monster (in terms of size) which makes following some of the changes difficult, especially as you touch so many different attributes. Also, I suspect some of the changes here may be duplicates of stuff that's since been added elsewhere since you opened this in July, but again a little difficult to say for sure.

Would you be able to split this PR into it's smaller constituents so we can process these one at a time? In my quick scan through I can identify at least the following sections:
```
metadata
spin-annihilation
rotational stuff
stationary point convergence
nmr
wiberg bond order
td energies
onion stuff
atomspins
energy stuff
```
Fortunately most of your changes are in self contained blocks, so splitting it out shouldn't be too difficult. I can lend a hand splitting it if you need, just let me know. For any new attributes, we'll also need example log files to test that the parsing is working correctly, but we can get to this once the PR is split.

berquist:
Hi @akalikadien, you're seeing this with the branch for this PR and not `master`, correct?

This PR is in a bit of limbo while we disentangle some unrelated changes.  Are there things in this PR that you need?

akalikadien:
Hi @berquist, I'm actually seeing this with master as well. I didn't need anything specifically from this PR, I was just about to open a PR and saw that someone was already working on fixing Gaussian parsing :) 
Should I make a seperate issue for this bug? I can then implement the fix in a separate PR.

berquist:
> Hi @berquist, I'm actually seeing this with master as well. I didn't need anything specifically from this PR, I was just about to open a PR and saw that someone was already working on fixing Gaussian parsing :) Should I make a seperate issue for this bug? I can then implement the fix in a separate PR.

Yes please, that would be very welcome.

akalikadien:
I actually figured it out already, so no additional issue needed. For those interested: when you do a single-point calculation with Gaussian (whether the system is open-shell or closed shell), you'll always extract natural charges from the first (and in the case of closed-shell systems, the only) NBO analysis table. 
When you combine optimization, frequency calculation and NBO calculation e.g. by using the keywords: `#p opt freq pop=nbo`, then there will be three NBO analysis tables where the first one is NBO analysis on the input structure, second one is the NBO analysis after opt and the third one after freq calculation. 

In cclib the implemented natural charge extraction will always get values from only the first table in a log file. Which is a widely applicable implementation. For my own specific case I adapted the gaussianparser's code to extract values from the last table. See code below:

```
  for line_index, line in enumerate(inputfile):
      if line.strip() == "Natural Population":
          if not hasattr(self.data, 'atomcharges'):
              # if the atomcharges attribute for the cclib data object does not exist, create it
              # very unlikely that this happens, but just in case
              self.atomcharges = {}
          line1 = inputfile[line_index + 1]
          line2 = inputfile[line_index + 2]
          # use line 1 and 2 to check if this is the correct table
          if line1.split()[0] == 'Natural' and line2.split()[2] == 'Charge':
              dashes = inputfile[line_index + 3]
              charges = []
              count = 3
              # read the charge table below the dashes for as long as there are atoms
              for i in range(len(self.elements)):
                  count += 1
                  nline = inputfile[line_index + count]
                  charges.append(float(nline.split()[2]))
              # each table replaces the previous one, so the last table is the one we want
              self.atomcharges["natural"] = charges
```

jvalegre:
hi @oliver-s-lee - I just started a PR with multiple commits, each tackling a separate piece of this PR. Please, take a look at PR #1269

PR #1144, opened 2022-07-05
Updated with current core devs

langner:
Noticed this was out of date when going through the 1.7.2 release process.

### DIFF
--- a/doc/sphinx/development.rst
+++ b/doc/sphinx/development.rst
@@ -161,18 +161,24 @@ Websites related to cclib
 Developers
 ==========
 
-Besides input from a number of people `listed in the repository`_, the following developers have contributed code to cclib (in alphabetical order):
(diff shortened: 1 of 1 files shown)


langner:
@mscho527 @amandadumi @ghutchis you were added here

ghutchis:
I'm flattered if you really think I'm a "core developer." I think a fairer description might be "occasional contributor" but I'll let the rest of the developers decide.

langner:
> I'm flattered if you really think I'm a "core developer." I think a fairer description might be "occasional contributor" but I'll let the rest of the developers decide.

I took the list from the list of owners+members so maybe "core devs" is not the best phrase. Feels more like "stewards" (also for me, since I don't contribute as much as I'd like). Suggestions?

berquist (approved):

shivupa (approved):

PR #1150, opened 2022-08-30
Fix for symmetry detection in Gaussian with ghost atoms

felixplasser:
Hello,

I just wanted to add a small fix to the Gaussian parser.

I am attaching the file that I am interested in, [neutral.log](https://github.com/cclib/cclib/files/9453835/neutral.log). This file could be parsed with cclib v1.7.1 but with v1.7.2 it tells me.

~~~~
[Gaussian neutral.log ERROR] Unexpectedly encountered end of logfile.
~~~~
Changing the line as shown below seems to fix this.

### DIFF
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -573,7 +573,7 @@ class Gaussian(logfileparser.Logfile):
                 line = next(inputfile)
 
         # Symmetry: point group
-        if line.strip() == "Symmetry turned off by external request.":
+        if "Symmetry turned off" in line:
             self.set_attribute('uses_symmetry', False)
         if "Full point group" in line:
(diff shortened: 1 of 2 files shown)


shivupa (approved):

shivupa:
Thanks Felix. I'll add this as a regression test and then this should be ready to go.

berquist (approved):
At least with current `master` most of the file was not parsed into attributes.

Also this reminds me that NMR parsing is currently only for ORCA.

felixplasser:
> At least with current `master` most of the file was not parsed into attributes.
> 
> Also this reminds me that NMR parsing is currently only for ORCA.

In this case I am actually just getting the geometry via cclib.

I am parsing the NMR data via my own code in TheoDORE.

PR #1152, opened 2022-09-11
Break apart atomcharges tests per type into Mulliken and Lowdin

berquist:
This also includes a fix for old versions of Psi4 where the charge header was inconsistent.

### DIFF
--- a/cclib/parser/psi4parser.py
+++ b/cclib/parser/psi4parser.py
@@ -48,7 +48,9 @@ class Psi4(logfileparser.Logfile):
         # Regular expression for matching the start of an atomic charges
         # section; the second format of the header (with square brackets) is
         # for old Psi4 versions.
-        self.re_atomic_charges_header = re.compile(r"^\s*(\w+) Charges(:?: \(a\.u\.\)| \[a\.u\.\]:)")
+        self.re_atomic_charges_header = re.compile(
+            r"^\s*(Mulliken|Lowdin|MBIS) Charges(:?: \(a\.u\.\)| \[a\.u\.\]:)"
+        )
 
     def after_parsing(self):
         super(Psi4, self).after_parsing()
--- a/cclib/parser/psi4parser.py
+++ b/cclib/parser/psi4parser.py
@@ -8,6 +8,7 @@
 """Parser for Psi4 output files."""
 
 from collections import namedtuple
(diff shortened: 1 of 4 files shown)


cclib/parser/psi4parser.py
```diff
@@ -44,6 +45,11 @@ def before_parsing(self):
         # with =>/<= rather than ==>/<==.
         self.subsection = None
 
+        # Regular expression for matching the start of an atomic charges
+        # section; the second format of the header (with square brackets) is
+        # for old Psi4 versions.
+        self.re_atomic_charges_header = re.compile(r"^\s*(\w+) Charges(:?: \(a\.u\.\)| \[a\.u\.\]:)")
```
langner, line 51: Do you think it's better to spell out which ones we expect here, as in `(Mulliken|Lowdin)` ?

test/data/testSP.py
```diff
@@ -68,11 +68,40 @@ def testatomnos(self):
     @skipForLogfile('Molpro/basicMolpro2006', "These tests were run a long time ago and since we don't have access to Molpro 2006 anymore, we can skip this test (it is tested in 2012)")
     @skipForParser('Turbomole','The parser is still being developed so we skip this test')
     def testatomcharges(self):
```
langner, line 70: Should this then have a Mulliken suffix?

langner (approved):

cclib/parser/psi4parser.py
berquist: According to https://psicode.org/psi4manual/master/oeprop.html there is only one other type (MBIS), so it is not bad to do `(Mulliken|Lowdin|MBIS)`.

test/data/testSP.py
berquist: This was intentional, to have `testatomcharges` be generic for all types and `testatomcharges_mulliken` and `testatomcharges_lowdin` to make sure Mulliken and Lowdin are present. Is there a better way of doing this? I can at least improve the docstring to something like `Are any present atomic charges consistent with natom?`.

amandadumi: The change in description helped this make sense to me. I see why both the generic and specific tests are needed.

shivupa (approved):

PR #1160, opened 2022-10-23
docs: update minimum recommended version of Python from 3.4 to 3.7

berquist:
When looking for the minimum recommended Python version in the docs I found a few outdated things. Our `INSTALL` was both outdated and duplicated from a section in Sphinx. Let me know if removing it is too extreme and we should just update it.


langner:
I'm confused by the failing coverage/ run-tests now, because they reference Shiv's branch. What's going on with that?

shivupa (approved):

PR #1161, opened 2022-10-25
Use new Docker image tag for Python 3.7

shivupa:
Test new py37 image

### DIFF
--- a/.github/workflows/cclib_pytest.yml
+++ b/.github/workflows/cclib_pytest.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


berquist (approved):
I know this PR was originally for testing the CI, but we'll probably switch to edge for all future Python versions anyway. Plus `py37` wasn't tagged and pushed for some reason, the metadata step must be wrong...

shivupa:
Yeah the metadata step documentation was confusing and there wasn't a good way to check the generated tag

berquist:
Oh, it's just `raw`: https://github.com/docker/metadata-action#typeraw

But maybe we don't actually want that, it's clearer by having the mix of edge + YYYYMMDD + sha.

Issue #1153, opened 2022-09-14
Gaussian parser: error on parsing output with MO coefficients


mariuszradon:
Gaussian parser (v 1.7.2) produces the following errors while parsing the attached Gaussian 16 output file (h2o_1.log):

logfileparser.py:321, in Logfile.parse(self, progress, fupdate, cupdate)
    316 # This call should check if the line begins a section of extracted data.
    317 # If it does, it parses some lines and sets the relevant attributes (to self).
    318 # Any attributes can be freely set and used across calls, however only those
    319 #   in data._attrlist will be moved to final data object that is returned.
    320 try:
--> 321     self.extract(inputfile, line)
    322 except StopIteration:
    323     self.logger.error("Unexpectedly encountered end of logfile.")

gaussianparser.py:1745, in Gaussian.extract(self, inputfile, line)
   1743         self.mocoeffs[1][base:base + len(part) // 10, i] = temp
   1744     else:
-> 1745         mocoeffs[0][base:base + len(part) // 10, i] = temp
   1747 if base == 0 and not beta:  # Do the last update of atombasis
   1748     self.atombasis.append(atombasis)

ValueError: could not broadcast input array from shape (5,) into shape (4,)

The second provided output file h2o_2.log is parsed properly. The only difference it that the second output was generated without "pop=regular GFInput" in the route section; therefore, it does not contain the MO coefficients (LCAO matrix).

[h2o_1.log](https://github.com/cclib/cclib/files/9565745/h2o_1.log)
[h2o_2.log](https://github.com/cclib/cclib/files/9565748/h2o_2.log)

The expected behavior is that both output files should be parsed properly. The data resultsing from the first file should additionally contain the "mocoeffs" attribute (LCAO matrix).

berquist:
Hi Mariusz, thanks for the report. It looks like the problem has something to do with `pop=regular` rather than `full` or `none`, even though there is at least one other `opt pop=regular` calculation in our test set. Until we fix this, try running with `pop=full` if you need MO coefficients to be extracted.

oliver-s-lee:
Hi folks, I've also come across this error. It's caused by the logic that detects whether a calculation was run with pop=full or not:
```
if base == 0 and int(colmNames.split()[0]) != 1:
                    # Implies that this is a POP=REGULAR calculation
                    # and so, only aonames (not mocoeffs) will be extracted
                    self.popregular = True
```
[...]

PR #1162, opened 2022-11-03
Gauss mocoeff

oliver-s-lee:
Fix for #1153


shivupa:
Generally looks like a fix to me
Tests are failing because
`data/regression/Gaussian/Gaussian03/chn1.log:487`:
```
      DENSITY MATRIX.
```

where as the case changed in newer gaussian versions https://github.com/cclib/cclib-data/pull/163/files
`issue1153.log:500`
```
     Density Matrix:
```

oliver-s-lee:
Hmmm, that's annoying. Obvious fix is to just check for either?

shivupa:
Yup that should be good! Thank you!

shivupa (approved):

berquist (approved):

PR #1164, opened 2022-11-07
Gaussian: increase specificity of Moller-Plesset checks

oliver-s-lee:
Increased specificity of MP checks in Gaussian parser.


shivupa:
In gaussian03 and older it seems like this space didn't exist.
data/regression/Gaussian/Gaussian03/water_mp5.log
```
E3=       -0.10518804E-01        EUMP3=      -0.75012800931E+02
```

That is causing the failure

shivupa:
Also in test/regression.py could a test for the file you added to cclib-data be added. Something along the lines of
```
def testGaussian_Gaussian16_Ethane_mp5_log(logfile):
+    """G16 error for parsing mp5 energies """
+    assert hasattr(logfile.data, "mpenergies")
+    test the mp5 energy value
```

berquist (approved):

PR #1165, opened 2022-11-16
Possible infinite loop?

shivupa:


cclib/method/ddec.py
```diff
@@ -736,6 +740,8 @@ def _converge_phi(self, phiA, superscript, atomi):
             candidates_phi.append(phiA)
             candidates_bigphi.append(bigphiA)
 
+        if curr_it == it_max:
```
langner, line 743: If you do this here, you don't need to tweak the loop definition.

shivupa:
😃

test/data/skip.py
```diff
@@ -31,3 +33,27 @@ def testwrapper(self, *args, **kwargs):
                 testfunc(self, *args, **kwargs)
         return testwrapper
     return testdecorator
+
+def skipForVersion(version, msg):
```
amandadumi, line 37: the pytest built in function may just let us have less code to maintain? https://docs.pytest.org/en/7.1.x/how-to/skipping.html

For our purposes:
`@pytest.mark.skipif(sys.version_info < (3, 8), reason="his test doesn't converge with newer psi4 versions >3.8")`

amandadumi (changes requested):
Other than one suggestion, this looks good!

test/data/skip.py
shivupa: Fixed!

shivupa:
Ready for review

PR #1166, opened 2022-11-16
Added excited state symmetry to etsyms param for Orca TD-DFT (previously only contained mult)

oliver-s-lee:


berquist (changes requested):
The code itself looks good, but since we don't have a test for the actual contents of the parsed `etsyms`, could you add one? It's ok if it's ORCA-specific for now until all the parsers are standardized.

oliver-s-lee:
Yeah can do, should be easy to cover Gaussian and Orca at least as they have the same format (multiplicity-symmetry eg `Singlet-Ag`).

On the topic, I think it makes sense at some point to split etsyms into separate multiplicity and symmetry parameters? They're really quite different obviously, and often you're only interested in one or the other. Would also help standardise the etsyms format (etsyms could be a numeric field for example).

berquist:
> On the topic, I think it makes sense at some point to split etsyms into separate multiplicity and symmetry parameters? They're really quite different obviously, and often you're only interested in one or the other. Would also help standardise the etsyms format (etsyms could be a numeric field for example).

Yes, we can have `etmults` or similar and keep `etsyms` just for irrep labels. If we convert `etsyms` to a numeric field then we'll also need the label for all irreps in the parsed point group. I'm not sure what `etmults` would be, since some programs will print the excited state <S^2> for unrestricted calculations.

oliver-s-lee:
Sorry I meant the hypothetical etmults would be numeric, 1 for singlet, 2 for doublet, 3 for triplet and so on. Whether fractional 'multiplicities' should be allowed to support <S^2> for unrestricted I'm not sure. I know Gaussian at least does this so it's not unreasonable to come across, and to me it seems better to support parsing something rather than returning nothing?

In the world where we have etmults I think it would make most sense to have etsyms just contain symmetries (`A`, `Ag` etc) rather than both multiplicity and symmetry. That way the irrep could be obtained by just combining the multiplicity and symmetry (eg `irrep = data.etmults[0] + "-" + data.etsyms[0]`). Perhaps another property (`etirreps`?) could be added that contains both (eg, `Singlet-Ag`) if it's convenient, although this property would be technically redundant.

oliver-s-lee:
Added some initial tests for Gaussian, Orca and Turbomole. They're pretty specific for now because etsyms is not standardised across parsers and some calcs are missing symmetry.

shivupa:
It seems like the errors are coming from regression tests that don't have symmetry.
If you modify:
https://github.com/cclib/cclib/blob/d96e9b38aa980999a1e8dec384537640ca83d3e8/test/regression.py#L3188
to not have the symmetries then we should be ok!

oliver-s-lee:
Thanks for the hint re. OrcaTDDFTTest_pre1085, these should all be fixed now?

The regression file ORCA4.1/porphine.out is particularly useful as it looks like its our only example of an optimised excited state calculation (which we couldn't parse properly). This is a good reminder that we should probably have at least one of these tricky cases in the main testing set...

berquist:
> This is a good reminder that we should probably have at least one of these tricky cases in the main testing set

Do you mean like formaldehyde, where it's pretty easy to see the order of states change?

cclib/parser/gaussianparser.py
```diff
@@ -1458,7 +1458,9 @@ def extract(self, inputfile, line):
             groups = p.search(line).groups()
             self.etenergies.append(utils.convertor(utils.float(groups[1]), "eV", "wavenumber"))
             self.etoscs.append(utils.float(line.split("f=")[-1].split()[0]))
-            self.etsyms.append(groups[0].strip())
+            # Fix Gaussian's weird capitalisation.
+            mult, symm = groups[0].strip().split("-")
+            self.etsyms.append("{}-{}".format(mult, self.normalisesym(symm)))
```
berquist, line 1463: Can you switch this to use `append_attribute`?

cclib/parser/orcaparser.py
```diff
@@ -78,6 +78,20 @@ def after_parsing(self):
                     )
                     break
                 self.scfenergies[i] += dispersionenergy
+        
+        # ORCA prints singlet and triplet excited states separately, so the energies are out of order.
+        if hasattr(self, "etenergies"):
+            # Sort them properly.
+            prop_names = ("etenergies", "etsyms", "etoscs", "etsecs", "etrotats")
+            #zipped_props = [zipped_state for zipped_state in sorted(zip_longest(getattr(self, 'etenergies', []), getattr(self, 'etsyms', []), getattr(self, 'etoscs', []), getattr(self, 'etsecs', []), getattr(self, 'etrotats', [])), key = lambda state: state[0])]
```
berquist, line 86: Can you clean this up a bit?

cclib/parser/orcaparser.py
```diff
@@ -78,6 +78,20 @@ def after_parsing(self):
...
+            zipped_props = [zipped_state for zipped_state in sorted(zip_longest(*[getattr(self, prop_name, []) for prop_name in prop_names]), key = lambda state: state[0])]
```
berquist, line 87: and break this across multiple lines?

cclib/parser/orcaparser.py
```diff
@@ -78,6 +78,20 @@ def after_parsing(self):
                     )
                     break
                 self.scfenergies[i] += dispersionenergy
+        
+        # ORCA prints singlet and triplet excited states separately, so the energies are out of order.
```
berquist, line 82: AFAIK we don't make a promise that energies (states) are in complete energetic order. Should we?

berquist (changes requested):
I am happy with this as a first pass, some things just need tidying up.

cclib/parser/orcaparser.py
oliver-s-lee: This is a fair point. In my mind I think it makes sense to re-order and it's not particularly difficult to do so. In particular, I think most programs most of the time will print excited states in increasing energy order, so to the user it will often seem like the order is being enforced. This is until they come across a strange case like this one where suddenly there's a drop in energy halfway through their excited states, which could easily look like a cclib bug (I assumed it was). Not a problem tho can remove if not?

oliver-s-lee:
There's actually a better version of the excited state reordering code in orca-eom which I've now brought back to this PR. The old code could hide parsing bugs by accidentally truncating attributes.

cclib/parser/orcaparser.py
berquist: Nope, I agree. It exposes an implementation detail that singlets and triplets are computed separately and leaves the user to suffer when they tend to care more about the energy as you say.

Issue #407, opened 2017-08-02
Using standardized inputs for testing


gaursagar:
A suggestion to make sure test inputs for all programs are standarized on coordinates, basis sets, and functionals came up during a discussion with @ATenderholt and @langner. 

I have access to versions of
```
ADF
Firefly
GAMESS(US)
Gaussian
Gaussian
Molpro
Mopac
NWChem
Orca
Psi4
```
and `Dalton` is going LGPL in mid-August 2017.

berquist:
ADF! That's great. Now we're only missing GAMESS-UK and Jaguar.

gaursagar:
ADF licence has probably expired recently :/  .Its still listed under list of programs available in the University clusters, but didn't load. Sorry, to ruin your excitement.

xymaxim:
If you need to prepare and run jobs in ADF, I could make a contribution. So feel free to ask for my help.

berquist:
Thanks, that's good to know. I think I wrote what I have at some point, but for completeness:

```
DALTON
GAMESS(US)
Molpro
Mopac
NWChem
ORCA
Psi4
Q-Chem
```

gaursagar:
Not sure what would be a good way to coordinate. For now I am using Google sheets to track things.

(https://docs.google.com/spreadsheets/d/1mA360PbAQKSWoM9nNp5oJntb9MYouDBxGFF8l5_43vE/edit?usp=sharing)

langner:
Which input are we using as canonical?

gaursagar:
Inputs used in `Datlon`, because the input coordinates there are quite common in other programs.

For C atom based inputs
`C    0.0  0.0  0.0`

dvb:
```
    C    -1.4152533224   0.2302217854   0.0000000000
    C     1.4152533224  -0.2302217854   0.0000000000
    C    -0.4951331558   1.3144608674   0.0000000000
    C     0.4951331558  -1.3144608674   0.0000000000
    C     0.8894090436   1.0909493743   0.0000000000
    C    -0.8894090436  -1.0909493743   0.0000000000
    H    -0.8795511985   2.3437343748   0.0000000000
    H     0.8795511985  -2.3437343748   0.0000000000
    H     1.5779041557   1.9450061275   0.0000000000
    H    -1.5779041557  -1.9450061275   0.0000000000
    C     2.8845844962  -0.5210893778   0.0000000000
    C    -2.8845844962   0.5210893778   0.0000000000
[...]

PR #1168, opened 2022-11-17
Orca Post-HF Excited States

oliver-s-lee:
Added parsing for post-HF (ADC(2) and various EOM-CCSD) excited states methods with ORCA.
Please note this request depends on #1166 

Also included are two convenience functions in logfileparser that may/may not wont to be kept.

No parsing/testing for unrestricted calcs or triplet excited states yet (except for STEOM-DPLNO which makes triplet/singlet obvious)


cclib/parser/orcaparser.py
```diff
@@ -1384,6 +1388,149 @@ def energy_intensity(line):
...
+                elif line.strip() == "":
+                    line = self.next_filled_line(inputfile)
+                
+                if "Percentage Active Character " in line:
+                    # Data currently not parsed. Just skip.
+                    line = self.next_filled_line(inputfile)
+                    
+                if "Warning:: the state may have not converged with respect to active space" in line:
+                    # Skip this line and the next (which both contain warnings).
+                    # TODO: Capture this warning?
```
berquist, line 1508: You can use `self.logger.warning` (the warning method on our instance of https://docs.python.org/3/library/logging.html) to report it.

test/data/testTD.py
```diff
@@ -366,6 +370,26 @@ class TurbomoleTDCC2TripTest(GenericTDTest):
...
         self.assertEqual(len(self.data.etenergies), self.number)
+
+class OrcaETPostHFTest(GenericTDTest):
+    """Tests for post-HF excited states with ORCA."""
+    
+    number = 2
+    symmetries = ["Singlet", "Singlet"]
+    # Not sure why this value != 1 for these methods?
+    # Perhaps remaining contributions were too small to print?
+    sumofsec = 0.43
```
berquist, line 381: I tried increasing `%mdci printlevel` to 4 and it didn't change this.

data/ORCA/basicORCA5.0/dvb_adc2.log
```diff
@@ -0,0 +1,1681 @@
...
+
+WARNING: MDCI localization with Augmented Hessian Foster-Boys
+  ===> : Switching off randomization!
+
+INFO   : the flag for use of the SHARK integral package has been found!
+
+================================================================================
+                                       INPUT FILE
+================================================================================
+NAME = dvb.inp
```
berquist, line 188: Can you also add the corresponding input file for each output file?

data/ORCA/basicORCA5.0/dvb_adc2.log
```diff
@@ -0,0 +1,1681 @@
...
+|  9> %mp2
+| 10>     end
+| 11> %mdci
+| 12>     NRoots      2
+| 13>     DoTriplet   false
+| 14>     STEOMSOC    false
+| 15>     DoDbfilter  false
+| 16>     end
+| 17> *xyz 0 1
+| 18> C            -1.41525000     0.23022000     0.00000000
```
berquist, line 206: Can you make sure these coordinates are consistent with the other input files? (It is possible they aren't even consistent among themselves already...)

data/ORCA/basicORCA5.0/dvb_adc2.log
```diff
@@ -0,0 +1,1681 @@
...
+WARNING: MDCI localization with Augmented Hessian Foster-Boys
+  ===> : Switching off randomization!
+
+INFO   : the flag for use of the SHARK integral package has been found!
+
+================================================================================
+                                       INPUT FILE
+================================================================================
+NAME = dvb.inp
+|  1> ! ADC2 NOCOSX cc-pVDZ/C cc-pVDZ TightSCF
```
berquist, line 189: Can you switch the primary basis to STO-3G? For the aux C basis, I am partial to how Psi4 auto pairs def2-SV(P) with STO-3G.

cclib/parser/logfileparser.py
```diff
@@ -530,6 +538,14 @@ def skip_lines(self, inputfile, sequence):
             lines.append(line)
 
         return lines
+    
+    def next_filled_line(self, inputfile):
```
berquist, line 542: This can be a `staticmethod` since `self` isn't used.

data/ORCA/basicORCA5.0/dvb_adc2.log
```diff
@@ -0,0 +1,1681 @@
...
+|  2> %Pal NProcs 8 end
```
berquist, line 190: Once you reduce the basis, you can probably nuke this...it took me less than a minute to run this on a single core.

berquist (changes requested):
It's good that we are starting to get serious coverage of post-HF excited states in the different parsers. I agree with putting outputs as unit tests and not regressions.

For the new utility methods, I don't see any future problems arising due to adding them. Others should chime in.

My only requests are regarding the new output files. I know that our unit tests are not very consistent, but I think sticking to STO-3G is the right idea save for 1-2 dedicated tests where the basis is different. (If you're interested in our standardization woes, you can see https://github.com/cclib/cclib/issues/407.)

data/ORCA/basicORCA5.0/dvb_adc2.log
oliver-s-lee: Sure, can do. For the aux basis, from the ORCA manual I don't think def2-SV(P)/C is included by default but I can do you def2-SVP/C? If it's an issue I think there's a version of def2-SV(P)/C on the BSE so I could try with that, but getting ORCA to work properly with BSE is a pain, so for ease of use I'd recommend def2-SVP/C.

berquist: def2-SVP/C is fine.

cclib/parser/orcaparser.py
```diff
@@ -1384,6 +1388,149 @@ def energy_intensity(line):
...
+                        end = coeff_split[2]
+                        contrib = "".join(coeff_split[4:])[1:-1]
+                    
+                    try:
+                        # TODO: Unrestricted?
+                        sec.append([(int(start), 0), (int(end), 0), float(contrib)])
+                        
+                    except ValueError:
+                        # Sometimes we come across lines like:
+                        # 0.690372    69 -> x
```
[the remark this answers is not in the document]
oliver-s-lee: I'm really not sure. This notation is shown in two places in the manual (one for high level excited states like we're seeing here and the second for ionisation). I suspect this notation was originally for ionisation-like processes and has since been reused in a different context.

These results are the pre-calculations before the final excited states are calculated (presumably at some lower-level of theory). My best guess is that the types of excited states being calculated at this stage either don't have the same concept of starting and final orbital transitions, and/or that the orbitals being transitioned to or from do not map well to the ground state orbitals.

data/ORCA/basicORCA5.0/dvb_adc2.log
oliver-s-lee: They match with the dvb_td.inp for ORCA 5.0 and 4.2

oliver-s-lee:
@berquist no rush, but gentle push that this is ready for review whenever you get time

berquist (approved):
I'm sorry I missed the rebase and force push.  This all looks good now.

PR #1169, opened 2023-01-05
Fixed another bug parsing mocoefficients with alpha/beta orbitals and pop=reg

oliver-s-lee:


cclib/parser/gaussianparser.py
```diff
@@ -1721,7 +1721,7 @@ def extract(self, inputfile, line):
 
                 colmNames = next(inputfile)
                 
-                if "Density Matrix:" in colmNames or "DENSITY MATRIX." in colmNames:
+                if "Density Matrix:" in colmNames or "DENSITY MATRIX." in colmNames or "Beta Molecular Orbital Coefficients" in colmNames:
```
berquist, line 1724: Can you do an `if any(name in colmNames for name in ["Density Matrix:", ...`? Sometimes when this happens we stick the variable with possible headers on the class instance in `before_parsing`. It gets cleaned up when attributes are filtered out and might make the parser a little faster since it doesn't need to reevaluate one list of strings every time the condition is checked. (See https://github.com/cclib/cclib/blob/49b0cd6056de292b5cb1eb4d66da96e540e94493/cclib/parser/qchemparser.py#L95.)

berquist (changes requested):

berquist:
Also, is there an output file this fails on?

oliver-s-lee:
There's actually not in the normal testing set, unfortunately all the unrestricted examples were run with pop=full so this bug was missed. I've added one at https://github.com/cclib/cclib-data/pull/167

berquist (approved):

PR #1172, opened 2023-01-11
Orca etoscs with SOC

oliver-s-lee:
Spin-orbit coupling calculations in Orca generate a number of SOC-corrected absorption spectra. The current version of cclib parses all of these and uses the corresponding values to overwrite values of etenergies and etoscs. This is probably undesirable, as in most cases it is the bare electronic transitions that are of interest, rather than the spin-mixed ones. 

Even if this is the desired behaviour, it is currently incompatible with the values of etsyms that are parsed, which are only for the un-mixed states (there are also more spin-mixed states than etsyms).

This PR changes this behaviour so only the etoscs corresponding to the unmixed states are parsed (but please note that this is technically a loss in functionality if it really was the spin-mixed states you were after). The test for the ORCA ROCIS file has been changed to reflect this.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -1379,7 +1379,8 @@ States  Energy Wavelength    D2        m2        Q2         D2+m2+Q2       D2/TO
                     else:
                         # Because these energies are new, we do not know if they correspond to the same level of theory
                         # as the previously parsed etsyms etc.
-                        self.logger.warning("New excited state energies encountered in spectrum section, resetting excited state attributes")
+                        self.logger.warning(
+                            "New excited state energies encountered in spectrum section, resetting excited state attributes")
                         
                         for attr in ("etenergies", "etsyms", "etoscs", "etsecs", "etrotats"):
                             if hasattr(self, attr):
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -1347,14 +1347,48 @@ States  Energy Wavelength    D2        m2        Q2         D2+m2+Q2       D2/TO
                 # Also, there are twice as many SOC states as true spin states, so half of the etenergies wouldn't 
                 # have a symmetry in etsyms at all...
                 #
-                # Only keep etosc given by the normal electric TDM.
-                if name == "ABSORPTION SPECTRUM VIA TRANSITION ELECTRIC DIPOLE MOMENTS":
-                    # There's no point overwriting the etenergies we've already parsed,
-                    # especially because the previously parsed values are probably more accurate.
-                    # We could do a check to make sure the energies match, but we will fall foul
-                    # of rounding errors.
+                # Don't parse from SOC sections.
+                # ROCIS COMBINED is combination of SOC and ROCIS (we still parse the normal ROCIS section).
+                if not any([soc_header in name for soc_header in ["SPIN ORBIT CORRECTED", "SOC CORRECTED", "ROCIS COMBINED"]]):
+                    # We need to be careful about how we parse etenergies from these spectrum sections.
+                    # First, and in most cases, energies printed here will be the same as those printed in 
+                    # previous sections, except to fewer decimal places. It would be disadvantageous to 
+                    # lose this extra precision that's already been parsed.
+                    # Secondly, it is challenging to determine exactly whether the energies printed here have already
+                    # been parsed, due to this aforementioned rounding.
+                    # Thirdly, some methods (ROCIS, CASSCF, SOC to name a few) may only print their final excited state
+                    # energies in this spectrum section, in which case the energies will not match those previously parsed
+                    # (which will be from lower levels of theory that we're not interested in). This means we cannot simply
+                    # ignore the energies printed. Also, in this case we must decide whether to discard other previously
+                    # parsed etdata (etsyms, etsecs etc).
+                    
+                    # If we have no previously parsed etnergies, there's nothing to worry about.
                     if not hasattr(self, "etenergies"):
                         self.set_attribute("etenergies", etenergies)
+                    
+                    # Determine if these energies are ~same as those previously parsed.
+                    elif len(etenergies) == len(self.etenergies) and all([self.etenergies[index] == etenergy for index, etenergy in enumerate(etenergies)]):
+                        pass
+                    
+                    # New energies.
+                    else:
+                        # Because these energies are new, we do not know if they correspond to the same level of theory
+                        # as the previously parsed etsyms etc.
+                        self.logger.warning("New excited state energies encountered in spectrum section, resetting excited state attributes")
+                        
+                        for attr in ("etenergies", "etsyms", "etoscs", "etsecs", "etrotats"):
+                            if hasattr(self, attr):
+                                delattr(self, attr)
+                                
+                        self.set_attribute("etenergies", etenergies)
+                    
+#                     # There's no point overwriting the etenergies we've already parsed,
+#                     # especially because the previously parsed values normally have greater precision.
+#                     # We could do a check to make sure the energies match, but we will fall foul
+#                     # of rounding errors.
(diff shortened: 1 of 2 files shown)


oliver-s-lee:
There is a regression test file https://github.com/cclib/cclib-data/pull/169, note that this seems to actually pass in the current cclib version, but only because the check to make sure etsyms and etenergies hasn't bee introduced yet.

cclib/parser/orcaparser.py
```diff
@@ -1339,9 +1339,61 @@ def energy_intensity(line):
...
-
-                self.set_attribute('etenergies', etenergies)
-                self.set_attribute('etoscs', etoscs)
+                
+                # Some of these sections contain data that we probably do not want to be populating etenergies
+                # and/or etoscs with.  For example, the SOC corrected spectra are for mixed singlet/triplet states,
+                # so they do not correspond to the symmetries given in etsyms, and the energy values given are
+                # probably not what the user would expect to find in etenergies anyway?
+                # Also, there are twice as many SOC states as true spin states, so half of the etenergies wouldn't 
+                # have a symmetry in etsyms at all...
```
berquist, line 1348: Another technical limitation. If we end up wanting to support this, the question becomes one like [struct-of-arrays vs. array-of-structs](https://en.wikipedia.org/wiki/AoS_and_SoA). Do we have SOC as the outermost grouping, and energies/symmetries/multiplicities/etc. inside there, or those attributes as top-level groupings, and the SOC-corrected states inside each of those attributes? Or something in between?

We are leaning towards the latter. Both still require a unifying marker for "these are from the SOC-corrected states".

test/data/testTD.py
```diff
@@ -226,6 +226,7 @@ def testoscs(self):
 class OrcaROCISTest(GenericTDTest):
     """Customized test for ROCIS"""
     number = 57
```
berquist, line 228: Did you forget to delete this line?

cclib/parser/orcaparser.py
```diff
@@ -1339,9 +1339,61 @@ def energy_intensity(line):
...
+                    # Determine if these energies are same as those previously parsed.
+                    # May want to use a smarter comparison?
+                    elif len(etenergies) == len(self.etenergies) and all([self.etenergies[index] == etenergy for index, etenergy in enumerate(etenergies)]):
+                        pass
+                    
+                    # New energies.
+                    else:
+                        # Because these energies are new, we do not know if they correspond to the same level of theory
+                        # as the previously parsed etsyms etc.
+                        self.logger.warning("New excited state energies encountered in spectrum section, resetting excited state attributes")
```
berquist, line 1378: So you already know about the logger; like in the other PR, you can also use it to re-echo warnings directly printed in the output file if you think they're important enough.

cclib/parser/orcaparser.py
```diff
@@ -1339,9 +1339,61 @@ def energy_intensity(line):
...
+                    # Thirdly, SOC prints spin-mixed excited state spectra. This is interesting, but does not match the 
+                    # number of states or symmetry of data parsed in previous sections, so is not used to overwrite etenergies.
+                    
+                    # If we have no previously parsed etnergies, there's nothing to worry about.
+                    if not hasattr(self, "etenergies"):
+                        self.set_attribute("etenergies", etenergies)
+                    
+                    # Determine if these energies are same as those previously parsed.
+                    # May want to use a smarter comparison?
+                    elif len(etenergies) == len(self.etenergies) and all([self.etenergies[index] == etenergy for index, etenergy in enumerate(etenergies)]):
```
berquist, line 1371: Can you break this up over at least 2 lines?

cclib/parser/orcaparser.py
```diff
@@ -1339,9 +1339,61 @@ def energy_intensity(line):
...
+                        # as the previously parsed etsyms etc.
+                        self.logger.warning("New excited state energies encountered in spectrum section, resetting excited state attributes")
+                        
+                        for attr in ("etenergies", "etsyms", "etoscs", "etsecs", "etrotats"):
+                            if hasattr(self, attr):
+                                delattr(self, attr)
+                                
+                        self.set_attribute("etenergies", etenergies)
+                    
+#                     # There's no point overwriting the etenergies we've already parsed,
```
berquist, line 1386: I can understand wanting to leave in a comment like this, but can you clean it up a little? Something like maybe, mostly the conditional on 1390/1391 is confusing. I also needed to remind myself that `etenergies` is indeed in scope for the entire block of this PR.
```patch
@@ -1383,13 +1383,12 @@ States  Energy Wavelength    D2        m2        Q2         D2+m2+Q2       D2/TO
                                 
                         self.set_attribute("etenergies", etenergies)
                     
-#                     # There's no point overwriting the etenergies we've already parsed,
-#                     # especially because the previously parsed values normally have greater precision.
-#                     # We could do a check to make sure the energies match, but we will fall foul
-#                     # of rounding errors.
-#                     #if not hasattr(self, "etenergies"):
-#                     if True:
-#                         self.set_attribute("etenergies", etenergies)
+                    # There's no point overwriting the etenergies we've already parsed,
+                    # especially because the previously parsed values normally have greater precision.
+                    # We could do a check to make sure the energies match, but we will fall foul
+                    # of rounding errors.
+                    # if not hasattr(self, "etenergies"):
+                    #     self.set_attribute("etenergies", etenergies)
                     
                     self.set_attribute('etoscs', etoscs)
                 
```

berquist (changes requested):
This looks good so that most CIS/TDDFT output is now consistent with other programs , so no surprises. There are just a few code cleanup things left. I've also put some comments for longer-term changes we're thinking about if you're interested.

oliver-s-lee:
Cheers for comments, I'll go through these soon. Sorry for some of messiness, there was quite a lot of back and forth trying to figure this one out and a lot of the comments were for my own sanity :)

test/data/testTD.py
oliver-s-lee: 🙃

cclib/parser/orcaparser.py
```diff
@@ -1339,9 +1339,61 @@ def energy_intensity(line):
                     etoscs.append(float(intensity))
 
                     line = next(inputfile)
-
-                self.set_attribute('etenergies', etenergies)
-                self.set_attribute('etoscs', etoscs)
+                
+                # Some of these sections contain data that we probably do not want to be populating etenergies
+                # and/or etoscs with.  For example, the SOC corrected spectra are for mixed singlet/triplet states,
```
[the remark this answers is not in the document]
oliver-s-lee: Aye that makes sense. Something that I'm increasingly thinking would be a useful change is to split multiplicity and symmetry into separate attributes (as you say they are distinct properties). Technically this might not be backwards compatible though (if we're taking multiplicity OUT of the symmetry attribute), so might need to wait...

cclib/parser/orcaparser.py
oliver-s-lee: Sorry I can actually remove most of this comment (and obviously the test is just from debugging). The energies printed in cm-1 are actually identical in both sections, so less of a problem than I initially thought (:

In terms of scoping, yes I also find this section a bit busy and confusing. In the future I think it would benefit from a re-write/tidy-up, but this can probably wait until after it's decided where all these different excited state energies should live long-term

cclib/parser/orcaparser.py
oliver-s-lee: Hmm yeah I think I see what you're getting at. In terms of style I think the second is more in-line with the current cclib style, but I might be wrong. In terms of whether it's useful I'm not sure, I've only ever come across SOC used to describe the degree of mixing between states, not the actual mixed states themselves, but my view is rather narrow. I suppose if it's there and easy to parse it would be a shame not to eventually support...

berquist (approved):
Thanks!

PR #1173, opened 2023-01-12
Fix for crash when parsing multiple files with various parsers

oliver-s-lee:
Different parsers use different styles for obtaining the next line from the current input file, either they tend to use `line = next(inputfile)` or `line = inputfile.next()`. This is sadly a problem because not all `inputfile` objects support the `next()` method.

In particular, when parsing a single log file, `inputfile` is normally a `FileWrapper` object, while when parsing multiple log files, the `inputfile` is instead a `FileInput` object. `FileWrapper` support the `next()` method while `FileInput` does not, so parsers that do not typically expect multiple log files as input will crash when given so. For example the Gaussian parser when given any two files:

```
Attempting to read 246tCzPPC.log and 246tCzPPC.log
[Gaussian ['246tCzPPC.log', '246tCzPPC.log'] ERROR] Encountered error when parsing.
  File "/home/oliver/dev/cclib/cclib Src/cclib/parser/logfileparser.py", line 321, in parse
    self.extract(inputfile, line)
  File "/home/oliver/dev/cclib/cclib Src/cclib/parser/gaussianparser.py", line 241, in extract
    line = inputfile.next()
AttributeError: 'FileInput' object has no attribute 'next'

During handling of the above exception, another exception occurred:

  File "/home/oliver/.local/bin/ccget", line 8, in <module>
    sys.exit(ccget())
  File "/home/oliver/dev/cclib/cclib Src/cclib/scripts/ccget.py", line 178, in ccget
    data = ccread(filename, **kwargs)
  File "/home/oliver/dev/cclib/cclib Src/cclib/io/ccio.py", line 170, in ccread
    return log.parse()
  File "/home/oliver/dev/cclib/cclib Src/cclib/parser/logfileparser.py", line 327, in parse
    self.logger.error(f"Last line read: {inputfile.last_line}")
AttributeError: 'FileInput' object has no attribute 'last_line'
```

This PR changes all `inputfile.next()` instances to the more general `next(inputfile)`, and also correctly wraps `FileInput` objects with `FileWrapper` (this also gives access to the `last_line` attribute, and perhaps other stuff too). The change from .next() to next() is technically unnecessary but I thought it made sense to address at the same time to prevent problems in the future?

### DIFF
--- a/cclib/parser/adfparser.py
+++ b/cclib/parser/adfparser.py
@@ -166,7 +166,7 @@ class ADF(logfileparser.Logfile):
         # this with the previous block, if possible.
         if line[:6] == "Create":
             while line[:5] != "title" and "NO TITLE" not in line:
-                line = inputfile.next()
+                line = next(inputfile)
 
         if line[1:10] == "Symmetry:":
             info = line.split()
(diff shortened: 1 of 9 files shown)


berquist (approved):
This is also a good reminder that we can't parse multiple files with regressions (I think).

PR #1175, opened 2023-01-19
Fix CI and increase Python version usage

berquist:
Closes https://github.com/shivupa/cclib-ci/issues/12


berquist:
Wow, Python 3.10 really is faster (this was a selling point).

PR #1176, opened 2023-02-08
install pyquante2 with Python 3 fixes

berquist:
My fork has some fixes that let us use pyquante2 with all new Python versions.

iodata is required by tests and wasn't being installed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,8 @@ periodictable
 # bridges
 biopython==1.76
 ase>=3.21 ; python_version >= '3.5'
-git+https://github.com/berquist/pyquante2.git@py37 ; python_version > '3.0'
+git+https://github.com/berquist/pyquante2.git@py37
+git+https://github.com/theochem/iodata.git@1.0.0a2
 
 # For building the documentation
 sphinx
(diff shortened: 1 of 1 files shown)


langner:
Would we depend on your fork long term then?

berquist:
I'm not sure why it didn't occur to me that we should put the fork under our org. I think I've fixed the pyquante2+Cython install problems, fixed the pyquante2 CI, and set up our fork properly.

berquist:
For some reason pip (in the RTD build) is trying to install the Git dependencies (pyquante2) before *its* dependencies (Cython). We need Cython in our `requirements.txt` because the pyquante2 `setup.py` requires Cython to run.

berquist:
> For some reason pip (in the RTD build) is trying to install the Git dependencies (pyquante2) before _its_ dependencies (Cython). We need Cython in our `requirements.txt` because the pyquante2 `setup.py` requires Cython to run.

Need a workaround for this behavior: https://pip.pypa.io/en/stable/cli/pip_install/#argument-handling

shivupa (approved):
Sorry about the delay Thanks LGTM. Thanks for putting the fork under cclib!

PR #1178, opened 2023-03-04
use 'is not None' instead of '!= None'

### LINKED ISSUE #991: Use the `is` operator instead of `!=`
Use the `is` operator instead of `!=` [here](https://github.com/cclib/cclib/blob/master/test/bridge/testhorton.py#L183) to compare to None, because `!=`  calls __eq__ which is slower and error-prone.

berquist:
Closes #991

### DIFF
--- a/test/bridge/testhorton.py
+++ b/test/bridge/testhorton.py
@@ -62,8 +62,8 @@ class HortonTest(unittest.TestCase):
             if (
                 hasattr(self.iodat, attr)
                 and hasattr(hortonequiv, attr)
-                and getattr(self.iodat, attr) != None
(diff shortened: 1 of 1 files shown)


langner (approved):

berquist:
Test failure is because I merged https://github.com/cclib/cclib-data/pull/169 before https://github.com/cclib/cclib/pull/1172.

langner:
> Test failure is because I merged [cclib/cclib-data#169](https://github.com/cclib/cclib-data/pull/169) before #1172.

Cool, is there something blocking #1172 now?

berquist:
I’d like him to clean up some things, but it’s minor and black would get
most of it anyway.

PR #1179, opened 2023-03-04
Add type annotations to all functions and methods

berquist:
Not code style exactly, but I see it as part of #1122.


cclib/bridge/cclib2psi4.py
```diff
@@ -7,19 +7,21 @@
...
+
 from cclib.parser.utils import find_package
 
 _found_psi4 = find_package("psi4")
 if _found_psi4:
     from psi4.core import Molecule
 
 
-def _check_psi4(found_psi4):
+def _check_psi4(found_psi4: bool) -> None:
```
langner, line 19: Do you think it's prudent to annotate internal functions? Something I normally don't do.

berquist:
I always do, since they may be reused many times, the body won’t get
checked and f the signature isn’t annotated, and I’ve been bitten by this
before.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
...
+    hooks:
+      - id: check-yaml
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: "v1.0.1"
+    hooks:
+      - id: mypy
+        additional_dependencies:
+          - numpy
+          - packaging
+          - pandas-stubs
```
langner, line 15: Not obvious to me why pandas is a dep.

berquist: https://github.com/cclib/cclib/blob/35d72b0c20696778b5396842e20641541d2e1c7d/cclib/io/ccio.py#L496

cclib/bridge/cclib2psi4.py
berquist: This comment should have been a response here: https://github.com/cclib/cclib/pull/1179#issuecomment-1454916110:
> I always do, since they may be reused many times, the body won’t get
checked if the signature isn’t annotated, and I’ve been bitten by this
before.

berquist:
I think what I'll do is convert this from draft to PR once CI is fixed and leave everything as-is, so at least once more round of adding annotations, fixing annotations (wrong ones, fixing `np.ndarray`, ...), and cleaning up mypy settings is required before adding this to CI. I'll also check pyright without mypy. Pyre probably isn't important...

.pre-commit-config.yaml
berquist: I should say exactly what this file is. It's technically not necessary for mypy to run.[^1] `.pre-commit-config.yaml` is a configuration file for https://pre-commit.com/, which is two things:
1. A tool that runs linters and fixers from the command line. You do *not* need to have it as a pre-commit hook. Just doing https://pre-commit.com/#4-optional-run-against-all-the-files is enough.
2. A free CI service that runs alongside GH workflows, ReadTheDocs, etc.

The benefit to having 1 is that everyone has a unified way of running all these tools that's independent from CI, specifically opaque GitHub Actions. You can have 1 without 2. 2 would mean that you get the benefits of the hook (clean repository) without forcing contributors to install the hook.

The reason [`pandas-stubs`](https://github.com/pandas-dev/pandas-stubs) appears in `additional_dependencies` is because of how the tool works. For Python-based tools, it creates a fresh venv to run that tool in, and by default it only installs that tool. This means that mypy will not have access to any 3rd-party type information unless the dependencies or packages with only type stubs (for dependencies that don't ship them together) are also installed.

[^1]: For mypy configuration, you only need `pyproject.toml` or `mypy.ini`. I chose the former since everything is moving toward it anyway.

shivupa (approved):

test/test_data.py
```diff
@@ -295,7 +296,7 @@ def visualtests(self, stream=sys.stdout):
         )
 
 
-def test_all(parsers, modules, terse, silent, loglevel, summary, visual_tests):
+def test_all(parsers, modules, terse: bool, silent: bool, loglevel: int, summary: bool, visual_tests: bool) -> None:
```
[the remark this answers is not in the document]
berquist: I think `parsers` is `Mapping[str, cclib.parser.logfileparser.Logfile]` and `modules` is `Mapping[str, importlib.???.module]`.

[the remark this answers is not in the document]
shivupa: i think not critical. LGTM

Issue #648, opened 2018-11-10
pytest: individual unit tests are not displayed


berquist:
From https://github.com/cclib/cclib/issues/455#issuecomment-409720496 and https://github.com/cclib/cclib/pull/647, when using pytest,

> Output from `test_all` and the visual tests isn't displayed.

PR #1181, opened 2023-03-10
Make all regressions individual tests under pytest

### LINKED ISSUE #649: pytest: treat each regression file as its own test
From https://github.com/cclib/cclib/issues/455#issuecomment-409720496 and https://github.com/cclib/cclib/pull/647, when using pytest,

> Changing `main` to `test_regressions` allows pytest to automatically run regressions, but the current setup treats it as one big test. It would be nice if each regression file was its own individual test.

berquist:
Fixes #649

Once this is done, I don't think there will be the ability to go back to plain unittest.


berquist:
Right now all of the standalone `def test...` functions in `regression.py` are detected as individual tests (`python -m pytest -v -s test/regression.py`). The `*Test*` classes for unit tests are being picked up too, and they need parameterization. That probably needs to be solved simultaneously. Since this isn't enabled for CI due to the `-k` regex filter https://github.com/cclib/cclib/blob/3cea5df4f90e462625b3ff0381d610e07836cdc4/.github/scripts/run_pytest.bash#L12 all of this supporting code may need to go in first, then #648 and #649 get solved together.

shivupa:
I am not sure if I am doing something wrong but if I try
`python -m pytest -v -s test/regression.py`
I get a ton of failures like
```
FAILED test/regression.py::PsiHFSPTest_noatommasses::testscfenergy - AttributeError: 'PsiHFSPTest_noatommasses' object has no attribute 'logfile'
FAILED test/regression.py::PsiHFSPTest_noatommasses::testscftargetdim - AttributeError: 'PsiHFSPTest_noatommasses' object has no attribute 'logfile'
FAILED test/regression.py::PsiHFSPTest_noatommasses::testscftargets - AttributeError: 'PsiHFSPTest_noatommasses' object has no attribute 'logfile'
FAILED test/regression.py::PsiHFSPTest_noatommasses::testscfvaluetype - AttributeError: 'PsiHFSPTest_noatommasses' object has no attribute 'logfile'
FAILED test/regression.py::PsiHFSPTest_noatommasses::testsymlabels - AttributeError: 'PsiHFSPTest_noatommasses' object has no attribute 'logfile'
FAILED test/regression.py::PsiHFSPTest_noatommasses::testtypemoenergies - AttributeError: 'PsiHFSPTest_noatommasses' object has no attribute 'data'
```

berquist:
No, you aren't doing anything wrong. I should have been clearer, all tests that are like `class *Test*(...)` need at least some fancy fixtures in `conftest.py` in order to dynamically inject `logfile` and `data` into them. For the unit tests (#648) they need this plus parameterization over the contents of `testdata`. This PR only get the bare `def test_*(...)` functions working with fixtures.

The test classes need additional work regardless of whether or not I go with this approach of trying to not modify any test code method signatures or if they get rewritten more. (Conversion to bare asserts in #1184 is required no matter what.)

Issue #536, opened 2018-06-12
Fix line ending, but not in logfiles


langner:
See #371 and #531 for context.

berquist:
Do we also want to convert every file to UTF-8? Most are ASCII.

berquist:
Done in https://github.com/cclib/cclib/pull/1307/commits/7b2e8701dcb1a1a6c437919b185be78a35c3e2a5.  If there were a change of an entire file back from LF to CRLF our setup would not catch that, but we'll cross that bridge when we come to it.

PR #1185, opened 2023-03-13
Psi4 parser update

mskblackbelt:
The existing parser fails when parsing current (v1.7) Psi4 outputs. I've added new sample data and fixed the parser for the Convergence Check and IR Vibrations sections. I also changed the line endings to LF.

### DIFF
--- a/cclib/parser/psi4parser.py
+++ b/cclib/parser/psi4parser.py
@@ -788,7 +788,7 @@ class Psi4(logfileparser.Logfile):
         # This message signals a converged optimization, in which case we want
         # to append the index for this step to optdone, which should be equal
         # to the number of geovalues gathered so far.
-        if "Optimization is complete!" in line:
+        if "Final optimized geometry and variables:" in line:
 
             # This is a workaround for Psi4.0/sample_opt-irc-2.out;
             # IRC calculations currently aren't parsed properly for
--- a/cclib/parser/psi4parser.py
+++ b/cclib/parser/psi4parser.py
@@ -896,7 +896,7 @@ class Psi4(logfileparser.Logfile):
                 while line.strip():
 
                     tokens = line.split()
-                    if tokens[0] == "Magnitude":
+                    if tokens[0] in ("Magnitude", "Traceless"):
                         line = next(inputfile)
                         continue
                     value = float(tokens[-1])
@@ -914,7 +914,7 @@ class Psi4(logfileparser.Logfile):
                 line = next(inputfile)
 
             if not hasattr(self, 'moments'):
-                self.moments = moments
+                self.set_attribute("moments", moments)
             else:
                 for im, m in enumerate(moments):
                     if len(self.moments) <= im:
--- a/cclib/parser/psi4parser.py
+++ b/cclib/parser/psi4parser.py
@@ -812,9 +812,16 @@ class Psi4(logfileparser.Logfile):
         # Properties will be evaluated at   0.000000,   0.000000,   0.000000 [a0]
         #
         if "Properties will be evaluated at" in line.strip():
-            self.origin = numpy.array([float(x.strip(',')) for x in line.split()[-4:-1]])
-            assert line.split()[-1] in ["Bohr", "[a0]"]
-            self.origin = utils.convertor(self.origin, 'bohr', 'Angstrom')
+            tokens = line.split()
+            assert tokens[-1] in ["Bohr", "[a0]"]
+            self.set_attribute(
+                "origin",
+                utils.convertor(
+                    numpy.array([float(x.strip(',')) for x in line.split()[-4:-1]]),
+                    'bohr',
+                    'Angstrom'
+                )
+            )
 
         # The properties section print the molecular dipole moment:
         #
--- a/cclib/parser/psi4parser.py
+++ b/cclib/parser/psi4parser.py
@@ -1237,12 +1237,12 @@ class Psi4(logfileparser.Logfile):
         line = next(inputfile)
         assert 'Reduced mass' in line
         chomp = line.split()
-        vibrmasses = [utils.float(x) for x in chomp[3:]]
+        vibrmasses = [float(x) for x in chomp[3:]]
 
         line = next(inputfile)
         assert 'Force const' in line
         chomp = line.split()
-        vibfconsts = [utils.float(x) for x in chomp[3:]]
+        vibfconsts = [float(x) for x in chomp[3:]]
 
         line = next(inputfile)
         assert 'Turning point' in line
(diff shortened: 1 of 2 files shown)


cclib/parser/psi4parser.py
```diff
@@ -707,26 +707,29 @@ def get_normalization_factor(exp, lx, ly, lz):
...
+        
         if (self.section == "Convergence Check") and line.strip() == "==> Convergence Check <==" \
             and not hasattr(self, 'finite_difference'):
 
             if not hasattr(self, "optstatus"):
                 self.optstatus = []
             self.optstatus.append(data.ccData.OPT_UNKNOWN)
 
-            self.skip_lines(inputfile, ['b', 'units', 'comment', 'dash+tilde', 'header', 'dash+tilde'])
+            self.skip_lines(inputfile, ['b', 'units', 'b', 'comment', 'b', 'dash+tilde', 'header', 'dash+tilde'])
```
berquist: Ok, so the header changed to include two new blank lines as well. Since this will break older versions of Psi4, you can add a version-specific conditional:
```python
from packaging.version import parse as parse_version
from packaging.version import Version
<snip>
self.skip_lines(inputfile, ['b', 'units'])
if packaging.version.parse(self.metadata["package_version"]) >= Version("1.7"):
    self.skip_line(inputfile, 'b')
self.skip_line(inputfile, 'comment')
if packaging.version.parse(self.metadata["package_version"]) >= Version("1.7"):
    ...
self.skip_lines(inputfile, ['dash+tilde', ...])
```
This isn't efficient but the logic should be right.

cclib/parser/psi4parser.py
```diff
@@ -1157,19 +1165,35 @@ def parse_gradient(self, inputfile, skip_lines):
...
-        #       4   H                0.02  0.32 -0.36    0.01  0.16 -0.34    0.60 -0.01  0.01
-        #       5   H                0.02 -0.33 -0.39    0.01  0.13  0.31    0.60  0.01  0.01
+        #  Vibration                       1                   8                   9
+        #  Freq [cm^-1]                698.2090i           1675.6707           1675.6899
+        #  Irrep                           B2                  B1                  A1
+        #  Reduced mass [u]              1.1842              1.0783              1.0783
+        #  Force const [mDyne/A]        -0.3401              1.7838              1.7839
+        #  Turning point v=0 [a0]        0.0000              0.2581              0.2581
+        #  RMS dev v=0 [a0 u^1/2]        0.0000              0.1895              0.1895
+        #  IR activ [km/mol]            695.8115            139.3986            139.4122
```
berquist: You might want to say something here about IR intensity only being available for >= 1.7.

cclib/parser/psi4parser.py
```diff
@@ -1197,6 +1221,11 @@ def parse_vibration(n, inputfile):
         line = next(inputfile)
         assert 'RMS dev' in line
 
+        line = next(inputfile)
```
berquist: Here is where you'd add another version conditional check.

cclib/parser/psi4parser.py
```diff
@@ -1197,6 +1221,11 @@ def parse_vibration(n, inputfile):
         line = next(inputfile)
         assert 'RMS dev' in line
 
+        line = next(inputfile)
+        assert 'IR activ' in line
+        chomp = line.split()
+        vibirs = [utils.float(x) for x in chomp[3:]]
```
berquist: Even though older versions won't be able to parse this, in order to have the return value of the method be consistent, you can probably return the empty list when they aren't available.

cclib/parser/psi4parser.py
```diff
@@ -1080,6 +1085,9 @@ def get_normalization_factor(exp, lx, ly, lz):
             if len(vibdisps) == n_modes:
                 self.set_attribute('vibfconsts', vibfconsts)
 
+            if len(vibirs) == n_modes:
```
berquist: ...and empty `vibirs` for older versions means this will still work.

berquist (dismissed):
Thanks for the fixes! A couple of things...

1. The Psi4 unit tests are failing since they're all for versions older than 1.7 (https://github.com/cclib/cclib/tree/3cea5df4f90e462625b3ff0381d610e07836cdc4/data/Psi4). We're long overdue for a set of additions, and adding 1.7 would require your fixes. The procedure for doing so is documented at https://cclib.readthedocs.io/en/latest/development.html#adding-a-new-program-version and https://github.com/cclib/cclib/issues/555#issuecomment-410431978), but I wouldn't want to make you go through all that work unless you were interested. Thoughts?

2. The line ending changes should be removed from the PR until we fix the line endings across the entire repository (#536) since it messes up `git blame` and we want to be able to use something like https://akrabat.com/ignoring-revisions-with-git-blame/ to hide those changes. Were you having a problem here?

mskblackbelt:
I'm fine nixing the line endings, just thought I'd clean them up while I was at it (my editor complained a bit, so I let it do the conversion in a second commit). 

Regarding the updated testing: this is my first significant commit to a project, thought I'd dip my toes in with something simple. I'm up for the challenge of trying to update the testing, but have not done anything on this scale before. At the moment, I've copied all of the existing tests in `data/Psi4/basicPsi4-1.3.1` and rerun them in a `basicPsi4-1.7` folder. I'll read through the guidelines in the Development docs and see where I need to go next… Forgot to consider that the updates to the parser might break the versions. It seems to me there should be a better way to read through headers than to count and skip lines. Am I being naïve there, or is there a simple way to account for a wildcard blank line in the grouping?

berquist:
> (my editor complained a bit, so I let it do the conversion in a second commit)

We'll most likely end up with https://stackoverflow.com/a/10855862 after converting everything to LF, so people don't have to mess with their Git config.

> At the moment, I've copied all of the existing tests in `data/Psi4/basicPsi4-1.3.1` and rerun them in a `basicPsi4-1.7` folder. I'll read through the guidelines in the Development docs and see where I need to go next…

Yes, it's more tedious than anything else. What often happens is that there is a PR that does this copy/update dance but does not actually add the new files to `testdata`, since (as you discovered) there are problems with the new outputs. Then there is a PR like this one that fixes the problems and turns on the new tests.

> It seems to me there should be a better way to read through headers than to count and skip lines. Am I being naïve there, or is there a simple way to account for a wildcard blank line in the grouping?

One advantage to doing it this way is that it's safe: users get at least a warning when something is off, and usually a gross traceback soon after. It comes at the cost of being unfriendly to users. We don't have wildcards of any kind for `skip_lines`, which might be a nice idea. When we want to skip lots of things and not care about the contents, `skip_lines` doesn't get used at all and we go straight to
```python
while "some important starting text" not in line:
    line = next(inputfile)
```

berquist:
Hi Dustin, not to pressure you, but let us know if you need any guidance or help with this.  I am happy to do all the boilerplate stuff (rerunning 1.7 calculations and adding them without enabling them) if you'd rather not.

shivupa (approved):

PR #1189, opened 2023-03-22
ORCA: fix parsing basis set, method, and etsyms from semiempirical methods

### LINKED ISSUE #1187: ZINDO vs CIS output in ORCA calculation
Dear all,

While using theoDORE for analysis of my calculations, I found out that the orcaparser.py only works for CIS type calculations, and that when doing a ZINDO one, the parser has some problems with the basis set. I know for sure this is the case because I tried both outputs and the CIS one works, while the ZINDO one does not because of this output format issue. Did anyone encountered this problem before? And if so, could you please tell me how did you fix it or how to move on? It is obvious that I can modify the source code, but this seems too much trouble, and I think it has to be possible either to modify the orca output or use a modified version of the orcaparser.py.

Thanks in advance!

berquist:
Fixes #1187

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -257,6 +257,13 @@ class ORCA(logfileparser.Logfile):
                             coords.append(splitter(line))
             self.metadata['keywords'] = keywords
             self.metadata['coords'] = coords
+
+        # Semiempirical methods use a minimal basis fit to Slater functions,
+        # not def2-SVP or whatever default is given before the input file is
+        # echoed.
+        if "FIT TO SLATER BASIS" in line:
+            self.metadata["basis_set"] = line.split()[0][4:]
+
         # If the calculations is a unrelaxed parameter scan then immediately following the 
         # input file block is the following section:
                 
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -436,7 +436,18 @@ class ORCA(logfileparser.Logfile):
                 line = next(inputfile)
             energy = utils.convertor(float(line.split()[3]), "hartree", "eV")
             self.scfenergies.append(energy)
-            self.metadata['methods'].append('HF' if not self.is_DFT else 'DFT')
+            if self.is_DFT:
+                method = "DFT"
+            else:
+                semiempirical_methods = _METHODS_SEMIEMPIRICAL & {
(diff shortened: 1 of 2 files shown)


shivupa (approved):

berquist:
This also needs at least a fix for `etsyms`.

berquist:
@LoraSM can you see if the contents of this pull request solve all the issues you were having?

berquist:
I can't tell if this is a cclib problem or a TheoDORE problem. @felixplasser Do you have any guidance?  The output file in question is at https://github.com/cclib/cclib-data/blob/007050c0a476c1865901b6cb570b5daa11919727/ORCA/ORCA4.2/ligando-30-SRM1-S_ZINDO.out.  I also took our `dvb_td` test file, replaced the keyword line with `zindo/s nousesym`, and it parses the same as in TDDFT.

felixplasser:
At first sight, it looks like this has gone successfully through all the cclib related steps.

What the error means is that TheoDORE is asked to analyze atom number 92 but it can't find 92 atoms. This means that either the input in `at_lists` is wrong or that the number of atoms is not correctly passed on to TheoDORE.

I'll have a look.

felixplasser:
I had a look. For me this works perfectly if I use the new cclib branch along with the following input file for TheoDORE

~~~~
rtype='cclib'
rfile='ligando-30-SRM1-S_ZINDO.out'
at_lists=[[1, 9, 13, 12, 17, 18, 3, 24, 23, 82, 8, 83, 22, 84, 25, 30, 26, 27, 31, 28, 32, 33, 29, 34, 46, 19, 16, 20, 21, 11, 10, 35, 36, 37, 38, 41, 40, 43, 39, 42, 44, 45, 86, 14, 2, 15], [4, 87, 47, 48, 56, 6, 58, 49, 85, 60, 50, 89, 91, 90, 51, 92, 52, 88, 5, 54, 53, 57, 7, 59, 55, 71, 73, 72, 74, 78, 75, 79, 76, 81, 80, 77, 61, 65, 64, 69, 63, 70, 67, 62, 68, 66]]
Om_formula=2
eh_pop=0
comp_ntos=True
jmol_orbitals=True
prop_list=['Om', 'POS', 'PR', 'CT', 'COH', 'CTnt', 'PRNTO', 'Z_HE', 'RMSeh']
~~~~

PR #1191, opened 2023-03-28
Orca Spin-Spin NMR Couplings

oliver-s-lee:
See #125 for discussion.


cclib/parser/data.py
```diff
@@ -68,6 +68,7 @@ class ccData:
         nbasis -- number of basis functions (integer)
         nmo -- number of molecular orbitals (integer)
         nmrtensors -- Nuclear magnetic resonance chemical shielding tensors (dict of dicts of array[2])
+        nmrcouplingtensors -- Nuclear magnetic resonance spin-spin coupling tensors (dict of dicts of array[2])
```
oliver-s-lee, line 71: Should these both not be `dict of dicts of array[3]`?

berquist: The number is the number of array axes/the tensor rank; how many indices are required to access an element?  1 for a vector, 2 a matrix, 3 a cube, etc.

cclib/parser/orcaparser.py
```diff
@@ -1507,6 +1507,94 @@ def energy_intensity(line):
                 anisotropic.append(float(aniso))
 
             self.set_attribute('nmrtensors', nmrtensors)
+            
+        # -----------------------------------------------------------
+        #  NUCLEUS A = C    0 NUCLEUS B = C    1 
+        #  ( 13C  gnA =  1.405  13C  gnB =  1.405) r(AB) =     2.8677
```
berquist, line 1513: Related to your comment about isotope numbers, we may end up wanting the nuclear g-factors themselves, even if they are derived.  That's a "later" thing though.

cclib/parser/orcaparser.py
```diff
@@ -1507,6 +1507,94 @@ def energy_intensity(line):
...
+        # 
+        #  Diagonalized sT*s matrix:
+        #  
+        #  ssDSO           -0.139           -0.218            0.452  iso=       0.032
+        #  ssPSO           -0.041           -0.592            1.227  iso=       0.198
+        #  ssFC             7.420            7.420            7.420  iso=       7.420
+        #  ssSD             3.609            1.085            0.735  iso=       1.810
+        #  ssSD/FC         -4.198            2.217            1.981  iso=      -0.000
+        #         ---------------  ---------------  ---------------  ----------------
+        #  Total            6.651            9.912           11.815  iso=       9.459
```
berquist, line 1550: At the risk of waffling I think we do want the final number, since most people will only care about that and shouldn't have to write NumPy when it's printed right there.

cclib/parser/orcaparser.py
```diff
@@ -1507,6 +1507,94 @@ def energy_intensity(line):
...
+                        t_type = "fermi"
+                    
+                    elif t_type == "spin-dipolar/fermi":
+                        t_type = "spin-dipolar-fermi"
+                
+                    #line = next(inputfile)
+                    
+                    # Read the tensor.
+                    tensor = numpy.zeros((3, 3))
+                    for j, row in zip(range(3), inputfile):
```
berquist, line 1590: I've never thought to do this, it's so obvious in hindsight...

cclib/parser/orcaparser.py
```diff
@@ -1507,6 +1507,94 @@ def energy_intensity(line):
...
+                        tensor[j, :] = list(map(float, row.split()))
```
berquist, line 1591: Decide for yourself which way is clearer, but if you last index runs over all elements, it can be left out:
```suggestion
                        tensor[j] = list(map(float, row.split()))
```
and the same goes for `[j, ...]` (all remaining indices).

berquist (changes requested):
This looks good.  Sorry to go back on https://github.com/cclib/cclib/issues/125#issuecomment-1481188491, but I'm thinking we might want the final isotropic value now.

cclib/parser/data.py
oliver-s-lee: Aye I got you, think this has confused me before for some reason

cclib/parser/orcaparser.py
oliver-s-lee: OK good to know. I ignored them because I couldn't decipher what gn was. This leans more towards having isotope numbers in the dict itself then

cclib/parser/orcaparser.py
oliver-s-lee: Haha thanks. I wish I could take credit for it, but I just copied it from the existing Orca NMR tensor parsing section

cclib/parser/orcaparser.py
oliver-s-lee: Thanks those are nice tips, but yeah `tensor[j]` is cleaner for me

test/data/testNMR.py
```diff
@@ -21,10 +22,36 @@ class GenericNMRTest(unittest.TestCase):
...
+        tensor = list(list(self.data.nmrcouplingtensors.values())[0].values())[0]
+        assert len(tensor) == 7
+        assert tensor['total'].shape == (3, 3)
+        
+        # Check the total isotropic value matches the computed value.
+        total = 0.0
+        for t_type in ("diamagnetic", "paramagnetic", "fermi", "spin-dipolar", "spin-dipolar-fermi"):
+            total += sum(numpy.linalg.eigvals(tensor[t_type])) / len(numpy.linalg.eigvals(tensor[t_type]))
+        
+        self.assertAlmostEqual(total, tensor['isotropic'], 3)
```
berquist, line 53: We are migrating away from unittest to pure pytest and will no longer be able to use unittest methods.  Pure pytest mostly uses bare asserts, though it does come with some numeric helpers like [`approx`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-approx).  The initial translation for `assertAlmostEqual` looks like https://github.com/cclib/cclib/pull/1186/files#diff-691e094e560c34e2a66142c5f70a2b6a8a733a69ad4db43c4eaf106b96f303cbR84 because I used an automated tool to do it.  Can you switch these to `pytest.approx`?

oliver-s-lee: Can do

test/data/testNMR.py
```diff
@@ -17,14 +19,60 @@
...
+class GenericNMRCouplingTest(unittest.TestCase):
+    """Generic NMR spin-spin coupling unittest"""
+    
+    def testkeys(self):
+        """Check dictionary keys are ints."""
+        assert all(set([all((type(key[0]) == int, type(key[1]) == int)) for key in self.data.nmrcouplingtensors.keys()]))
+        
+        assert all(
+            [
+                all((type(isotropekey[0]) == int, type(isotropekey[1]) == int))
```
berquist, line 53: isotrope -> isotope

oliver-s-lee: I knew I was going to make this mistake eventually

berquist (approved):
I am happy with this. If we need to make changes to the format as we introduce other parses, that's fine. Thanks for being patient!

PR #1192, opened 2023-03-29
datetime fix orca

alvarovm:
This fixes how times are printed, also prevents error when the dictionary is dump in jsons.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -2015,21 +2015,17 @@ States  Energy Wavelength    D2        m2        Q2         D2+m2+Q2       D2/TO
             if "cpu_time" not in self.metadata:
                 self.metadata['cpu_time'] = []
             
-            self.metadata['wall_time'].append(datetime.timedelta(
+            dtime = datetime.timedelta(
                 days = days,
                 hours = hours,
                 minutes = minutes,
                 seconds = seconds,
                 milliseconds = milliseconds
-            ))
+            )
+            dtime = f"{dtime}"
+            self.metadata['wall_time'].append(dtime)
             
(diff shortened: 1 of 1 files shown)


cclib/parser/orcaparser.py
```diff
@@ -2015,21 +2015,17 @@ def energy_intensity(line):
...
-            self.metadata['wall_time'].append(datetime.timedelta(
+            dtime = datetime.timedelta(
                 days = days,
                 hours = hours,
                 minutes = minutes,
                 seconds = seconds,
                 milliseconds = milliseconds
-            ))
+            )
+            dtime = f"{dtime}"
```
shivupa, line 2025: is this what is fixing the dump to json? Before we are storing the actual `timedelta` object. If this is what is fixing it, we should fix the json at the dumping point to change `metadata['wall_time']` to str before dumping to json. This would currently be an issue in other parsers as well

shivupa:
Hi @alvarovm thanks for the fix just a quick question.

I'd prefer to carry around `dtime` objects and serialize at the point that we dump to json

alvarovm:
Ok, I need to check how that works, I am not using `.writejson` within cclib, I just pass metadata to larger dictionary then I dump.
is `.writejson` already serializing ? if so, I could just follow the same

langner:
I agree -- let's stick to the more semantic `timedelta` until it actually needs to be serialized.

cclib/parser/orcaparser.py
langner: Also, `str(dtime)` works just as well.

shivupa:
Closing. If this is still necessary, feel free to reopen!

Issue #806, opened 2020-03-10
No SCF correction read from ORCA logfiles (dispersion, gCP, correlation from coupled cluster methods, etc.)


schneiderfelipe:
Hi,

It seems to me that cclib can't read dispersion corrections from ORCA logfiles (#660, #664 and #418 are related to this). But this happens in general with any added correction to the SCF energies, such as gCP or correlation energy from CCSD(T) or DLPNO-CCSD(T). The following input, for instance,

```bash
➜ h2 git:(master) ✗ cat singlepoint.inp 
! DLPNO-CCSD(T) gCP(HF/TZ) def2-TZVP def2-TZVP/C TightSCF Grid5 NoFinalGrid GridX6 NoFinalGridX
%maxcore 1000

*xyz 0 1
H 0 0 0
H 0 0 1
*
```

processes the following information:

```python3
➜ h2 git:(master) ✗ ipython3
Python 3.6.9 (default, Nov 7 2019, 10:44:02) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.8.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import cclib

In [2]: data = cclib.ccopen("singlepoint.out").parse()

In [3]: from scipy.constants import physical_constants

In [4]: c = physical_constants["Hartree energy in eV"][0]

In [5]: data.scfenergies[0] / c
Out[5]: -1.1015466615850964

In [6]: data.getattributes()
Out[6]: 
{'atomcharges': {'mulliken': array([ 0., -0.]), 'lowdin': array([ 0., -0.])},
 'atomcoords': array([[[0., 0., 0.],
  [0., 0., 1.]]]),
 'atommasses': array([1.008, 1.008]),
 'atomnos': array([1, 1], dtype=int32),
 'charge': 0,
 'coreelectrons': array([0, 0], dtype=int32),
 'homos': array([0], dtype=int32),
 'metadata': {'package': 'ORCA',
  'methods': [],
  'success': True,
  'package_version': '4.2.1',
  'warnings': ['MDCI localization with Augmented Hessian Foster-BoysSwitching off randomization!',
  'Post HF methods need fully converged wavefunctionsSetting SCFConvForced trueYou can overwrite this default with %scf ConvForced false'],
  'info': ['the flag for use of LIBINT has been found!'],
  'input_file_name': 'singlepoint.inp',
[...]

PR #1194, opened 2023-04-24
ORCA: parse CCSD(T) energy

berquist:
Part of #806, but does not close it (that might require reading gCP).

I also removed some unnecessary `utils.float` calls for numbers that are not in scientific notation and will never be nan.


amandadumi (approved):
This looks good, thank you!

Issue #590, opened 2018-08-07
Consistency of mpenergies


berquist:
These are the current values of `mpenergies` for every MP2 calculation on water. The correlation energy (first column) is defined as the difference between the total MP2 energy and the HF energy. The table is sorted by `mpcorr`. All values have units of hartree.

|                |        mpcorr |   mpenergy |   scfenergy |
|:---------------|--------------:|-----------:|------------:|
| GAMESS-UK 7.0  | -1964.96      | -2039.92   |    -74.9643 |
| GAMESS-UK 8.0  | -1964.96      | -2039.92   |    -74.9643 |
| Molpro 2006    |    -0.203229  |   -76.2281 |    -76.0249 |
| Molpro 2012    |    -0.203229  |   -76.2281 |    -76.0249 |
| Jaguar 8.3     |    -0.130596  |   -76.1126 |    -75.982  |
| Jaguar 7.0     |    -0.122774  |   -75.7077 |    -75.585  |
| Q-Chem 4.2     |    -0.03805   |   -75.0024 |    -74.9643 |
| NWChem 6.0     |    -0.0380499 |   -75.0024 |    -74.9643 |
| NWChem 6.5     |    -0.0380499 |   -75.0024 |    -74.9643 |
| Psi4           |    -0.0380417 |   -75.0028 |    -74.9648 |
| Gaussian03     |    -0.0379533 |   -75.0023 |    -74.9643 |
| Gaussian09     |    -0.0379533 |   -75.0023 |    -74.9643 |
| Gaussian16     |    -0.0379533 |   -75.0023 |    -74.9643 |
| GAMESS-US 2014 |    -0.0379533 |   -75.0023 |    -74.9643 |
| GAMESS-US 2017 |    -0.0379533 |   -75.0023 |    -74.9643 |
| Firefly 8.0    |    -0.0379533 |   -75.0023 |    -74.9643 |
| Molcas 2018    |    -0.0364238 |   -75.0009 |    -74.9645 |
| DALTON 2013    |    -0.0117853 |   -73.4464 |    -73.4346 |
| DALTON 2015    |    -0.0117853 |   -73.4464 |    -73.4346 |

Only the GAMESS-UK values are clearly incorrect.

langner:
@berquist is there, in your mind, some other action item here other than fixing the GAMES-UK number?

berquist:
Not until the coordinates of every test job are checked.

PR #1196, opened 2023-04-24
Test reference values for coupled cluster energies

berquist:
I discovered when working on #1070 that it's easy to break these (and also MP), and also that the structures must be pretty different (#590, #407).


shivupa (approved):

shivupa:
More specialized versions of tests, but there doesn't seem to be a better way to do this. LGTM

Issue #1174, opened 2023-01-18
utf-8 decode error when parsing some Turbomole files with --multi


oliver-s-lee:
Some Turbomole output files (for example the output generated by the define program) contain non-utf-8 characters (mostly these are things like the Germanic umlaut character in references). These files will fail to parse when used with the --multi option (the second file is irrelevant, just needed to pass validation):
```
ccget --multi define.out foobar metadata
Attempting to read define.out and dvb
Traceback (most recent call last):
  File "/home/oliver/.local/bin/ccget", line 8, in <module>
    sys.exit(ccget())
  File "/home/oliver/dev/cclib/cclib Src/cclib/scripts/ccget.py", line 178, in ccget
    data = ccread(filename, **kwargs)
  File "/home/oliver/dev/cclib/cclib Src/cclib/io/ccio.py", line 170, in ccread
    return log.parse()
  File "/home/oliver/dev/cclib/cclib Src/cclib/parser/logfileparser.py", line 313, in parse
    for line in inputfile:
  File "/usr/lib/python3.8/fileinput.py", line 248, in __next__
    line = self._readline()
  File "/usr/lib/python3.8/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 505: invalid continuation byte
```

But weirdly parses fine when used alone (this might be specific to my machine?):
```
ccget define.out metadata
Attempting to read define.out
metadata
{'basis_set': 'cc-pVDZ',
 'legacy_package_version': '7.5.0',
 'methods': [],
 'package': 'Turbomole',
 'package_version': '7.5.0',
 'success': True}
```

oliver-s-lee:
Eg file: [define.out.zip](https://github.com/cclib/cclib/files/10444340/define.out.zip)

berquist:
Sorry that this slipped through. Would you be able to try fixing it yourself?

oliver-s-lee:
No problem. Yeah I can look into it when I get a sec.

At first look this seems like another `fileinput` problem (considering this is the major difference between single file and multi-file parsing), so a fix might be tricky. Just to get a sense, how married are we to `fileinput`? I've come across some problems with it before (eg, certain crashes during parsing can leave the module in an unusable state and prevent any further reads by the program), so I wonder if maybe this would be a good nudge to look at something else (or look to see if we can use `fileinput` better).

[...]

PR #1197, opened 2023-04-26
Multi parsing with unusual encodings (fix for #1174)

oliver-s-lee:
Just as a note, multi-file parsing and single-file parsing now both use `errors = 'ignore'`, which can potentially hide encoding errors.


cclib/parser/logfileparser.py
```diff
@@ -124,6 +97,46 @@ def seek(self, pos: int, ref: int) -> None:
             self.pos = self.size
 
 
+def opencompressedfile(filename, mode = "r", object = None, wrap = False):
```
berquist, line 100: Let's start trying to use type annotations, even if we don't check them and they aren't necessarily correct the first time, they can be refined later; I'm not sure about `object` or the return type:
```suggestion
def opencompressedfile(filename: str, mode: str = "r", object: typing.Optional[typing.IO] = None, wrap: bool = False) -> typing.IO:
```

berquist (changes requested):

cclib/parser/logfileparser.py
```diff
@@ -134,39 +147,20 @@ def openlogfile(filename: str, object=None):
...
 
         # This is needed, because fileinput will assume stdin when filename is empty.
         if len(filename) == 0:
             return None
-
-        return FileWrapper(fileinput.input(filename, openhook=fileinput.hook_compressed))
+        
+        # The 'errors' argument of fileinput.FileInput() looks like it should do what we want here,
+        # but it's only available in python3.10 and even then doesn't work properly in conjunction with openhook...
+        return fileinput.FileInput(filename, openhook= lambda filename, mode, encoding = None, errors = None: opencompressedfile(filename, mode))
```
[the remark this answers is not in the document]
oliver-s-lee: Yeah no worries can do. It's essentially just juggling the call signature so it fits with what FileInput expects, but I think I can write it better. 

I think logging should be possible yeah, let me take a look.

cclib/parser/logfileparser.py
berquist: Oh, it was not so much about the argument being called `object` despite overwriting a built-in...it's just that I don't know what the type is here.

I personally annotate all functions, no matter their visibility.

berquist (approved):
Thanks for writing tests!

berquist:
I'm going to make another issue to update the logging, because this is scary looking:
Screenshot from 2023-04-27 22-22-07

PR #1200, opened 2023-05-10
GAMESS: parse hessian

berquist:
A followup to #1199.


cclib/parser/gamessparser.py
```diff
@@ -386,7 +386,7 @@ def extract(self, inputfile, line):
             sym += '-' + line.split()[-1]
             self.etsyms.append(sym)
             # skip 5 lines
-            for i in range(5):
+            for _ in range(5):
```
langner, line 389: You could do skip_lines for these, instead.

langner (approved):

Issue #1127, opened 2022-04-26
MOLDEN: output frequency information


andre-cloud:
Hi there, 

I need a molden file to generate a wigner sampling esamble. I was wondering if cclib could generate it. It has to contain geometry, frequency modes and displacement vector in it. 

Orca v. 5.0.3

shivupa:
We don't currently write out frequency and displacement vectors. Specifically from the tags here which ones would you need from `FREQ FR-COORD FR-NORM-COORD INT` https://www3.cmbi.umcn.nl/molden/molden_format.html

andre-cloud:
FREQ FR-COORD FR-NORM-COORD. I know for sure that ccib outputs FREQ in normal situations, but in this case, ORCA 5.0.3 is not parsed correctly and throws an error.

danielhollas:
@andre-cloud which cclib version are you using? Support for ORCA 5.0 has been merged to the master branch, but not officially released yet. Using master branch I can parse frequency outputs without issue.

andre-cloud:
@danielhollas I'm using 1.7.1

berquist:
> FREQ, FR-COORD, and FR-NORM-COORD. I know for sure that ccib outputs FREQ in normal situations, but in this case, ORCA 5.0.3 is not parsed correctly and throws an error.

>  which cclib version are you using? Support for ORCA 5.0 has been merged to the master branch, but not officially released yet. Using master branch I can parse frequency outputs without issue.

> I'm using 1.7.1 from pip

This is good incentive for us to cut 1.7.2 then. @langner 

I will try and implement the frequency bits for the MOLDEN writer.

PR #1202, opened 2023-05-10
Psi4: parse hessian

berquist:
A followup to #1199.

### DIFF
--- a/cclib/parser/psi4parser.py
+++ b/cclib/parser/psi4parser.py
@@ -929,11 +929,14 @@ class Psi4(logfileparser.Logfile):
                 for info in Psi4.GRADIENT_TYPES.values()
                 if info.header == line.strip()
             ][0]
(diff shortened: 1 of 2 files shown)


shivupa:
Regarding [e3dafd9](https://github.com/cclib/cclib/pull/1202/commits/e3dafd90ec36cc1a0194468f7ed4c299987b00e7) are you going to rerun the job to include that or is that for the future?

berquist:
I'm not sure it's printable with 1.2.1 and 1.3.1.  Things that have changed since then are analytic frequencies implemented and used by default, at least for RHF.  I know it's printed because I was using this code to figure out #1127.  I expect it to be there when I generate the 1.7 test data for #1185.

shivupa:
Ok then I think this is fine to move forward with

shivupa (approved):

PR #1205, opened 2023-05-16
Add an initial bridge from ccdata to rdkit Mol objects

ghutchis:


ghutchis:
Is it worth grabbing all the atomic coordinates or an option to select which set of coordinates to use for bond perception?

berquist:
> Is it worth grabbing all the atomic coordinates or an option to select which set of coordinates to use for bond perception?

We don't standardize on this yet, but in a few places we take the approach of having `atomcoords_index: int = -1` or a similar name in the method call, so that you can specify ones other than the last set if desired.

Are you asking about that generally, or if bond perception should happen on a set of coordinates different from those initially loaded into the molecule (I guess a different rdkit conformer)?

ghutchis:
Using a specific index is a nice idea. I'll also add a parameter if someone wants to grab all the coordinates as separate conformers for some reason.

I'm basically looking for feedback / code review. For example, should I check the RDKit version string rather than trying to import the bond perception module?

berquist:
Ok, I understand better now.  Historically we did try/except for imports before adding this `find_package` helper, but did do version checks for Python itself.  The only other outlier was keeping OB behind try/except still in order to deal with the package restructuring at version 3.

It's really up to you, but my thinking is that `find_package` should always be the first layer that guides someone to install the package, then you can either do an explicit version check or have a try/except that implicitly checks the version if the importable API changed across version boundaries.  If `__version__` is available, that's probably cleanest.

langner:
Bumping this to next release (1.8.1 tentatively), since we want to push out 1.8 soonish.

berquist:
Would it help to move this long if I wrote some tests?

berquist:
This got lost in a Slack chat, but I am wary of us merging any code that does not come with tests.

ghutchis:
If you can help get some tests started, I can make sure they provide good coverage. Sorry for this getting lost in the shuffle.

berquist:
I need to install rdkit in our CI images.

Issue #1171, opened 2023-01-06
Additional Metadata


oliver-s-lee:
Happy new year folks, hope all is well,

I wanted to start this issue to collect some ideas on some possible expansions to the metadata object that would be useful for me (and hopefully others too). This is just a list of suggested new data that could be collected and possible field names for people to discuss. 

Possible new data:
--
- Solvation. Both the solvent and model used would be useful. Might be slightly complicated in that certain solvation methods parametize each solvent by dielectric constant (and/or refractive index for excited states),  and only these parameters are printed in the output (I think ORCA does this?). So it might make sense to have fields for both solvent name and/or parameters and capture whatever is available.
  - `solvent_model` (PCM, COSMO etc)
  - `solvent_name` (Toluene etc)
  - `solvent_params: {'dielectric': 2.4, 'refractive': 1.497}` (ie, a dictionary perhaps of printed solvent parameters? Could also be split into `solvent_dielectric` and `solvent_refractive`?)
- Resolution of the identity and auxiliary basis sets. RI techniques are heavily used in many modern codes to accelerate SCF and post-HF calculations (ORCA and Turbomole at least offer lots of these). These methods use a number of auxiliary basis sets which could be captured in their own fields. The use of RI or not could be a simple flag (`RI: False` eg), but this might not offer enough fidelity as there are many different combinations of RI that can be used. Eg, an MP2 calc could use RI-MP2 but not RIJK etc, so instead it might make more sense to instead modify the method field (eg, `RI-MP2`, `RIJK-MP2`, etc instead of just `MP2`?).
  - `aux_basis: ['def2/J', 'cc-pVDZ/C']` perhaps? Could also use a dict instead?
  - `ri: {'J': True, 'JK': False, 'C': True}` maybe or just `methods: ['RI-HF', 'RI-MP2']`
- Excited state methods. Not sure about this one, but would include things like TD-DFT, TDA, CIS, ADC(2), EOM etc. Could just add to the existing methods field or create a new one?
- Date. A datetime object of when the calculation completed (or a list for each individual calc?)

If any of these look interesting I can potentially write implementations for Gaussian, Turbomole and/or ORCA.

oliver-s-lee:
Some other bits and pieces that came to mind:

- Ownership data
  - `username`: The name of the user who owns the file?
  - `hostname`: Ideally on which the calculation was performed, but this might not be obvious. A list may be required to support multi-node jobs etc.
 - Performance data
[...]

Issue #1093, opened 2022-01-18
Adding symmetry and rotational info for Gaussian outputs


jvalegre:
Dear cclib developers,

We're trying to adapt some of our codes to cclib and found that some features that the codes use weren't available in cclib. In short, there are 7 properties:

- Symmetry number (symmno)
- Point group (point_group)
- Rotational constant (roconst)
- Rotational temperature(s) (rotemp)
- Grid size (grid_size)
- <S**2> after spin annihilation (s2_operator)
- <S**2> before spin annihilation (s2_preannhi)

Currently, we retrieve these parameters in the GoodVibes program (https://github.com/patonlab/GoodVibes) through a for loop in outlines as follows:

`
symmno, point_group, roconst, rotemp = '','',[],[]
grid_size,s2_operator,s2_preannhi = '',0,0
grid_lookup = {1: 'sg1', 2: 'coarse', 4: 'fine', 5: 'ultrafine', 7: 'superfine'}

for i in range(0,len(outlines)):
	if outlines[i].find('Rotational symmetry number') > -1:
		symmno = int((outlines[i].strip().split()[3]).split(".")[0])

	elif outlines[i].find('Full point group') > -1:
		point_group = outlines[i].strip().split()[3]

	elif outlines[i].find('Rotational constants (GHZ):') > -1:
		try:
			roconst = [float(outlines[i].strip().replace(':', ' ').split()[3]),
							float(outlines[i].strip().replace(':', ' ').split()[4]),
							float(outlines[i].strip().replace(':', ' ').split()[5])]
		except ValueError:
			if outlines[i].find('********') > -1:
				roconst = [float(outlines[i].strip().replace(':', ' ').split()[4]),
								float(outlines[i].strip().replace(':', ' ').split()[5])]
	
	elif outlines[i].find('Rotational temperature ') > -1:
		rotemp = [float(outlines[i].strip().split()[3])]
	elif outlines[i].find('Rotational temperatures') > -1:
		try:
			rotemp = [float(outlines[i].strip().split()[3]), float(outlines[i].strip().split()[4]),
						float(outlines[i].strip().split()[5])]
		except ValueError:
			if outlines[i].find('********') > -1:
				rotemp = [float(outlines[i].strip().split()[4]), float(outlines[i].strip().split()[5])]

	elif outlines[i].find('ExpMin=') > -1:
		IRadAn = int(outlines[i].strip().split()[-3])
		grid_size = grid_lookup[IRadAn]
		
[...]

PR #1215, opened 2023-06-22
NWChem new metadata (parsed) and objects (unparsed)

amandadumi:
This is to move forward on (and will replace) #1143 . 

Changes:
- [x] This brings the changes from @jvalegre and myself, but rebased against main. (I wasn't able to figure out how to do this by pushing to the pull request, but can try again if preferred).
- [x] implement parsing of requested metadata objects into the NWChem parser
- [x] @jvalegre introduced attributes he would like parsed in the data.py file, though they are not yet parsed. 

Questions/To Do:
Is it okay to introduced data attributes that are not yet parsed? (See `oniom_energy` and `nmr_anis` as examples). 
Possible solutions:
- Only add the attributes for metadata that is being parsed, and separate the other unparsed attributes to a draft PR.  
- Merge as is  since the unparsed attributes do not introduce a breaking change and  parsing can be implemented in a future PR. 

Notes:
Once the newly parsed metadata attributes  are merged, the cjson writer additions for metadata (#1148) can be merged.


berquist:
> I wasn't able to figure out how to do this by pushing to the pull request, but can try again if preferred

I don't think it's so bad as long as he's the author of the commit, which he still is.  "Add more commits by pushing to the [nwchem_fixes](https://github.com/jvalegre/cclib/tree/nwchem_fixes) branch on [jvalegre/cclib](https://github.com/jvalegre/cclib)." means it should work.  What I do is `gh pr checkout 1143` and it automatically sets up the remote and its tracking branch.  I have `gh` set to use SSH instead of HTTPS for remotes, but I think an HTTPS remote will work too if you log in via `gh auth ...`, not sure though.  Mine looks like
```
$ gh auth status
github.com
  ✓ Logged in to github.com as berquist (/home/eric/.config/gh/hosts.yml)
  ✓ Git operations for github.com configured to use ssh protocol.
  ✓ Token: gho_************************************
  ✓ Token scopes: gist, read:org, repo
```

> Is it okay to introduced data attributes that are not yet parsed?

No, because it can make the PR harder to review if they're intermingled with other new attributes that are used.  That holds even if we don't have the previous problem of multiple PRs adding the same things, used or otherwise.  It also means we'll be able to isolate discussions for them better and most changes are isolated to the parser(s).

My vote is for 1, "Only add the attributes for metadata that is being parsed, and separate the other unparsed attributes to a draft PR".

cclib/parser/nwchemparser.py
```diff
@@ -59,6 +59,17 @@ def extract(self, inputfile, line):
                         "package_version"
                     ] = f"{self.metadata['package_version']}+{revision}"
 
+        if "nproc" in line:
+            self.metadata['processors'] = line.split()[-1]
+        if "Memory information" in line:
+            line_count = 0
```
berquist, line 65: Most of this logic should be replaced with `skip_lines`.

cclib/parser/nwchemparser.py
```diff
@@ -59,6 +59,17 @@ def extract(self, inputfile, line):
                         "package_version"
                     ] = f"{self.metadata['package_version']}+{revision}"
 
+        if "nproc" in line:
+            self.metadata['processors'] = line.split()[-1]
```
berquist, line 63:
```suggestion
            self.metadata['num_processors'] = line.split()[-1]

```
otherwise it could be mistaken for processor names or types or ...

cclib/parser/nwchemparser.py
```diff
@@ -359,11 +370,15 @@ def extract(self, inputfile, line):
...
 
-        #DFT functional information
+        # DFT functional and basis set information
         if "XC Information" in line:
             line = next(inputfile)
             line = next(inputfile)
             self.metadata["functional"] = line.split()[0]
+        
+        if "Grid Information" in line:
+            line = next(inputfile)
```
berquist, line 380: This isn't right.  The section looks like
```
             Grid Information
             ----------------
          Grid used for XC integration:  xfine
          Radial quadrature: Mura-Knowles
          Angular quadrature: Lebedev.
          Tag              B.-S. Rad. Rad. Pts. Rad. Cut. Ang. Pts.
          ---              ---------- --------- --------- ---------
          C                   0.70      100          13.0      1454
          H                   0.35      100          13.0      1202
          Grid pruning is: on
          Number of quadrature shells:  1000
          Spatial weights used:  Erf1
```
so this just takes the dashes.  We need a `skip_line` on dashes.

cclib/parser/nwchemparser.py
```diff
@@ -605,6 +620,12 @@ def extract(self, inputfile, line):
         if "Dispersion correction" in line:
             dispersion = utils.convertor(float(line.split()[-1]), "hartree", "eV")
             self.append_attribute("dispersionenergies", dispersion)
+            
+        # type of dispersion
+        if line.strip().find('disp vdw 3') > -1:
```
berquist, line 625: We need a test for this because these lines aren't present at all in our examples.  From `dvb_dispersion_bp86_d3zero.out`,
```
      Dispersion Parameters
      ---------------------

             DFT-D3 Model
          s8 scale factor  :        1.000000000000
          sr6 scale factor :        1.683000000000
          sr8 scale factor :        1.139000000000
              vdW contrib  :        1.000000000000


             DFT-D3 Model
          s8 scale factor  :       -0.014719930985
          sr6 scale factor :
```
This is D3 (D3(zero), no damping), I haven't tried D3(BJ) in NWChem yet.

cclib/parser/nwchemparser.py
```diff
@@ -1261,6 +1282,27 @@ def extract(self, inputfile, line):
                 self.append_attribute("vibirs", float(line.split()[5]))
                 line = next(inputfile)  # next line
             self.set_attribute("vibfreqs", vibfreqs)
+        # properties related to symmetry
+        if line.strip().find('symmetry #') != -1:
```
berquist, line 1286: I don't think this matches anything.  Do you know what this block is and what "symmetry number" is?  Number of irreps in the PG?

cclib/parser/nwchemparser.py
```diff
@@ -1261,6 +1282,27 @@ def extract(self, inputfile, line):
...
             self.set_attribute("vibfreqs", vibfreqs)
+        # properties related to symmetry
+        if line.strip().find('symmetry #') != -1:
+            symmno = int(line.strip().split()[-1][0:-1])
+            self.set_attribute('symmno', symmno)
+
+        # Grab point group
+        if line.strip().find('symmetry detected') != -1:
+            point_group = line.strip().split()[0]
+            self.set_attribute('point_group', point_group)
```
berquist, line 1293: This is already in metadata as `symmetry_detected` and `symmetry_used` (full and highest	Abelian groups).

cclib/parser/nwchemparser.py
```diff
@@ -1261,6 +1282,27 @@ def extract(self, inputfile, line):
...
+
+        # Grab rotational constants (convert cm-1 to GHz)
+        if line.strip().startswith('A='):
+            roconst, rotemp = [], []
+            roconst.append(float(line.strip().split()[1])*29.9792458)
```
berquist, line 1298: This should use `utils.converter` and `line.split()` should be stored in a variable to avoid multiple `split` calls.

cclib/parser/nwchemparser.py
```diff
@@ -1261,6 +1282,27 @@ def extract(self, inputfile, line):
...
+            symmno = int(line.strip().split()[-1][0:-1])
+            self.set_attribute('symmno', symmno)
+
+        # Grab point group
+        if line.strip().find('symmetry detected') != -1:
+            point_group = line.strip().split()[0]
+            self.set_attribute('point_group', point_group)
+
+        # Grab rotational constants (convert cm-1 to GHz)
+        if line.strip().startswith('A='):
```
berquist, line 1296: I don't think this (or `B=`, `C=`) matches anything.  I'm not finding rotational constants in NWChem at all.

cclib/parser/data.py
```diff
@@ -9,7 +9,7 @@
```
berquist, line 9: We probably need to wait until our meeting to discuss this file.

berquist (changes requested):

cclib/parser/nwchemparser.py
amandadumi: I see the `disp vdw 3` in the input, but that is not repeated in the output from what I can tell.  This might need to be changed in general.  (mostly just a note to myself so i don't forget)

cclib/parser/nwchemparser.py
amandadumi: This is present in `cclib/data/NWChem/basicNWChem7.0/dvb_ir.out`. This certainly means I'll have to add a test for it, so i've added it to the to do list.

cclib/parser/nwchemparser.py
amandadumi: Hm this is also present in `cclib/data/NWChem/basicNWChem7.0/dvb_ir.out`. It is on line 1997, though it's not clear what it represents yet.

amandadumi: Okay, yes it is symmetry factor that goes into calculating the rotational entropy. it is related to the point group, but determined as discussed on [this page](http://molecularmodelingbasics.blogspot.com/2012/12/conformational-and-rotational-entropy.html).

I think this should be removed from this pull request and instead added in the follow when we revisit #1143 and discuss on how to organize these new pieces of information.

cclib/parser/nwchemparser.py
berquist: https://nwchemgit.github.io/Density-Functional-Theory-for-Molecules.html?h=disp#disp-empirical-long-range-contribution-vdw

cclib/parser/nwchemparser.py
berquist: Not sure where I was looking then...

cclib/parser/nwchemparser.py
```diff
@@ -1261,6 +1295,22 @@ def extract(self, inputfile, line):
...
+            roconst.append(utils.convertor(float(split_line[1]),'wavenumber','Hz'))
+            rotemp.append(float(split_line)[4])
+            line=next(inputfile)
+            while line.strip().startswith('B=') or line.strip().startswith('C='):
+                split_line = line.split()
+                roconst.append(utils.convertor(float(split_line[1]),'wavenumber','Hz'))
+                rotemp.append(float(split_line)[4])
+                line=next(inputfile)
+            self.set_attribute('rotconsts', roconst)
+            self.set_attribute('rottemp', rotemp)
```
berquist, line 1312: What do we think about rotational temperature? https://github.com/cclib/cclib/issues/1093#issuecomment-1018112286 definitely a method in the future, but should we parse it too?

berquist:
It's dumb but if you apply https://github.com/cclib/cclib/commit/9f7dea17ab0c305c5818683a52f9f6ce60cc7487 until we remove this it'll solve the CI problem.

cclib/parser/nwchemparser.py
amandadumi: I've added a file to test this for D3BJ with the dvb molecule, but it wouldn't be used through the typical dispersion tests on D3, but instead just to check metadata parsing. Is this overkill?

cclib/parser/nwchemparser.py
amandadumi: Hm, if they are available, having them immediately available and not have to do extra work of a recalculation through a method seems like a benefit to me, but maybe I am not seeing a reason that this could be problematic?

cclib/parser/nwchemparser.py
berquist: Sorry I missed this, I think it's ok since more people are doing D3(BJ) than D3(0).

cclib/parser/data.py
```diff
@@ -155,12 +155,14 @@ class ccData:
...
-       "nsocoeffs":         Attribute(list,    'TBD',                         'N/A'),
-       "nsooccnos":         Attribute(list,    'TBD',                         'N/A'),
+       "nsocoeffs":        Attribute(list,             'TBD',                         'N/A'),
+       "nsooccnos":        Attribute(list,             'TBD',                         'N/A'),
        "optdone":          Attribute(list,             'done',                        'optimization'),
        "optstatus":        Attribute(numpy.ndarray,    'status',                      'optimization'),
        "polarizabilities": Attribute(list,             'polarizabilities',            'N/A'),
        "pressure":         Attribute(float,            'pressure',                    'properties'),
+       "processors":       Attribute(int,              'processors',                  'metadata'),
+       "qm_program":       Attribute(str,              'QM program',                  'metadata'),
```
berquist, line 165: I think we want to get rid of these now since
1. `metadata['num_cpus']` from #1171, #1271
2. `qm_program` can be parsed from `str(self)` or something similar (the class name).

cclib/parser/nwchemparser.py
```diff
@@ -59,6 +59,13 @@ def extract(self, inputfile, line):
                         "package_version"
                     ] = f"{self.metadata['package_version']}+{revision}"
 
+        if "nproc" in line:
+            self.metadata['num_processors'] = line.split()[-1]
```
berquist, line 63:
```suggestion
            self.metadata['num_processors'] = int(line.split()[-1])
```

cclib/parser/nwchemparser.py
```diff
@@ -59,6 +59,13 @@ def extract(self, inputfile, line):
                         "package_version"
                     ] = f"{self.metadata['package_version']}+{revision}"
 
+        if "nproc" in line:
+            self.metadata['num_processors'] = line.split()[-1]
+        if "Memory information" in line:
+            self.skip_lines(inputfile,['d','b','heap','stack','global'])
+            self.metadata['memory'] = line.split()[-2:]
```
berquist, line 66: `int`, but we also want this in bytes.  I don't know if `Mbytes` is actually megabytes.  I guess it isn't, since the number of doubles should be the number of bytes.

cclib/parser/nwchemparser.py
```diff
@@ -605,6 +630,12 @@ def extract(self, inputfile, line):
         if "Dispersion correction" in line:
             dispersion = utils.convertor(float(line.split()[-1]), "hartree", "eV")
             self.append_attribute("dispersionenergies", dispersion)
+            
+        # type of dispersion
+        if line.strip().find('DFT-D3 Model') > -1:
+            self.metadata['dispersion'] = "D3"
```
berquist, line 636: Can we do `D3(0)` and `D3(BJ)`?  (unless there is a better way to resolve the ambiguity)

berquist (changes requested):
I haven't investigated but it hangs for me on parsing the vibrational frequency output.

cclib/parser/nwchemparser.py
amandadumi: yes, absolutely. good call.

cclib/parser/data.py
amandadumi: sorry yes, left these in by mistake.

cclib/parser/nwchemparser.py
```diff
@@ -59,6 +59,13 @@ def extract(self, inputfile, line):
                         "package_version"
                     ] = f"{self.metadata['package_version']}+{revision}"
 
+        if "nproc" in line:
+            self.metadata['num_processors'] = int(line.split()[-1])
+        if "Memory information" in line:
+            self.skip_lines(inputfile,['d','b','heap','stack','global'])
+            self.metadata['memory'] = int(line.split()[2:])*8
```
berquist, line 66:
```suggestion
            self.metadata['memory'] = int(line.split()[2])*8

```

PR #1222, opened 2023-06-30
Fix for not all input objects supporting `last_line`

oliver-s-lee:
Parser now checks before accessing `last_line` in case it's not available, and `fileinput.FileInput` is now correctly wrapped with `FileWrapper` (so it now does have `last_line`)

In theory, we probably want all input types to be wrapped with FileWrapper, but I didn't make this change because 1) there's a weird amount of IO boilerplate code and I don't understand what a lot of it is doing and 2) I'm not sure how well covered by tests IO is...

### DIFF
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -162,7 +162,7 @@ def ccread(source, *args, **kwargs):
         return fallback(source)
 
 
-def ccopen(source, *args, quiet = False, **kwargs):
+def ccopen(source, *args, quiet = False, cjson = False, **kwargs):
     """Guess the identity of a particular log file and return an instance of it.
 
     Inputs:
@@ -192,7 +192,8 @@ def ccopen(source, *args, quiet = False, **kwargs):
         # If the input file isn't a standard compchem log file, try one of
         # the readers, falling back to Open Babel.
         if not filetype:
-            if kwargs.get("cjson"):
+            
+            if cjson:
                 filetype = readerclasses['cjson']
                 
             else:
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -109,25 +109,17 @@ class UnknownOutputFormatError(Exception):
 
 def guess_filetype(inputfile) -> Optional[logfileparser.Logfile]:
     """Try to guess the filetype by searching for trigger strings."""
-    if not inputfile:
-        return None
 
     filetype = None
-    if isinstance(inputfile, str):
-        for line in inputfile:
-            for parser, phrases, do_break in triggers:
-                if all([line.lower().find(p.lower()) >= 0 for p in phrases]):
-                    filetype = parser
-                    if do_break:
-                        return filetype
-    else:
-        for fname in inputfile:
-            for line in inputfile:
-                for parser, phrases, do_break in triggers:
-                    if all([line.lower().find(p.lower()) >= 0 for p in phrases]):
-                        filetype = parser
-                        if do_break:
-                            return filetype
+    print(inputfile)
+
+    for line in inputfile:
+        for parser, phrases, do_break in triggers:
+            if all([line.lower().find(p.lower()) >= 0 for p in phrases]):
+                filetype = parser
+                if do_break:
+                    return filetype
+    
     return filetype
 
 
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -207,6 +207,10 @@ def ccopen(source, *args, quiet = False, cjson = False, **kwargs):
         # We're going to swallow this exception if quiet is True.
         # This can hide a lot of errors, so we'll make sure to log it.
         logging.error("Failed to open logfile", exc_info = True)
+        
+    # If we get here, we've failed to identify the log file type.
+    if not quiet:
+        raise ValueError("Unable to determine the type of logfile {}".format(source))
 
 
 def fallback(source):
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -145,7 +145,7 @@ def ccread(source: Union[str, typing.IO, FileWrapper, list], *args, **kwargs):
     try:
         log = ccopen(source, *args, **kwargs)
         if log:
-            logging.getLogger("cclib").info(f"Identified logfile to be in {log.logname} format")
+            logging.getLogger("cclib").info("Identified logfile to be in {} format".format(type(log).__name__))
 
             return log.parse()
         else:
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -160,8 +160,8 @@ def ccread(source: Union[str, typing.IO, FileWrapper, list], *args, **kwargs):
 def ccopen(
         source: Union[str, typing.IO, FileWrapper, list],
         *args,
-        quiet = False,
-        cjson = False,
+        quiet: bool = False,
+        cjson: bool = False,
         **kwargs
     ):
     """Guess the identity of a particular log file and return an instance of it.
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -182,7 +182,7 @@ def ccopen(
       Molpro, MOPAC, NWChem, ORCA, Psi3, Psi/Psi4, QChem, CJSON or None
       (if it cannot figure it out or the file does not exist).
     """
-    if isinstance(source, str) or not hasattr(source, "__iter__"):
+    if not isinstance(source, list):
         source = [source]
         
     inputfile = None
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -205,7 +205,7 @@ def ccopen(
         # the readers, falling back to Open Babel.
         if not filetype:
             # TODO: This assumes we only got a single file...
-            filename = list(inputfile.input_files)[0]
+            filename = list(inputfile.filenames)[0]
             ext = pathlib.Path(filename).name[1:].lower()
             
             for extension in readerclasses:
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -145,14 +145,9 @@ def ccread(source: Union[str, typing.IO, FileWrapper, list], *args, **kwargs):
     try:
         log = ccopen(source, *args, **kwargs)
         if log:
-            # If the input file is a CJSON file and not a standard compchemlog file
-            cjson_as_input = kwargs.get("cjson", False)
-            if cjson_as_input:
-                return log.read_cjson()
-            else:
-                return log.parse()
             logging.getLogger("cclib").info(f"Identified logfile to be in {log.logname} format")
 
+            return log.parse()
         else:
             logging.getLogger("cclib").info('Attempting to use fallback mechanism to read file')
             return fallback(source)
--- a/cclib/io/ccio.py
+++ b/cclib/io/ccio.py
@@ -6,16 +6,10 @@
 # the terms of the BSD 3-Clause License.
 """Tools for identifying, reading and writing files and streams."""
 
-import atexit
 import io
 import os
-import sys
-import re
 import pathlib
-from tempfile import NamedTemporaryFile
 from typing import Optional
-from urllib.request import urlopen
-from urllib.error import URLError
 
 from cclib.parser import data
 from cclib.parser import logfileparser
@@ -38,6 +32,7 @@ from cclib.parser.psi3parser import Psi3
 from cclib.parser.psi4parser import Psi4
 from cclib.parser.qchemparser import QChem
 from cclib.parser.turbomoleparser import Turbomole
+from cclib.parser.logfilewrapper import FileWrapper
 
 from cclib.io import cjsonreader
 from cclib.io import cjsonwriter
@@ -54,18 +49,6 @@ if _has_cclib2openbabel:
 _has_pandas = find_package("pandas")
 if _has_pandas:
     import pandas as pd
-
-# Regular expression for validating URLs
-URL_PATTERN = re.compile(
-
-    r'^(?:http|ftp)s?://'  # http:// or https://
-    r'(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|'  # domain...
-    r'localhost|'  # localhost...
-    r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})'  # ...or ip
-    r'(?::\d+)?'  # optional port
-    r'(?:/?|[/?]\S+)$', re.IGNORECASE
-
-)
 
 # Parser choice is triggered by certain phrases occurring the logfile. Where these
 # strings are unique, we can set the parser and break. In other cases, the situation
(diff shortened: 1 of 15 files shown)


berquist:
I've converted it to draft only so I know when you're done.

For the coverage, we don't have it turned on in GitHub right now, but you can get it by doing
```bash
python -m pytest -v --cov=cclib --cov-report=html -k 'test_parser' test
```
and open `htmlcov/index.html` in your browser.

oliver-s-lee:
Ok, just right off the bat two io tests are failing because the depend on stdin, which is disabled by pytest. The recommended way around this is to emulate with StringIO, but the tests seem to depend on stdin's specific seek method for some reason? Not really sure what's going on but try to fix them up as best I can

test/parser/testlogfileparser.py
```diff
@@ -80,13 +88,12 @@ def test_data_stdin(self):
             expected_attributes = get_attributes(cclib.io.ccread(path))
             with open(path) as handle:
                 contents = handle.read()
-            # This is fix strings not being unicode in Python2.
-            try:
-                stdin = io.StringIO(contents)
-            except TypeError:
-                stdin = io.StringIO(unicode(contents))
-            stdin.seek = sys.stdin.seek
```
oliver-s-lee, line 88: I don't understand why this is necessary for the test

test/parser/testlogfileparser.py
```diff
@@ -62,11 +63,18 @@ def test_url_seek(self):
 
     def test_stdin_seek(self):
```
oliver-s-lee, line 64: This test does nothing currently...

oliver-s-lee:
I've had a look-over the IO internals and there's a fair bit to do I think. We support lots of different file objects, which is great, but they all get processed at different points. Some processing is done in `io/ccio.py`, some is in `parser/logfileparser.py`. Some functions relating to IO are in `logfileparser.py` and some relating to log files are in `ccio.py`. Some files get opened and closed multiple times for no apparent reason. We do a lot of type checking in multiple places, and ideally it would be moved to one central location.

I think a class based approach is probably the way forward here; have one `InputFile` class that wraps whatever input files/stream we want to support and everything else steps away as much as possible from caring about where it's reading from. I'll hack away and see where I get.

oliver-s-lee:
For (my) reference, a non-exhaustive list of the types of input we can read from at various points:
 - A string (a filename)
 - A pathlib.Path (a filename)
 - A string (a URL)
 - A list of strings (containing any mix of the above)
 - An archived version (.gz, .zip, .bz, .bz2) of any of the above
 - An open file object (a stream, which may or may not be seekable)
At some point, some of the above get wrapped/converted to:
 - cclib.logfileparser.FileWrapper
 - fileinput.FileInput
 - another file object (with `open`)

berquist:
FYI I think some of my type annotations are wrong and shouldn't be trusted.  The code will have to be read as you've done.

> a non-exhaustive list of the types of input we can read from at various points

A bad part of this is that it isn't clear what functionality other people are even using.  I doubt that anyone is passing URLs directly or using the streaming functionality (that's defeated if you're gonna call `seek`...), and probably few are reading archives.

oliver-s-lee:
Yeah no problem at all, I'll go through the type annotations at the end once everything's working ok.

Totally agree re. not clear what's actually being used. I presume at one point all these different inputs were being used by someone, but whether that's still the case is hard to say for sure. The weirdest one for me to understand is the unseekable stream, which I suspect is really just to support `stdin`, but could be wrong. I do wonder how many people are using the archive stuff too, but there is some sense to it I think.

oliver-s-lee:
Ok, must of the grunt work on this is now done and it's ready for review. There's a few more things that would be nice to do at some point, but can't presently without badly breaking backwards compatibility (which should be mostly maintained as is). Some things I don't like:
 - The FileWrapper class is still in `cclib.parser` rather than `cclib.io` where it clearly belongs. Sadly it can't move to IO because of circular dependency issues, which would require some restructuring to get around.
 - There is some low level IO (opening files etc) happening transparently inside FileWrapper (and also by extension in ccopen) which makes it difficult to keep track of open files. Ideally opening files would happen in context managers at the top-most level.
 - FileWrapper isn't a context manager or a real file object (it's probably missing some useful methods)
 - FileWrapper only supports seek() to the start or end. Not a problem at the moment as none of the parsers do seeking anymore, but might be nice to have.

oliver-s-lee:
Also the test is failing for some reason relating to logging that I can't fathom, do we overload the logging module as part of CI?

berquist:
The problem is that the logger is being monkey-patched via https://github.com/numpy/numpy/blob/dabb12ec5c27846b1ded065b5e61eabb89d029ed/numpy/distutils/log.py#L46 on PySCF import:
```
$ python try.py
> /home/eric/.pyenv/versions/miniforge3-4.10.3-10/envs/cclib_310/lib/python3.10/site-packages/numpy/distutils/log.py(46)<module>()
-> _global_log.__class__ = Log
(Pdb) bt
  /home/eric/development/forks/python/cclib_pr_review/try.py(1)<module>()
-> import pyscf
  <frozen importlib._bootstrap>(1027)_find_and_load()
  <frozen importlib._bootstrap>(1006)_find_and_load_unlocked()
  <frozen importlib._bootstrap>(688)_load_unlocked()
  <frozen importlib._bootstrap_external>(883)exec_module()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
  /home/eric/.pyenv/versions/miniforge3-4.10.3-10/envs/cclib_310/lib/python3.10/site-packages/pyscf/__init__.py(113)<module>()
-> from pyscf import lib
  <frozen importlib._bootstrap>(1078)_handle_fromlist()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
  <frozen importlib._bootstrap>(1027)_find_and_load()
  <frozen importlib._bootstrap>(1006)_find_and_load_unlocked()
  <frozen importlib._bootstrap>(688)_load_unlocked()
  <frozen importlib._bootstrap_external>(883)exec_module()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
  /home/eric/.pyenv/versions/miniforge3-4.10.3-10/envs/cclib_310/lib/python3.10/site-packages/pyscf/lib/__init__.py(24)<module>()
-> from pyscf.lib import numpy_helper
  <frozen importlib._bootstrap>(1078)_handle_fromlist()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
  <frozen importlib._bootstrap>(1027)_find_and_load()
  <frozen importlib._bootstrap>(1006)_find_and_load_unlocked()
  <frozen importlib._bootstrap>(688)_load_unlocked()
  <frozen importlib._bootstrap_external>(883)exec_module()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
  /home/eric/.pyenv/versions/miniforge3-4.10.3-10/envs/cclib_310/lib/python3.10/site-packages/pyscf/lib/numpy_helper.py(39)<module>()
-> _np_helper = misc.load_library('libnp_helper')
  /home/eric/.pyenv/versions/miniforge3-4.10.3-10/envs/cclib_310/lib/python3.10/site-packages/pyscf/lib/misc.py(55)load_library()
-> return numpy.ctypeslib.load_library(libname, _loaderpath)
  /home/eric/.pyenv/versions/miniforge3-4.10.3-10/envs/cclib_310/lib/python3.10/site-packages/numpy/ctypeslib.py(137)load_library()
-> from numpy.distutils.misc_util import get_shared_lib_extension
  <frozen importlib._bootstrap>(1027)_find_and_load()
  <frozen importlib._bootstrap>(992)_find_and_load_unlocked()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
  <frozen importlib._bootstrap>(1027)_find_and_load()
  <frozen importlib._bootstrap>(1006)_find_and_load_unlocked()
  <frozen importlib._bootstrap>(688)_load_unlocked()
  <frozen importlib._bootstrap_external>(883)exec_module()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
  /home/eric/.pyenv/versions/miniforge3-4.10.3-10/envs/cclib_310/lib/python3.10/site-packages/numpy/distutils/__init__.py(24)<module>()
-> from . import ccompiler
  <frozen importlib._bootstrap>(1078)_handle_fromlist()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
  <frozen importlib._bootstrap>(1027)_find_and_load()
  <frozen importlib._bootstrap>(1006)_find_and_load_unlocked()
  <frozen importlib._bootstrap>(688)_load_unlocked()
  <frozen importlib._bootstrap_external>(883)exec_module()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
  /home/eric/.pyenv/versions/miniforge3-4.10.3-10/envs/cclib_310/lib/python3.10/site-packages/numpy/distutils/ccompiler.py(20)<module>()
-> from numpy.distutils import log
  <frozen importlib._bootstrap>(1078)_handle_fromlist()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
  <frozen importlib._bootstrap>(1027)_find_and_load()
  <frozen importlib._bootstrap>(1006)_find_and_load_unlocked()
  <frozen importlib._bootstrap>(688)_load_unlocked()
  <frozen importlib._bootstrap_external>(883)exec_module()
  <frozen importlib._bootstrap>(241)_call_with_frames_removed()
> /home/eric/.pyenv/versions/miniforge3-4.10.3-10/envs/cclib_310/lib/python3.10/site-packages/numpy/distutils/log.py(46)<module>()
-> _global_log.__class__ = Log
```

Until NumPy is fixed (this behavior is terrible, I will try to make a ticket there), we can try putting back the proper logger at the point where we instantiate it.

oliver-s-lee:
I've gotten around this for now by replacing all instances of the root logger with a custom one for cclib, which we probably want to do anyways?

Sorry I only just saw your comment and already went ahead and opened an issue for this [here](https://github.com/numpy/numpy/issues/24213).

berquist:
> I've gotten around this for now by replacing all instances of the root logger with a custom one for cclib, which we probably want to do anyways?

I think so, but also we should not proliferate logging instances.  Is this a separate one from what's instantiated for each `Logfile` (derived) instance?
 
> Sorry I only just saw your comment and already went ahead and opened an issue for this [here](https://github.com/numpy/numpy/issues/24213).

No, that's great, it's a much better report than I was going to give.

oliver-s-lee:
Yeah agreed. Yes it's separate, but all logging done outside of the parser is now with the same `cclib` instance (instead of the root logger it was before this PR). Each parser still creates its own unique instance using the name of the logfile being parsed, which is quite nice for seeing where the parser's warnings are coming from.

Haha no problem. They have just closed the bug report though so what good it will do remains to be seen...

berquist:
This is why I'm tiring of open source. That person obviously didn't read any of what you said and assumed you were intentionally using distutils.

We could propose an issue to PySCF since all their downstream projects would be affected.

berquist:
I promise to give this a review by the end of my day today.

cclib/io/ccio.py
```diff
@@ -161,24 +141,35 @@ def ccread(source, *args, **kwargs):
...
         else:
-            return log.parse()
-    else:
-        if kwargs.get('verbose', None):
-            print('Attempting to use fallback mechanism to read file')
-        return fallback(source)
-
-
-def ccopen(source, *args, **kwargs):
+            if kwargs.get('verbose', None):
```
berquist, line 157: Not originally your code, but this should be
```suggestion
            if kwargs.get('verbose', False):
```

cclib/io/ccio.py
```diff
@@ -161,24 +141,35 @@ def ccread(source, *args, **kwargs):
...
+                print('Attempting to use fallback mechanism to read file')
+            return fallback(source)
+    
+    finally:
+        if log:
+            log.inputfile.close()
+
+
+def ccopen(
+        source: Union[str, typing.IO, FileWrapper, list],
```
berquist, line 167: `list` of what?  It should be `typing.List[...]` and `...` might be the union of the other three (I don't know).

cclib/io/ccio.py
```diff
@@ -161,24 +141,35 @@ def ccread(source, *args, **kwargs):
...
+        *args,
+        quiet = False,
```
berquist, line 169:
```suggestion
        quiet: bool = False,
```
etc.

test/parser/testlogfileparser.py
```diff
@@ -25,48 +26,68 @@
...
         with open(fpath, 'r') as fobject:
             wrapper = cclib.parser.logfileparser.FileWrapper(fobject)
-            wrapper.seek(0, 0)
-            assert wrapper.pos == 0
-            wrapper.seek(10, 0)
-            assert wrapper.pos == 10
-            wrapper.seek(0, 2)
-            assert wrapper.pos == wrapper.size
+            self.check_seek(wrapper)
+#             wrapper.seek(0, 0)
```
berquist, line 46: Let's just get rid of all the commented blocks.  That will make it clearer how much of an improvement this is.

test/parser/testlogfileparser.py
```diff
@@ -17,6 +17,7 @@
 
 import cclib
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
```
berquist, line 20: Does
```suggestion
from pytest.monkeypatch import MonkeyPatch
```
not work?  `_pytest` is the private implementation that we technically shouldn't rely on.

cclib/parser/turbomoleparser.py
```diff
@@ -963,10 +1001,10 @@ def extract(self, inputfile, line):
         ## For UHF:
         #       occ. orbital   energy / eV   virt. orbital     energy / eV   |coeff.|^2*100
         #         7 a   alpha          -15.12     12 a   alpha            4.74       24.5
-        #         7 a   beta           -15.12     12 a   beta             4.74       24.5        
-        if "Excitation energy:" in line and "excitation" in self.last_lines[-5].split():
+        #         7 a   beta           -15.12     12 a   beta             4.74       24.5
+        if "Excitation energy:" in line and "excitation" in self.inputfile.last_lines[-6].split():
```
berquist, line 1005: For the index change from -5 to -6, what's the trigger line look like?  It should probably go as a code comment.

cclib/parser/turbomoleparser.py
```diff
@@ -64,9 +63,48 @@ def __init__(self, *args, **kwargs):
         self.hours_regex = re.compile(r"([0-9.]*) hours")
         self.minutes_regex = re.compile(r"([0-9.]*) minutes")
         self.seconds_regex = re.compile(r"([0-9.]*) seconds")
-
-        # A list of previous lines to allow look-behind functionality.
-        self.last_lines = collections.deque([""] * 10, 10)
+        
+    def sort_input(self, file_names: list) -> list:
```
berquist, line 67: Putting this alongside the parser is ok, but the reason it was a bare function is that `self` isn't necessary.  I could see it being an abstract class method in the future, for other parsers that can take multiple files like Molpro, but a top-level (private) function would work just as well.

Same comment about using `list` as a type annotation.  (It translates to `typing.List[Any]` which isn't great.)

cclib/parser/logfilewrapper.py
```diff
@@ -0,0 +1,312 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2023, the cclib development team
+#
+# This file is part of cclib (http://cclib.github.io) and is distributed under
+# the terms of the BSD 3-Clause License.
+
+# TOOD: This file belongs in cclib.io, but circular dependency issues mean it can't go there just now.
```
berquist, line 8: What's the circular dependency?

cclib/parser/logfileparser.py
```diff
@@ -200,35 +51,22 @@ def __init__(self, source: Union[str, Iterable[str], fileinput.FileInput], logle
...
-        else:
-            raise ValueError("Unexpected source type.")
+        if not isinstance(source, FileWrapper):
+            source = FileWrapper(source)
+            # Probably the wrong type given.
+            #raise TypeError("Source does not have an 'input_files' attribute, are you sure it inherits from FileWrapper?") from None
+            
+        self.inputfile = source
+        # If our parser needs a certain file ordering, set that now.
+        self.inputfile.sort(self.sort_input(self.inputfile.filenames))
```
berquist, line 61: Ignore my comment about converting it back to a standalone function.

cclib/parser/logfileparser.py
```diff
@@ -200,35 +51,22 @@ def __init__(self, source: Union[str, Iterable[str], fileinput.FileInput], logle
...
-        elif hasattr(source, "read"):
-            self.filename = f"stream {str(type(source))}"
-            self.isstream = True
-            self.stream = source
-        else:
-            raise ValueError("Unexpected source type.")
+        if not isinstance(source, FileWrapper):
+            source = FileWrapper(source)
+            # Probably the wrong type given.
+            #raise TypeError("Source does not have an 'input_files' attribute, are you sure it inherits from FileWrapper?") from None
```
berquist, line 57: Is it safe to get rid of this?

berquist (changes requested):
I'm not done, but here is a first batch of questions.

oliver-s-lee:
> This is why I'm tiring of open source. That person obviously didn't read any of what you said and assumed you were intentionally using distutils.
> 
> We could propose an issue to PySCF since all their downstream projects would be affected.

Yeah indeed, happily though they have now reopened the issue, so maybe some hope does yet remain :) Hmm yeah I did consider opening an issue on PySCF too, the problem for them though is that there's no easy workaround if they need that ctypeslib function (except to manually fix the logger afterwards I guess?)

cclib/io/ccio.py
oliver-s-lee: Thanks good spot. I'll change this to be just a normal kwarg because I don't see why we'd need to use a kwarg dict here.

oliver-s-lee: Ok it's because the kwargs dict is passed to `ccread`, which passes it unmodified to the `LogFile` or `filereader.Reader` constructor, neither of which inspect for the `verbose` key 🙃. I'll replace all these weird kwarg dicts with proper signatures.

test/parser/testlogfileparser.py
oliver-s-lee: Yeah sorry this is a mistake; my IDE is eclipse which tries to guess where to import stuff from automatically. Doesn't always get it quite right :)

cclib/parser/turbomoleparser.py
oliver-s-lee: This is a bit weird but the trigger's not actually changed, it's just that the new `last_lines` includes the current line automatically, while before it was only updated at the end of the log file parsing for the current line.

The trigger section is in the comment block above, looks like this:
```
        # Excited state info from escf.
        #                          1 singlet a excitation
        # 
        # 
        #  Total energy:                           -112.9060086800507    
        # 
        #  Excitation energy:                      0.3111453714493050    
```

cclib/parser/turbomoleparser.py
oliver-s-lee: Yeah ok I see your point. My main motivation for making it a class method was to take advantage of inheritance to automatically pick the correct sorting function, seeing as how we've already got the correct parser class at this point. Should make future additions easier because there's no need to maintain a table of functions or similar.

I'll change it to a classmethod, although if we'd rather go with non-class function I'm not totally opposed either.

oliver-s-lee: Also, I actually rely on `cclib.io.ccio.sort_turbomole_outputs()` in my own code, which this change obviously breaks. For me that's not a problem, but for others who might rely on it this is a non backwards compatible change. Think it's worth adding back `cclib.io.ccio.sort_turbomole_outputs()` as an alias to the new class method (or whatever we end up going with)?

cclib/parser/logfilewrapper.py
oliver-s-lee: Isn't that the question :)

1) `cclib.parser.logfileparser` imports `FileWrapper` (currently from `cclib.parser.logfilewrapper`) in order to be backwards compatible (so you can pass plain strings to `Logfile`'s constructor and it will automatically wrap with `FileWrapper`)
2) `cclib.io.ccio` imports all parsers which inherit from `cclib.parser.logfileparser.LogFile` in order to return the program specific parser it needs
3) `cclib.io.__init__` imports `cclib.io.ccio`, purely for convenience

Thus `cclib.parser.logfileparser` is unable to import anything under `cclib.io`, because the `__init__` file loads `cclib.io.ccio`, which in turn imports `cclib.parser.logfileparser` again.

I think this should be fixable if `cclib.io.__init__` imports `FileWrapper` early (which thankfully has no cclib dependencies), then you could do `from cclib.io import FileWrapper` and that should avoid the circular dependency I think (although doing `import cclib.io.logfilewrapper` would still be problematic). However, `cclib.io.cjsonreader` also import stuff back from `cclib.parser`, complicating matters. I suspect a lot of bits and pieces could do with moving around here, so maybe it's best for this to wait until a v2.x is in the works?

cclib/parser/logfileparser.py
oliver-s-lee: Should be aye, the comment is from an earlier version of mine. Current implementation in master is this:
```
        # Set the filename to source if it is a string or a list of strings, which are
        # assumed to be filenames. Otherwise, assume the source is a file-like object
        # if it has a read method, and we will try to use it like a stream.
        self.isfileinput = False
        if isinstance(source, str):
            self.filename = source
            self.isstream = False
        elif isinstance(source, list) and all([isinstance(s, str) for s in source]):
            self.filename = source
            self.isstream = False
        elif isinstance(source, fileinput.FileInput):
            self.filename = source
            self.isstream = False
            self.isfileinput = True
        elif hasattr(source, "read"):
            self.filename = f"stream {str(type(source))}"
            self.isstream = True
            self.stream = source
        else:
            raise ValueError("Unexpected source type.")
```
All this type juggling is now done by FileWrapper, if the source isn't FileWrapper we simple make it one and all should be dandy.

cclib/parser/turbomoleparser.py
berquist: Right, my brain went right to thinking it was `line.split()[-6]`.  This makes sense.

cclib/parser/logfilewrapper.py
berquist: Breaking the dependency can wait.  The big problem is number 2.  I think there is a design pattern that helps with this particular case.

berquist:
> Yeah indeed, happily though they have now reopened the issue, so maybe some hope does yet remain :) Hmm yeah I did consider opening an issue on PySCF too, the problem for them though is that there's no easy workaround if they need that ctypeslib function (except to manually fix the logger afterwards I guess?)

We can wait for changes on NumPy's end.  PySCF isn't doing anything wrong by using ctypeslib.  But letting them know would be good too, since I've seen they perform workaround for other NumPy API changes.

cclib/parser/turbomoleparser.py
berquist: Breaking a function like that across a major or even minor (but not patch) release is ok, at least when you consider how many people were likely using it compared to `ccread` or `ccopen`.

Keeping it as an alias is ok, and if we had an official way of doing deprecation (it's never come up) we could get rid of it after 1.8.x.

But I am thinking the method is the better approach.  It just isn't clear to me yet which one of instance/class/static it should be.  Class is just a good compromise.

cclib/scripts/ccget.py
```diff
@@ -159,19 +159,28 @@ def ccget() -> None:
...
         if verbose:
-            kwargs['verbose'] = True
             kwargs['loglevel'] = logging.INFO
+            logging.getLogger("cclib").setLevel(logging.INFO)
+        
         else:
-            kwargs['verbose'] = False
+            # TODO: Are we sure we want to ignore warnings by default?
             kwargs['loglevel'] = logging.ERROR
+            logging.getLogger("cclib").setLevel(logging.WARNING)
```
berquist, line 179: These two lines look conflicting...is that "error" for the global logger and "warning" for the logfile instance logger?

I would agree with your comment that both should be warning.

berquist (changes requested):
One last thing with log levels.

cclib/parser/turbomoleparser.py
oliver-s-lee: Cool, I'll add an alias and make it emit a warning on first use, that way at least the user/developer is aware they're using a deprecated function.

Class or abstract makes the most sens I think, and to me class always feels more natural than abstract...

cclib/scripts/ccget.py
oliver-s-lee: Yes exactly, but I'll update so both are WARNING so it's less weird :). This will make some parsers appear more noisy (Orca in particular emits quite a few warnings I think), but to me this is the better behaviour so a non-issue.

berquist (approved):
:tada:

PR #1240, opened 2023-07-30
Gaussian: move formatted checkpoint files to testing directory

berquist:
This moves the Gaussian formatted checkpoint outputs under the `data/FChk` directory and starts testing them.


test/data/testTD.py
```diff
@@ -164,10 +169,6 @@ class GaussianTDDFTTest(GenericTDTest):
 
     expected_l_max = 48000
 
-    def testrotatsnumber(self):
```
berquist, line 167: This is duplicated from the parent test class.

berquist:
I think it can still use things for `moments`...

berquist:
A bunch of Horton bridge stuff is broken.  I am not sure how this code wasn't being tested before.

berquist:
Moments can wait, this is too big already.

shivupa (approved):

Issue #1241, opened 2023-07-31
Nuclear Repulsion Energy Parsing for Gaussian Logs


migatt:
Does cclib provide a method for parsing the nuclear repulsion energy from gaussian logfiles?

I did not find anything in the documentation, but PR #142.

berquist:
Yes, the code is located at https://github.com/cclib/cclib/blob/9020b6a839e43573dc5490d976a36dc88907d6ec/cclib/method/nuclear.py#L103 with an example of how to use it at https://github.com/cclib/cclib/blob/9020b6a839e43573dc5490d976a36dc88907d6ec/test/method/testnuclear.py#L70.

I'll make a note that it could be added to docs.  Is this enough to get started with?

migatt:
Thank you very much for your quick and comprehensive answer!

Just to be sure, about which unit are we talking here?

shivupa:
According to the discussion in #1242, The units in the NRE calculation need to be converted to eV. This issue cannot be closed until this is addressed.

PR #1242, opened 2023-07-31
Add nuclear repulsion energy to docs

migatt:
This PR adds a small part of the nuclear method to the docs describing how to obtain the nuclear repulsion energy using cclib.

#1241

### DIFF
--- a/doc/sphinx/methods.rst
+++ b/doc/sphinx/methods.rst
@@ -465,7 +465,7 @@ Calculate the nuclear repulsion energy using the snippet below:
 
     parser = cclib.io.ccopen("test.log")
     data = parser.parse()
-    nuclear = Nuclear.Nuclear(data)
+    nuclear = Nuclear(data)
 
     nre = nuclear.repulsion_energy()
     print(nre)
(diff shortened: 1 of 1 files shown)


doc/sphinx/methods.rst
```diff
@@ -450,6 +450,33 @@ After the calculate() method is called, the following attributes are available:
...
+    from cclib.method import Nuclear
+
+    parser = cclib.io.ccopen("test.log")
+    data = parser.parse()
+    nuclear = Nuclear(data)
+
+    nre = nuclear.repulsion_energy()
+    print(nre)
+
+The value can be also converted to Hartrees (to directly compare with Gaussian logfiles):
```
berquist, line 473: You've exposed a bug.  All energies that aren't excitation energies are supposed to be in eV (yeah, I know...), but your conversion snippet below does indeed produce the number in Hartrees.  That makes the result sort of doubly wrong.

I'd rather fix the units on the return value before updating these docs.  One way around that though is to fix the unit bug separately after merging this PR with the `convertor(...` snippet removed, and one can safely assume it has the same units as all other energies that are documented.

What do you think?

migatt: This sounds great.

shivupa (approved):
Unit fix can be in a following PR. Corresponding issue updated

berquist (approved):
Thanks for doing all of this!

PR #1243, opened 2023-08-02
Output nuclear repulsion energy in eV

### LINKED ISSUE #1241: Nuclear Repulsion Energy Parsing for Gaussian Logs

migatt:
Fixes #1241

### DIFF
--- a/cclib/method/nuclear.py
+++ b/cclib/method/nuclear.py
@@ -15,6 +15,7 @@ import numpy as np
 from cclib.method.calculationmethod import Method
 from cclib.parser.utils import PeriodicTable
 from cclib.parser.utils import find_package
+from cclib.parser.utils import convertor
 
 _found_periodictable = find_package("periodictable")
 if _found_periodictable:
@@ -111,7 +112,7 @@ class Nuclear(Method):
                 zj = self.data.atomnos[j]
                 d = np.linalg.norm(ri-rj)
                 nre += zi*zj/d
-        return nre
+        return convertor(convertor(nre, "bohr", "Angstrom"), "hartree", "eV")
 
     def center_of_mass(self, atomcoords_index: int = -1) -> float:
         """Return the center of mass."""
--- a/cclib/method/nuclear.py
+++ b/cclib/method/nuclear.py
@@ -103,15 +103,15 @@ class Nuclear(Method):
     def repulsion_energy(self, atomcoords_index: int = -1) -> float:
         """Return the nuclear repulsion energy."""
         nre = 0.0
+        atomcoords = convertor(self.data.atomcoords[atomcoords_index], "Angstrom", "bohr")
         for i in range(self.data.natom):
-            atomcoords = self.data.atomcoords[atomcoords_index]
             ri = atomcoords[i]
             zi = self.data.atomnos[i]
             for j in range(i+1, self.data.natom):
                 rj = atomcoords[j]
                 zj = self.data.atomnos[j]
                 d = np.linalg.norm(ri-rj)
-                nre += zi * zj / convertor(d, "Angstrom", "bohr")
+                nre += zi * zj / d
         return convertor(nre, "hartree", "eV")
 
     def center_of_mass(self, atomcoords_index: int = -1) -> float:
--- a/cclib/method/nuclear.py
+++ b/cclib/method/nuclear.py
@@ -14,8 +14,7 @@ import numpy as np
 
 from cclib.method.calculationmethod import Method
 from cclib.parser.utils import PeriodicTable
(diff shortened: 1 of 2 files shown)


migatt:
I am unsure if the index in L110 should be changed to `atomcoords_index`?
https://github.com/cclib/cclib/blob/9020b6a839e43573dc5490d976a36dc88907d6ec/cclib/method/nuclear.py#L107
https://github.com/cclib/cclib/blob/9020b6a839e43573dc5490d976a36dc88907d6ec/cclib/method/nuclear.py#L110

berquist:
It should.  I would take the opportunity to refactor this so that `self.data.atomcoords[atomcoords_index]` gets pulled out to its own var and then `ri` and `rj` use that, so the indexing soup only appears once.

test/method/testnuclear.py
```diff
@@ -74,7 +74,7 @@ def test_repulsion_energy(self) -> None:
             output = f.read()
         line = re.search('Nuclear Repulsion Energy = .* hartrees', output).group()
         nre = float(line.split()[4])
-        nre = utils.convertor(nre, 'Angstrom', 'bohr')
+        nre = utils.convertor(nre, 'eV', 'hartree')
```
berquist, line 77: Remember that this is the value coming directly from the output file, which will always be in atomic units, so
```suggestion
        nre = utils.convertor(nre, 'hartree', 'eV')
```

berquist (changes requested):

cclib/method/nuclear.py
```diff
@@ -111,7 +112,7 @@ def repulsion_energy(self, atomcoords_index: int = -1) -> float:
                 zj = self.data.atomnos[j]
                 d = np.linalg.norm(ri-rj)
                 nre += zi*zj/d
-        return nre
+        return convertor(convertor(nre, "bohr", "Angstrom"), "hartree", "eV")
```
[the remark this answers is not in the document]
migatt: The nuclear repulsion energy is $\sum_{i, j} \frac{z_i\ z_j}{d}$. We have to convert the distance $d$, which is in the numerator, from Angstrom to Bohr radius. If we switch the conversion order, it should result in dividing our nuclear repulsion energy by 1 Bohr. What do you think?

The solution could be to convert the distance directly using the provided convertor method, and then only convert the energy to eV at the end.

cclib/method/nuclear.py
```diff
@@ -104,14 +104,15 @@ def repulsion_energy(self, atomcoords_index: int = -1) -> float:
         """Return the nuclear repulsion energy."""
         nre = 0.0
         for i in range(self.data.natom):
-            ri = self.data.atomcoords[atomcoords_index][i]
+            atomcoords = self.data.atomcoords[atomcoords_index]
```
berquist, line 107: I know what's going on now, sorry for the confusion.  If you place this outside the loop (since it doesn't depend on the loop variable `i`) and wrap it in `convertor(..., "Angstrom", "bohr")`, ...

cclib/method/nuclear.py
```diff
@@ -104,14 +104,15 @@ def repulsion_energy(self, atomcoords_index: int = -1) -> float:
...
+            ri = atomcoords[i]
             zi = self.data.atomnos[i]
             for j in range(i+1, self.data.natom):
-                rj = self.data.atomcoords[0][j]
+                rj = atomcoords[j]
                 zj = self.data.atomnos[j]
                 d = np.linalg.norm(ri-rj)
-                nre += zi*zj/d
-        return nre
+                nre += zi * zj / convertor(d, "Angstrom", "bohr")
```
berquist, line 114: ...and revert this one, everything else should be correct.

berquist (changes requested):

cclib/method/nuclear.py
berquist: I mean have
```python
nre = 0.0
atomcoords = convertor(self.data.atomcoords[atomcoords_index], "Angstrom", "bohr")
for i in range(self.data.natom):
...
```
and then getting rid of `/ convertor(d, "Angstrom", "bohr")`.

migatt: Thank you!

migatt:
Using my example Gaussian logfile, it matches only to the 4th decimal place.

shivupa (approved):

berquist (approved):
I haven't checked it locally, but if the failure is due to the precision change you're mentioning, reducing it is ok.

shivupa (approved):

PR #1250, opened 2023-08-10
Update changelog and bump version to v1.8

berquist:


CHANGELOG
```diff
@@ -1,3 +1,36 @@
...
+Bugfixes:
+    * Fix edge case of XYZ files being interpreted as Turbomole outputs (#1207, #1238)
+    * Fix units for calculating nuclear repulsion energy (migatt, #1241, #1242, #1243)
+    * Fix infrastructure for reading CJSON (#1222, #1234)
+    * Fix edge cases when parsing certain Turbomole outputs (#1174, #1197, #1220, #1222)
+    * Fix automatically updating both cclib.github.io and cclib.readthedocs.io and testing docs build (#709, #1154, #1158, #1203, #1216, #1223, #1227)
+    * Fix precision in parsing Hirshfeld charges from ORCA (#1209, #1213)
+    * Fix parsing Hessian from formatted checkpoint files (#1204)
+
+Developer facing:
```
berquist, line 27: Historically we haven't done this.  Users will not largely not care, but maybe it's helpful for us.  What do we think?

CHANGELOG
```diff
@@ -1,3 +1,36 @@
...
+    * Update copyright years to 2023 (#1245, #1246)
```
berquist, line 28: The same thing goes for tracking the related issue(s)/PR(s).

CHANGELOG
langner: I think it's nice, but is it obvious what the number refers to?

langner:
Thank you for doing this, Eric.

berquist:
> Thank you for doing this, Eric.
> 
> Why did you skip ANNOUNCE?

I skipped ANNOUNCE for now because it requires writing prose, is probably best done after the CHANGELOG, and I was too tired to finish the CHANGELOG.  I'll do more tonight.

CHANGELOG
berquist: Probably not; I could add a line at the top to the effect of "Numbers after pound sign refer to relevant GitHub issues and pull requests".

Perhaps related, maintaining two changelogs (this "traditional" one and the Sphinx one, which I'll update after this is finalized) seems silly.  Can we only keep the Sphinx one or is there a good reason for both?

berquist:
Please review what's currently done for the changelog.  Content and wording are all fair game.  Once this has been iterated on and the Sphinx version is updated I'll update the ANNOUNCE file.

shivupa (approved):

shivupa:
Looks good IMO and I like adding PR and issue numbers. Does it sound good to move forward with the ANNOUNCE?

langner:
Let's consider ANNOUNCE in a separate PR. BTW, do we still need it?

langner (approved):

berquist:
Ok, I will iterate on this a little and then update the Sphinx changelog.  I guess maintaining the two is ok, since the `CHANGELOG` is a standard Unix-style one but the non-standard Sphinx one is pretty in the docs.  I would like to write a silly script that takes the Sphinx one and generates the `CHANGELOG` or whichever direction is easiest.

I am ok with ANNOUNCE in a separate PR as long as that's been done in the past, I'll look tonight.  I also want to look more closely at what files were updated at the time of tagging.

Regarding whether or not to keep going with ANNOUNCE, I assumed (again, not having looked yet) that it was exactly the text used for the mailing list announcement like https://lists.gnu.org/archive/html/info-gnu/2022-02/msg00013.html.  I did check some GNU repositories and while they have `NEWS` (a reduced changelog plus breaking and other usability changes), none store `ANNOUNCE`, it's only on their mailing lists.  If we have to write something for the mailing list, keeping a copy in the repository is ok too.  Going through history and combining all past ones into a `NEWS` might be even better.

I like having developer-facing changes in addition to the features and bugfixes.

langner:
> Regarding whether or not to keep going with ANNOUNCE, I assumed (again, not having looked yet) that it was exactly the text used for the mailing list announcement like https://lists.gnu.org/archive/html/info-gnu/2022-02/msg00013.html. I did check some GNU repositories and while they have `NEWS` (a reduced changelog plus breaking and other usability changes), none store `ANNOUNCE`, it's only on their mailing lists. If we have to write something for the mailing list, keeping a copy in the repository is ok too. Going through history and combining all past ones into a `NEWS` might be even better.
> 
> I like having developer-facing changes in addition to the features and bugfixes.

Sounds good! Started to do that in #1252.

berquist:
> > Regarding whether or not to keep going with ANNOUNCE, I assumed (again, not having looked yet) that it was exactly the text used for the mailing list announcement like [lists.gnu.org/archive/html/info-gnu/2022-02/msg00013.html](https://lists.gnu.org/archive/html/info-gnu/2022-02/msg00013.html). I did check some GNU repositories and while they have `NEWS` (a reduced changelog plus breaking and other usability changes), none store `ANNOUNCE`, it's only on their mailing lists. If we have to write something for the mailing list, keeping a copy in the repository is ok too. Going through history and combining all past ones into a `NEWS` might be even better.
> > I like having developer-facing changes in addition to the features and bugfixes.
> 
> Sounds good! Started to do that in #1252 (and pushed to the main repo by accident first :D revert, feel free to squash that away)

How so, isn't `cclib:release-1.8` correct?

shivupa (approved):

berquist:
I will make the last commit here the tag.

berquist:
Pushing tags is different from branches.  Pushing it to my branch places it at https://github.com/berquist/cclib/releases/tag/v1.8 and (maybe obviously) it doesn't appear here in any way.  I think it needs to be pushed directly to `cclib:release-1.8` once this branch is merged?

berquist:
This is ready.

shivupa (approved):

shivupa:
i think tag gets pushed to master right not to the release branch?

PR #1261, opened 2023-08-27
Update Zenodo citations and links for 1.8

berquist:
Draft until what appears to be the reserved DOI is confirmed.


berquist:
I forgot to update the original comment text.  That is definitely the final record number and DOI.

PR #1271, opened 2023-08-31
Meta performance

oliver-s-lee:
Adds support for number of CPUs and memory used in the calculation in two new metadata keys:
 - `metadata['num_cpus']` - Number of parallel CPUs, integer
 - `metadata['memory']` - Max allowed memory usage in bytes (total, across all CPUs)

WIP for now because we're missing tests


shivupa:
An acceptable test could just be making sure that the attributes are present for say `dvb_sp` tests. I think this is a nice bite size PR and we could have further improvements be other PRs? or did you want this for other parsers as well?

oliver-s-lee:
No no I was thinking of leaving it just for these 3 for now, they are my favourites  anyway :)

I think it shouldn't be too hard to check for actual numbers of CPUs/memory, I'll have a look later this week

oliver-s-lee:
Split memory into `metadata['memory_available']`, which corresponds to the amount of memory requested in the input file, and `metadata['memory_used']` which is the max amount of memory used, as reported by the program (if available).

data/Gaussian/basicGaussian16/dvb_gopt.cjson
shivupa, line 1: this file doesn't need to be in this right?

shivupa (changes requested):
1 change then Good to merge

data/Gaussian/basicGaussian16/dvb_gopt.cjson
oliver-s-lee: oops, that's what you get for bulk adding folders, deleted

shivupa (approved):

PR #1276, opened 2023-09-13
Added a fix for orca excited states sometimes not being printed identically across sections

oliver-s-lee:
Regression test file will follow


shivupa (approved):

oliver-s-lee:
https://github.com/cclib/cclib-data/pull/176 is the regression file

PR #1286, opened 2023-10-10
added attribute nfrozenatom and altered disps loop to only loop over …

hklem:
…atoms that have frequency information.

When using the '-1' atom flags in the molecule specification to freeze atoms during gaussian calculation the gaussian output file does not list those atoms in the vibrational frequency section. This causes issue with gaussianparser.py which loops over self.nqmf (or natoms) to store 'numbers' into self.vibdisps. The error will appear because natoms > the number of lines with that data, so the loop will run over into the next lines. I changed this behavior by setting a self.nfrozenatom attribute that is defined at the same time as self.natoms. Now the loop in question goes over the range self.nqmf-self.nfrozen atoms.


ghutchis:
Do you have an example file? I'm pretty sure this would also break a lot of other parsers. I'm thinking of Open Babel at the moment, but adding a test file would be important.

hklem:
Hi Geoff,

Yeah, of course.

I have uploaded an example file to my branch of the forked repo: https://github.com/hklem/cclib/tree/gaussianparser_atomflags

It should be in data/Gaussian/basicGaussian16/frozen_atom_flags.log

Let me know if you have any issues.

--
Heidi

From: Geoff Hutchison ***@***.***>
Date: Tuesday, October 10, 2023 at 6:17 PM
To: cclib/cclib ***@***.***>
Cc: Klem, Heidi R. (Fed) ***@***.***>, Author ***@***.***>
Subject: Re: [cclib/cclib] added attribute nfrozenatom and altered disps loop to only loop over … (PR #1286)

Do you have an example file? I'm pretty sure this would also break a lot of other parsers. I'm thinking of Open Babel at the moment, but adding a test file would be important.

cclib/parser/gaussianparser.py
```diff
@@ -323,10 +323,14 @@ def extract(self, inputfile, line):
             # Also, in older versions there is bo blank line (G98 regressions),
             # so we need to watch out for leaving the link.
             natom = 0
+            nfrozenatom = 0
             while line.split() and not "Variables" in line and not "Leave Link" in line:
                 natom += 1
+                if line.split()[1] == '-1':
```
berquist, line 329: This is causing failures in other files (`water_ccsd.log`) where there aren't at least two tokens in a line.

berquist (changes requested):
Thanks for this fix.  My concerns are, since I'm not familiar with the frozen atom feature,

1. Keeping track of which atoms are frozen.  We should store their indices (or at least have the Gaussian input file kept) in `metadata`.  This way we don't have to worry about if the indices are contiguous, etc.
2. I agree about testing other things that consume `ccdata` objects, like the Open Babel bridge, in part because it isn't clear which atoms are frozen, but also just general consistency.  A test for this can be added to https://github.com/cclib/cclib/blob/master/test/regression_io.py (making a new `regression_bridge.py` may be better, not sure).
3. The actual location of this output should go in https://github.com/cclib/cclib-data/ and a test for `vibdisps` length in https://github.com/cclib/cclib/blob/master/test/regression.py.

berquist:
@hklem poke.

hklem (commented):
The "frozen atom indices" will be easier to parse and the previous "atom indices" key was already being used for the frags attribute.

berquist:
This PR is almost a year old.  I will revisit it soon to try and figure out what the path forward should be.

PR #1296, opened 2023-10-26
Support for parsing xTB outputs

### LINKED ISSUE #789: Support for xTB
Is support for xtb ([docs](https://xtb-docs.readthedocs.io/en/latest/contents.html), [github](https://github.com/grimme-lab/xtb)) planned? (the xtb program is a package developed by the Grimme group in Bonn that relies on the semiempirical quantum mechanical methods GFNn-xTB.)

Andrew-S-Rosen:
This is a follow-up to the unmerged PR from #1129 and expands upon it substantially, using it as a starting point. Most of the content can be found in [the xtbparser.py file](https://github.com/Andrew-S-Rosen/cclib/blob/rosen-xtb/cclib/parser/xtbparser.py).

Some improvements made here compared to the prior PR:
- The logic behind parsing the various attributes works regardless of if the calculation is a relaxation or static calculation.
- The logic behind parsing the various attributes is no longer based on very strict string searches on things like the number of white spaces and number of "---" characters, which would have been extremely prone to breaking.
- The new parser addresses recent changes to the xTB output file format.
- Significant refactoring was carried out to improve readability and ease of maintenance.
- Several cclib-supported attributes that were not included in the prior PR are now included here.
- Type hints have been added.
- Allow for parsing of the `gradients` file
- Add unit tests
- Miscellaneous bugfixes.

The only aspects I have _not_ carried over are the catchall attributes `atomprop` and `bondprop` proposed in #789. If they are of interest, I recommend porting them over from #1129 after this gets merged. I only wanted to deal with official cclib-supported attributes for now.

Closes #789
Closes #1129.


Andrew-S-Rosen:
@berquist, @shivupa: Could you please let me know where (in the test suite) I should be adding the tests for the xTB logfile parsing? Thanks!

shivupa:
> @berquist, @shivupa: Could you please let me know where (in the test suite) I should be adding the tests for the xTB logfile parsing? Thanks! The development instructions are a little vague about how precisely the logfiles are being tested and what assertions are being applied.

I'd focus on a single point test first. If you have a `dvb_sp.out`, then you can add that in `data/xTB`. Then add a line in this file https://github.com/cclib/cclib/blob/91c8d60a1a3e7b4b1bf902982dddbda60d53551d/test/testdata#L318
That will let the SP test run (https://github.com/cclib/cclib/blob/master/test/data/testSP.py), and then decorate the functions that need skipped for now (or inherit the `Generic_SP_test` if needed).

> Also, looks like we need to add an `XTB` folder in the regression suite: https://github.com/cclib/cclib-data

Yes, cclib-data is for regression tests only. Unit and integration tests live in this repo. I don't expect you need integration tests yet. We mainly add things there if there was a bug report that leads to a fix.

Andrew-S-Rosen:
@awvwgk: I'm giving the preliminary xTB parser a full overhaul since it was a good start but not really in a place I'd feel confident for production.

In giving it a refresh, I realized that I remain unsure where (if anywhere) the spin is reported in the xTB logfile. If the user supplied it as an input, e.g. via `--uhf`, then it will appear in the Calculation Setup block. If they don't specify it all, then there's no mention of spin in the output. It'd be tempting to say "if it's not in the log file, it's spin multiplicity 1," but that isn't true; if someone runs xTB using the detailed input format, those input parameters aren't echoed anywhere in the log file to the best of my knowledge. So, one could run with spin using detailed input, and it'd be nowhere in the logs.

Do you happen to have any suggestions? If my understanding is indeed correct, then it might just have to be an optional return.

Andrew-S-Rosen:
@shivupa: I am pretty much done with the refactoring, updating it for the current version of xTB, bugfixing, and adding features. Mostly just have the tests to run now. However, I have one remaining question in the meantime.

Namely, what is the ideal approach if the gradients (for `grads`) are not stored in the main logfile but are instead stored in a separate file called `gradient`. Is there a way to allow the user to pass in a directory or, perhaps, multiple file paths?

shivupa:
We should be able to handle multiple files (for example, turbomole requires this). I can provide some details later today about this!

Andrew-S-Rosen:
@shivupa: Thanks! A bit of insight would be great when you have a moment 😄 In the meantime, I'll sort out some of these tests. Because xTB is not a DFT method, there are a fair number of fields that will need to be skipped, but that's okay.

shivupa:
Sorry traveling.

So this works for turbomole
```
#cclib/data/Turbomole/basicTurbomole7.5
from cclib.io import ccread
files = ["water_hf_solvent_cosmo/basis", "water_hf_solvent_cosmo/control", "water_hf_solvent_cosmo/coord", "water_hf_solvent_cosmo/energy", "water_hf_solvent_cosmo/mos", "water_hf_solvent_cosmo/statistics", "water_hf_solvent_cosmo/water.log"]
cclib_obj = ccread(files)
print(cclib_obj.getattributes())
```

I think this should just work straight away for the turbomole parser but you might need to write an xtb specific `sort_input`. https://github.com/cclib/cclib/blob/91c8d60a1a3e7b4b1bf902982dddbda60d53551d/cclib/parser/turbomoleparser.py#L73

@oliver-s-lee does that sound right?

oliver-s-lee:
Hi folks, the mechanism for parsing from multiple files is relatively simple, it just uses some magic to make all the files appear as one long file. From the parser's point of view, all you need is to define a `sort_input()` method if you need the files to appear in a given order (if it doesn't matter you can leave this out). `sort_input()` is called with a list of the available input files, and should return a list with the desired order.

If calling from the code, you can just specify a list of files as @shivupa demonstrated. From the command line, you need to give the `--multi/-m` option:
`ccget --multi main.log gradient` etc

Andrew-S-Rosen:
Thanks!! That's very helpful! I'll give that a go.

Andrew-S-Rosen:
Thanks for merging in the RTD fix, @berquist! Haven't forgotten about this PR btw. Had to take a short break after getting a little overwhelmed trying to sort out the logic behind the test suite. 😅 Will be back at it shortly.

Andrew-S-Rosen:
@oliver-s-lee, @shivupa: A follow-up question about parsing data from multiple files. It sounds like cclib operated by concatenating everything in memory to one big file. On the surface, this seems simple, but in practice it leads to some complexities. For instance, xTB outputs a `charges` file that (as the name suggests) contains the charges. It looks like so:

```bash
   -0.56476049
    0.28238024
    0.28238024
```

As you can see, there are no strings in the file that one can use as a flag that this is the `charges` file. The only way to know the strings contain charge data is that it's from the `charges` file itself.

With that said, is there any mechanism in cclib to either parse files independently and then merge the data/metadata or to introduce some sort of flag in the parser that knows the filename associated with the currently parsed line?

oliver-s-lee:
Hi @Andrew-S-Rosen, great to see the xTB support is progressing so well! With regards to multi-file parsing, yes you are largely correct. A minor (but important) technical difference is that the files are not actually all loaded into memory at once, because this would require a large continuous block of memory that would be very wasteful. Rather, each line is read from each file one-by-one, but there's some magic as you say to make it appear like one long-file.

Regardless, the issue of detecting which file is which is an interesting one, and not something we've come up against before I don't think. You can access the name of the current file like this:
`inputfile.filenames[inputfile.file_pointer]`
Which should work fine for most cases. I personally like `pathlib` for dealing with filenames, where you could do something like:
```
if Path(inputfile.filenames[inputfile.file_pointer]).name == "charges":
  ...
```
There might be some fringe cases where this doesn't work, like if we're reading files from a URL where the filename might not be available, but I'd recommend not worrying about those for now :)

For testing, I'll let the other guys confirm as they have more experience with the testing suite, but yes creating a custom XTBSPTest class inheriting from `GenericSPTest` and modifying based on what's available is totally fine. For concepts which don't exist in xTB, you can either just skip the test entirely or change the value to whatever is the closest equivalent, whichever makes the most sense. You can of course add new tests for concepts which are unique to xTB. As a rough example:
```
class XTBSPTest(GenericSPTest):

    @skipForParser('xTB', 'xTB has no concept of basis sets')
    def testnbasis(self):
        pass

    def testscfenergy(self):
        """Is the SCF energy within the target?"""
        assert abs(self.data.scfenergies[-1] - 1.12345) < 40
```

berquist:
> Do I create a custom XtbSPTest class that inherits from GenericSPTest and overwrite any default assertions so they are relevant for xTB? I could use some assistance on how to best do this without writing my own tests that don't conform to the cclib testdata stuff.

Yes.  I will make an example for you and commit it, then you can expand on that.

> I'm assuming y'all would not appreciate if I wrote my own test_xtb_parser.py and just made a manual set of assertions for the various attributes?

Suffer through this for now, but the test suite is being incrementally rewritten to properly follow pytest expectations.

---

One important thing is that `skipFor` decorators go on methods in the parent class and those methods are not overridden in the child.

Andrew-S-Rosen:
@berquist, @oliver-s-lee: Thank you both for the very helpful comments! I should have a clear path forward with this :)

berquist:
Should've said this earlier, but I will comment again when I'm done committing to the branch.

berquist:
Done; my last few commits show how to write customized test cases and skip those test methods whose implementation is missing or doesn't make sense for the program.  The strategy is to incrementally remove each decorator as functionality is added or fixed.

I'd start with `homos`, I saw the attribute is the wrong length.

Andrew-S-Rosen:
@berquist: Thanks a ton!! I really appreciate you setting the stage for this. I'll take it from here and patch things up as needed. I'll let you know when it's ready for review.

Andrew-S-Rosen:
@berquist: How am I meant to run the tests locally? Doing `pytest .` within the `cclib/tests/data` directory doesn't start any tests. Additionally, running `pytest testSP.py` results in all failures because none of the `self` attributes are defined.

berquist:
The principal driver of tests is https://github.com/cclib/cclib/blob/master/.github/scripts/run_pytest.bash.  The key is, from the root directory,
```bash
python -m pytest -v -s --terse test -k "not test_method"
```
where `-s` is so that the main parser tests actually display their output, and `--terse` is a cclib-specific flag to keep some of that main parser output tame.  But because there is a wrapper over all these parser tests so that they only appear as a single pytest test, you can't use `-k` to isolate xTB.

An alternative that might work is
```bash
python -m test.test_data --terse XTB
```
which uses the old `unittest` functionality.

Andrew-S-Rosen:
@berquist: Thanks for the ping. Admittedly, I have been a bit overwhelmed by the gnarly merge conflicts (see below) and the test suite. I just tried to resolve some of the conflicts, but I ended up making things worse so reverted it.

I'm going to put the merge conflicts aside for a moment and try to focus on increasing the test coverage today. If you have a spare moment, getting some help with the conflicts would be huge... some of the files have very large and obscure "conflicts" e.g. see test/regression.py

berquist:
It slipped my mind that you would have to deal with the rebase/merge over the reformatted code.  Just looking at the filenames, the reason for most (all?) of these conflicts is actually line endings: these files were CRLF, I converted them all to LF, and Git is incredibly stupid about attempting to resolve this for reasons I don't know.

I can't get to this during the week, so hack on tests for now, and ~Saturday I will rebase the whole thing (unless you want to...).  Then I will review, since I don't think anyone's done that yet.

Andrew-S-Rosen:
@berquist: Ah, thanks! Good to know about the line endings. With that in mind, I might be able to get the conflict resolved. I'll take a stab but won't spend too much time if it's still a mess.

Will work on more tests!

berquist:
If you do want to try tackling it yourself (you probably don't...), you will want to `edit` each commit in the interactive rebase, do `pre-commit run -a`, and if there are changes, add and amend.

(For the pre-commit command, you can ignore `run -a` currently complaining about how some ReST markup looks like a merge conflict.  I need to fix that...)

Andrew-S-Rosen:
@berquist: Alright, the test suite and I became friends. Everything passes now, and we have three test files being tested: an SP, a GeomOpt, and an IR. I'm going to stop here and await your review. Any further features will be added in a follow-up PR since the essentials should all be here. Tests pass locally.

I will happily take you up on the offer to fix the conflicts... :) 🙏

berquist:
Not sure what the deal is with CI but it doesn't matter since the rebase will happen anyway. I will review today and tomorrow.

cclib/parser/xtbparser.py
```diff
@@ -0,0 +1,861 @@
...
+        """
+        return "$hessian" in line
+
+    def extract(self, inputfile: FileWrapper, line: str) -> None:
+        
+        if self.metadata.get("success") is None:
+            # Initialize as False. Will be overwritten to True if/when appropriate.
+            self.metadata["success"] = False
+
+        version = self._extract_version(line)
```
berquist, line 618: Since most (all?) of the methods don't use `self` or even `cls`, they can be standalone functions placed at the bottom of the file.

cclib/parser/xtbparser.py
```diff
@@ -0,0 +1,861 @@
...
+        if coord_filename is not None:
+            self.metadata["coord_type"] = coord_filename.split(".")[-1]
+
+        method = self._extract_method(line)
+        if method is not None:
+            self.metadata["methods"] = [method]
+
+        charge = self._extract_charge(line)
+        if charge is not None:
+            self.charge = charge
```
berquist, line 636:
```suggestion
            self.set_attribute("charge", charge)
```
https://github.com/cclib/cclib/blob/a4831f9c25327f7a49a263da275c9268c58a2f27/cclib/parser/logfileparser.py#L269
Unfortunately there's no equivalent for `metadata` right now.

cclib/parser/xtbparser.py
```diff
@@ -0,0 +1,861 @@
...
+
+        mult = self._extract_multiplicity(line)
+        if mult is not None:
+            self.mult = mult
+
+        # TODO: Use the `xtbopt.xyz` file for SCF energies if available since it has higher precision.
+        # But will need to be careful about what might happen if other formats are supplied,
+        # such as sdf or POSCAR.
+        if not hasattr(self, "scfenergies"):
+            self.scfenergies = []
```
berquist, line 646: This can go away when...

cclib/parser/xtbparser.py
```diff
@@ -0,0 +1,861 @@
...
+        if self._is_geom_cycle_line(line):
+            scf_energies = []
+            while not self._is_geom_end_line(line):
+                scf_energy = self._extract_geom_energy(line)
+                if scf_energy is not None:
+                    scf_energies.append(scf_energy)
```
berquist, line 652:
```suggestion
                    self.append_attribute("scfenergies", scf_energy)
```

cclib/parser/xtbparser.py
```diff
@@ -0,0 +1,861 @@
...
+
+                line = next(inputfile)
+            self.scfenergies.extend(scf_energies)
```
berquist, line 655: and then this supporting stuff can go away too.

data/XTB/basicXTB6.6.1/test.ipynb
```diff
@@ -0,0 +1,1604 @@
+{
```
berquist, line 1: Accidental addition?

berquist (changes requested):
Overall looks good.  There are two main areas:
1. I'd like to see the instance methods that don't require `self` be moved to standalone (but still private) functions.
2. A fair bit of the code for modifying attributes can be replaced with `set_attribute` and `append_attribute` (there are other helpers too).  I've only commented on the first time I encountered where this could be done, but it's in lots of places.

Andrew-S-Rosen:
@berquist: Thanks for your comments! I believe I have addressed them all now.

berquist (approved):
Thanks! I will now rebase and merge once done.

Andrew-S-Rosen:
Thank you very much for all your help, @berquist!!

PR #1302, opened 2023-11-05
Additional parsers

amandadumi:
This incorporates the skip_line functionality from logfileparser   as a class function into file_handler.  This function might only ever be used in a virtual way, so the extra argument I've added might not be needed, but I wanted to be safe.  

This is adding atomnos and atomcoords.

### DIFF
--- a/cclib/combinator/combinator.py
+++ b/cclib/combinator/combinator.py
@@ -4,6 +4,7 @@ from cclib.parser_properties import atomcoords
 from cclib.parser_properties import atomnos
 from cclib.parser_properties import atommasses
 from cclib.parser_properties import nbasis
+from cclib.parser_properties import nmo
 
 @dataclass
 class combinator():
@@ -15,5 +16,5 @@ class combinator():
 class sp_combinator(combinator):
     def __init__(self):
         self.name = 'single_point'
-        self.job_list =  [[scfenergies,atomcoords,atomnos,atommasses,nbasis]]
+        self.job_list =  [[scfenergies,atomcoords,atomnos,atommasses,nbasis,nmo]]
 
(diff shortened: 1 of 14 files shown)


amandadumi:
Questions:
1. Should i be formatting with black now?
2. Can we discuss what is the plan for porting tests? (or if this was discussed and I'm forgetting, please remind me.)

shivupa:
> Questions:
> 
>     1. Should i be formatting with black now?

Let's do this once on `main` at the end.

>     2. Can we discuss what is the plan for porting tests? (or if this was discussed and I'm forgetting, please remind me.)

This needs to be a topic of discussion on tuesday

cclib/combinator/combinator.py
```diff
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from cclib.parser_properties import scfenergies
+import cclib.parser_properties as cpp
```
shivupa, line 2: lol cpp is very close to `c++`

shivupa: nothing to change

cclib/parser_properties/atomcoords.py
```diff
@@ -0,0 +1,38 @@
...
+        return None
+
+    @staticmethod
+    def parse(file_handler, program:str, ccdata) -> np.ndarray:
+        constructed_data = None
+        if program in atomcoords.known_codes:
+            file_handler.virtual_set()
+            program_parser = getattr(atomcoords, program)
+            constructed_data = program_parser(file_handler, ccdata)
+            file_handler.virtual_reset()
```
shivupa, line 35: it feels like this might be a repeated paradigm can we factor this out? nothing to change right now

shivupa (approved):

cclib/combinator/combinator.py
berquist, line 2: "This triggers me for obvious reasons"
/click resolved conversation/
"Apparently I'm not the only one"

berquist (approved):

berquist:
Formatting will be done after the rebase.

Tests I'm still not sure how this will work with the new pytest stuff which I increasingly think should be merged into `master` and not `main`.

Issue #1180, opened 2023-03-10
Migrate from `setup.py` to `pyproject.toml`


berquist:
https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/

An initial `pyproject.toml` is being added with only mypy for starters in #1179. This would entail creating `[project]` and removing `setup.py` entirely. It will work on all Python versions as long as the `pip` is sufficiently new enough, and it is in our 3.7 CI image (https://github.com/shivupa/cclib-ci/blob/2f8301d98a3a54d1a8cc7e97523fa34d1efc394c/py37/environment.yml#L203).

Reminder to not let #833 get left behind.

PR #1303, opened 2023-11-05
Apply some refactors to code base

Anselmoo:
Expect the files of `cclibparser`

Via https://docs.sourcery.ai applied some basic refactors to make the code a little bit more _pythonic_

### DIFF
--- a/cclib/bridge/cclib2ase.py
+++ b/cclib/bridge/cclib2ase.py
@@ -95,22 +95,18 @@ def write_trajectory(filename, ccdata, popname="mulliken", index=None):
 
         properties = {}
         if hasattr(ccdata, "scfenergies"):
-            properties.update({"energy": ccdata.scfenergies[i]})
+            properties["energy"] = ccdata.scfenergies[i]
         if hasattr(ccdata, "grads"):
             try:
-                properties.update(
-                    {"forces": -ccdata.grads[i] * units.Hartree / units.Bohr}
-                )
+                properties["forces"] = -ccdata.grads[i] * units.Hartree / units.Bohr
             except IndexError:
                 pass
 
         if i == len(ccdata.atomcoords) - 1:  # last geometry
             if hasattr(ccdata, "moments"):
-                properties.update({"dipole": ccdata.moments[1] * units.Bohr})
+                properties["dipole"] = ccdata.moments[1] * units.Bohr
             if hasattr(ccdata, "free_energy"):
-                properties.update(
-                    {"free_energy": ccdata.freeenergy * units.Hartree}
-                )
+                properties["free_energy"] = ccdata.freeenergy * units.Hartree
 
         traj.write(atoms, **properties)
 
--- a/cclib/bridge/cclib2openbabel.py
+++ b/cclib/bridge/cclib2openbabel.py
@@ -7,6 +7,7 @@
 
 """Bridge between cclib data and openbabel (http://openbabel.org)."""
 
+
 from typing import Optional, Set, Union
 
 import numpy as np
@@ -20,7 +21,7 @@ if _found_openbabel:
     # The `try` block is for OB >= 3.0, and `except` is for 2.4.x and older.
     try:
         from openbabel import openbabel as ob
-    except:
+    except Exception:
         import openbabel as ob
 
 
@@ -85,7 +86,7 @@ def makeopenbabel(
     obmol.ConnectTheDots()
     obmol.PerceiveBondOrders()
     obmol.SetTotalSpinMultiplicity(mult)
-    obmol.SetTotalCharge(int(charge))
+    obmol.SetTotalCharge(charge)
     return obmol
 
 
--- a/cclib/bridge/cclib2pyquante.py
+++ b/cclib/bridge/cclib2pyquante.py
@@ -32,9 +32,7 @@ def makepyquante(data):
 
     # Check required attributes.
     required_attrs = {"atomcoords", "atomnos"}
-    missing = [x for x in required_attrs if not hasattr(data, x)]
-
-    if missing:
+    if missing := [x for x in required_attrs if not hasattr(data, x)]:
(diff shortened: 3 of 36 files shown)


Andrew-S-Rosen:
Sweet! I was going to open some PRs for sourcery refactorings but now I don't have to 😄

Anselmoo:
> Sweet! I was going to open some PRs for sourcery refactorings, but now I don't have to 😄

Unfortunately, most of the parsers of https://github.com/cclib/cclib/tree/master/cclib/parser cannot be optimized by `sourcery` in a reasonable time due to the length of the `extract-function.` Even after several hours, the ORCA parser still stays untouched ...  😢 

It might be worse to think about totally refactoring these functions with code patterns like **adapter** or **factory**; see also https://github.com/faif/python-patterns.

Andrew-S-Rosen:
@Anselmoo: Interesting. Are you relying on the cloud version or did you try running `sourcery` locally?

Anselmoo:
> @Anselmoo: Interesting. Are you relying on the cloud version or did you try running `sourcery` locally?

@Andrew-S-Rosen  thx for question

```terminal
sourcery review --check --fix --verbose .
```

I run locally ...

Andrew-S-Rosen:
Ooof! Oh well... 😅

Anselmoo:
1. In terms of cloud solution, you are talking about https://github.com/marketplace/sourcery-ai ? I find that the local solution, after 2-3 time running over and over, find more reactors.
2. Also it can suggest `extract_methods` for avoiding duplicating code sections

Andrew-S-Rosen:
@Anselmoo: 

1. Yes, that's right. I also prefer the local solution. 
2. Makes sense. I was going to suggest something similar, e.g. having it run on a subset of files or for a subset of rules at a time.

Andrew-S-Rosen:
One cautionary point: Did you set the minimum Python version when you ran it locally? By default, it will make some changes that are not compatible with Python 3.8, which cclib still supports.

Anselmoo:
I realized that a little later during the PR. I was taking a look at the `pyproject.toml`,

https://github.com/cclib/cclib/blob/8800925f9e83704f00cf632bf82b467ac8319426/pyproject.toml#L1-L3

where the definitions stay at `3.9`. Probably, somebody should open an Issue and definite the minimum / supported versions, in a `setup.cfg` or/and `setup.py`.

I am more curious about `master` vs. `main`, so if it runs once, the reactors should be applied for `python 3.8`.

Andrew-S-Rosen:
> Probably, somebody should open an Issue and definite the minimum / supported versions, in a setup.cfg or/and setup.py.

Yup, I just raised that issue in #1305.

berquist (changes requested):
Apologies for not responding to this on time, especially since a lot of good discussion has happened here.  Rather than just say this is a "drive by", I'll try and respond to everything here and provide some reasoning for why things are the way they are, for better or for worse.

First, the PR itself.  This is difficult to review because of its size and how many files it touches, but the whole thing doesn't need review for us to discuss it.  It looks like there are at least three kinds of changes here:
- Positive stylistic changes that are more than just formatting.
- Formatting changes like we'd get with Black or Ruff.
- Some stylistic changes that are harmful.

I've left comments about the kinds of stylistic changes are (not so) good.  I'd expect many but not necessarily all of these to be cause by pylint or Ruff, and wouldn't be added as part of reformatting.  We would be open to these changes, separate from any formatting changes. 

The idea with #1122 is that we would do reformatting in one pass once ~most of the open PRs have been merged, so it can be contained in a single PR with a `.git-blame-ignore-revs` that keeps `git blame` helpful and doesn't hurt contributors.  It turns out that PRs are not getting merged and we're hurting contributors anyway.  A reason we are not doing an implementation refactor is that we are working on version 2.0 of the library based on small parser combinators.  It's a WIP in the `main` branch which will be changed to the default once everything is done.  We'll run Black on all the new code as we develop and I'm not sure yet what we'll do on all the old code.  It will not be a full rewrite since all the nitty-gritty for each attribute will be taken out of the individual `extract` methods.

For versioning, currently supported versions are visible in our CI image repository (https://github.com/shivupa/cclib-ci).  There is no particular reason why cclib shouldn't currently work with 3.11 or 3.12.  We'll remove 3.7 and 3.8 from CI once 3.11 and 3.12 are added.  The full `pyproject.toml` migration (#1180) will also happen soon.

Back to this PR.  If you are willing to make smaller PRs with some of these non-formatting changes, that would be helpful.  I'll close this one once the discussion reaches some conclusion.

Anselmoo:
@berquist thx, for your very detailed answer!

> * Formatting changes like we'd get with Black or Ruff.


I would prefer to use `Ruff`, because it also include `isort`, and [many more](https://docs.astral.sh/ruff/rules/)
The nice thing is that we can also include the target version in `pyproject.toml` and use the built-in [PyUpgrade](https://docs.astral.sh/ruff/rules/#pyupgrade-up) to support the maintenance.

Using https://github.com/chartboost/ruff-action as a GitHub Action allows building dependent CI pipelines and quickly rejecting `improper` linted PRs.

So, maybe first taking care about the `pyproject.toml`?

berquist:
I am going to use https://pre-commit.ci/.  Local development should allow as close to the same workflow as CI as possible.

Anselmoo:
> I am going to use https://pre-commit.ci/. Local development should allow as close to the same workflow as CI as possible.

The great thing about https://pre-commit.ci is that it can automatically create pull requests to fix some issues.

Anselmoo:
See also: https://github.com/cclib/cclib/pull/1307#issue-2000379992

berquist:
Closing because #1307 will make this a mess and we have some understanding of the kinds of changes we're looking for.

Anselmoo:
@berquist  should this PR not be reopen after make rebase to #1307? Or better make from stretch again?

Issue #946, opened 2020-09-15
Developer documentation is missing information for local testing


ATenderholt:
Only relevant documentation states:
> When adding new features and tests, the Travis CI testing script can be run locally to generate the HTML coverage pages and ensure that the tests exercise the feature code.

How does one run a single test? This is what I get running a single script:

```
$ python test/test_data.py 
Couldn't find cython becke routine
Traceback (most recent call last):
  File "test/test_data.py", line 46, in <module>
    for tn in module_names}
  File "test/test_data.py", line 46, in <dictcomp>
    for tn in module_names}
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 965, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'test.data.testSP'
```

berquist:
I think you have to run them as modules, not loose files. For example, to run the ORCA TDDFT tests the old way,
```bash
python -m test.test_data --terse ORCA TD
```
but I don't know how to narrow it down further. The way tests are read from `testdata` right now also means that we can't use pytest to automatically discover individual unit tests for the parsers, only for IO, bridges, etc.

PR #1309, opened 2023-12-01
Replace `unittest` with `pytest` for `testdata` and regressions

### LINKED ISSUE #648: pytest: individual unit tests are not displayed

berquist:
Closes https://github.com/cclib/cclib/issues/648
Closes https://github.com/cclib/cclib/issues/649

This PR makes it so that running all tests now consists of
``` bash
python -m pytest
pushd data
bash ./regression_download.sh
popd
python -m pytest test/regression.py
python -m pytest test/regression_io.py
```
where every single test method inside of a class is now a single pytest test, parameterized over files as specified in `testdata`.  The same goes for test functions and classes in `regression.py`.  As a result, `pytest-xdist` now works, with the following timings on my machine for `regression.py`:
| run options             | time (s) |
|-------------------------|----------|
| none (serial)           | 79       |
| `-n 4 --dist worksteal` | 40       |
| `-n 8 --dist worksteal` | 26       |

I tried 16 workers (the number of physical cores on my machine), but this triggers a race condition due to the naive caching mechanism used for storing already parsed results.  Seeing as most people will not likely run tests in parallel, and the current serial implementation would at least double the time for regressions, it seems acceptable.  That could possibly be optimized, along with the test collection time, but my preference is to live with things for a while, and other more important modifications may be necessary as part of the testing the forthcoming 2.0 release.

In order to skip testing methods, that now becomes `-m 'not is_method'` instead of using `-k`.

How unit and regression tests are written is not fundamentally different from before, but most of the magic has been hidden in `conftest.py`.  `testdata` maintains the current format for now, but https://github.com/cclib/cclib-data/blob/master/regressionfiles.yaml replaces `regressionfiles.txt` (see https://github.com/cclib/cclib-data/pull/179).  Instructions are largely in the docstring of the new `Regression` class.  Basic docs for how to write these tests [exists](https://cclib.github.io/development.html#testing) but doesn't help first-time contributors, so that will be a part of https://github.com/cclib/cclib/issues/555 and https://github.com/cclib/cclib/issues/946.

## Questions

Should we continue to require having `regression*.py` run separately, or should `python -m pytest` discover all available tests?  (This should likely wait for the future.)

## Annoyances

Running both unit and regression tests will show a lot of skips with the reason something like
```
SKIPPED [2] test/data/testSP.py: got empty parameter set ['data'], function test_solvent_model at /home/eric/development/cclib_berquist/test/data/testSP.py:769
```
which has something to do with test classes with the fixture that are defined/imported but not used.  For example, `GenericPolarTest` is used as a base class of `ReferencePolarTest`, and is discovered during parameterization, but is not used directly.  This is harmless though.

## Future work

- There is still some `unittest.TestCase` inheritance usage in places.  I'll remove this in a quick follow-up PR.
- `testdata` will be replaced with `testdata.yaml`.
- There is some hacky junk in the `skipForLogfile` and `skipForParser` decorators having to do with fixtures.  The decorators should be made to work generically with any fixtures.
- `make_regression_entries` is used both in a fixture and at file execution for storage in a top-level variable.  Ideally all of these top-level vars, even the caches, would go away.

## Remaining TODOs for this PR

- [x] confirm unit tests
  - Two non-data unit tests now seem to be running that weren't before:
    - `ccframeTest::test_ccframe_call_without_pandas`
    - `_determine_output_formatTest::test_outputclass`
- [x] confirm regression tests
- [x] a `is_.` marker is being created, maybe from the `.` directory? get rid of it
- [x] fix coverage collection in docs build
- [x] figure out `no/nso` stuff
  - Shiv explained to me that without `--future`, the NO attributes are 1-dimensional and NSO 2-D; with the flag, the NSO attributes go away and NO attributes are 1- or 2-D depending upon the reference wavefunction.

## Pytest details

I don't think this stuff needs documentation in the code/official docs itself, since it's either "standard" pytest or already conforms to existing limitations (fixture names/usage), but some of the latter is already documented in the code.

The standard [pytest hooks](https://docs.pytest.org/en/latest/reference/reference.html#hooks) used are
- `pytest_collect_file`: Hijacked to make sure all regressions are parsed even if they are not explicitly tested.
  - Function- and class-level testing works automatically because they all start with `test`/`*Test` and I think they're collected with `pytest_collect_module`.
- `pytest_configure`: Automatically generate markers for `pytest -m`, such as `is_method`, `is_data`, ...
- `pytest_collection_modifyitems`: Add autogenerated markers to matching tests.
- `pytest_generate_tests`: For class-based tests that accept the data fixture, ensure they are parameterized over all requested unit and regression test entries.
- `pytest_session_finish`: Write out coverage information used for building [documentation](https://cclib.github.io/data_dev.html#details-of-current-implementation).

Fixtures created are:
- `_regression_entries`: map normalized filenames to Regression instances, only used internally
- `filename`: (regression) Get the location for a "testnoparse" function.
- `logfile`: (regression) Get the parsed result for a "test" function.
- `data`: (unit) Parse a unit test, placing the filenames and parser name on the data instance.

These fixtures could be generalized to work across both regression and unit tests, and without the `test` and `testnoparse` function name prefix requirements, but that is another thing for the future.


berquist:
I'm fairly certain that test failures are happening because certain tests weren't triggering before that are being properly run now...

berquist:
Oops, wrong button.

berquist:
Leaving marked as draft because there is still some verification and extremely minor code changes required, but requesting review because 99.9% of the code is done.

shivupa:
Nice work. Note sure if I am not 100 percent clear on how the regression tests are being generated exactly, but I roughly follow.

The only thing I have is -- Should we run `-n 2 --dist worksteal`? It seems like we get two cores on github runners https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

Maybe just modify `run_pytest.sh` to take an argument of num jobs and call it with an arg of 2 in the actions?

shivupa (approved):

berquist:
Old on the left, new on the right...trying to get more detail on exact methods from the old unittest code, but MOPAC has few enough I might be able to understand this by hand:
```
Stat(name='ADF', passed=152, skipped=38) != Stat(name='ADF', passed=150, skipped=38) 2 0
Stat(name='DALTON', passed=305, skipped=86) != Stat(name='DALTON', passed=301, skipped=86) 4 0
Stat(name='FChk', passed=164, skipped=139) == Stat(name='FChk', passed=164, skipped=139)
Stat(name='GAMESS', passed=357, skipped=57) != Stat(name='GAMESS', passed=354, skipped=57) 3 0
Stat(name='GAMESSDAT', passed=17, skipped=22) == Stat(name='GAMESSDAT', passed=17, skipped=22)
Stat(name='GAMESSUK', passed=249, skipped=33) != Stat(name='GAMESSUK', passed=246, skipped=33) 3 0
Stat(name='Gaussian', passed=411, skipped=23) != Stat(name='Gaussian', passed=403, skipped=23) 8 0
Stat(name='Jaguar', passed=232, skipped=61) != Stat(name='Jaguar', passed=227, skipped=60) 5 1
Stat(name='MOPAC', passed=10, skipped=12) != Stat(name='MOPAC', passed=9, skipped=12) 1 0
Stat(name='Molcas', passed=81, skipped=29) != Stat(name='Molcas', passed=81, skipped=28) 0 1
Stat(name='Molpro', passed=187, skipped=40) != Stat(name='Molpro', passed=185, skipped=40) 2 0
Stat(name='NBO', passed=6, skipped=34) == Stat(name='NBO', passed=6, skipped=34)
Stat(name='NWChem', passed=455, skipped=60) != Stat(name='NWChem', passed=449, skipped=60) 6 0
Stat(name='ORCA', passed=412, skipped=50) != Stat(name='ORCA', passed=409, skipped=50) 3 0
Stat(name='Psi4', passed=546, skipped=54) != Stat(name='Psi4', passed=540, skipped=54) 6 0
Stat(name='QChem', passed=273, skipped=29) != Stat(name='QChem', passed=271, skipped=29) 2 0
Stat(name='Turbomole', passed=177, skipped=50) != Stat(name='Turbomole', passed=175, skipped=50) 2 0
Stat(name='XTB', passed=27, skipped=43) != Stat(name='XTB', passed=26, skipped=43) 1 0
```

berquist:
> Old on the left, new on the right...trying to get more detail on exact methods from the old unittest code, but MOPAC has few enough I might be able to understand this by hand:

Most of the differences are accounted for by having removed this now-useless test:
```diff
diff --git test/data/testGeoOpt.py test/data/testGeoOpt.py
index 5a13398c..826c4f96 100644
--- test/data/testGeoOpt.py
+++ test/data/testGeoOpt.py
@@ -91,12 +91,6 @@ class GenericGeoOptTest(unittest.TestCase):
         ans = numpy.zeros(self.data.natom, "i")
         numpy.testing.assert_array_equal(self.data.coreelectrons, ans)

-    @skipForParser("Molcas", "The parser is still being developed so we skip this test")
-    def testnormalisesym(self):
-        """Did this subclass overwrite normalisesym?"""
-        # https://stackoverflow.com/a/8747890
-        self.logfile.normalisesym("A")
-
     @skipForParser("Molcas", "The parser is still being developed so we skip this test")
     @skipForParser("MOPAC", "Not implemented.")
     @skipForParser("xTB", "Not implemented yet")
```
so all that remains is
```
Stat(name='Gaussian', passed=409, skipped=23) != Stat(name='Gaussian', passed=403, skipped=23) 6 0
Stat(name='Jaguar', passed=229, skipped=61) != Stat(name='Jaguar', passed=227, skipped=60) 2 1
```

berquist:
It must be from removing these tests that required `future=False`.  I didn't remove the `optdone_bool` `ccData` class, it's just a similar scenario as the NO/NSO testing.
```diff
263a264
> Scan      Gaussian     GenericRelaxedScanTest_optdone_bool   basicGaussian09       dvb_scan.log
264a266
> Scan      Gaussian     GenericRelaxedScanTest_optdone_bool   basicGaussian16       dvb_scan_relaxed.log
266a269
> Scan      Jaguar       GenericRelaxedScanTest_optdone_bool   basicJaguar8.3        dvb_scan.out
```

PR #1307, opened 2023-11-18
Apply ruff for linting and formatting

Anselmoo:
This pull request applies the `ruff` tool for linting and formatting the codebase. It includes the following commits:

- `chore: :wastebasket: Fix formatting and minor issues in code` for Python Files as a preparation commit for the `ruff`-implementation

- `chore: :wastebasket: Fix formatting and minor issues in code` for markdown as a preparation commit for the `ruff`-implementation

- `chore: :wastebasket: Fix formatting and minor issues in code` for test-data as a preparation commit for the `ruff`-implementation

- `feat: :sparkles: Introduce `ruff` for linting`

- `chore: :recycle: Apply new `pre-commit``

- Apply `fix end of files` and default `ruff`

> NOTE: Foe this PR `MyPy` has to be temporary deactivated and then the ruff advanced [rule-set](https://docs.astral.sh/ruff/rules/#rules) activated upon request. 

Fixes: #1306

PR #1310, opened 2023-12-02
Ignore revisions from #1307 in git blame

### LINKED ISSUE #1122: Should we have some sort of code style?

berquist:
Closes #1122

I did everything as outlined in https://www.michaelheap.com/git-ignore-rev/, though one may not want to set a global file:
```bash
git config [--global] blame.ignoreRevsFile .git-blame-ignore-revs
# Mark any lines that have had a commit skipped using --ignore-rev with a `?`
git config --global blame.markIgnoredLines true
# Mark any lines that were added in a skipped commit and can not be attributed with a `*`
git config --global blame.markUnblamableLines true
```

And here are some examples of the output:
```
*06e54bd1 cclib/__init__.py     (Eric Berquist    2023-12-02 09:29:14 -0500 24) # isort: off
?4d22f3a2 src/cclib/__init__.py (Eric Berquist    2017-12-12 11:33:08 -0500 25) from cclib import parser, progress, method, bridge, io
*06e54bd1 cclib/__init__.py     (Eric Berquist    2023-12-02 09:29:14 -0500 26) # isort: on
```
where I think the isort directives show as `*` because they are _only_ from an ignored commit.  The same appears for lines with isolated parens that were not previously on their own lines (see `qchemparser.py` and grep for `2023-12`).

A slightly more confusing case is `adfparser.py`, but the lines from that date are all whitespace and correspond to the CRLF -> LF commit, and those changes seem to constitute a complete change rather than reuse whatever empty line was there before.  But, they're not code lines.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,14 @@
+## https://github.com/cclib/cclib/pull/1307
+# remove trailing whitespace
+37740d43064bc13445b19ff2d3c5f1154f202896
+# fix end of file
+8f3fbf7d1fc060a3c8522343dd103604bd946e5d
+# fix line endings to be LF
+7b2e8701dcb1a1a6c437919b185be78a35c3e2a5
+# ruff for all but cclib/parser
+e986fd9bb37ca0113707f88f6fea7f2318671cdd
+# ruff for cclib/parser
+04b3e4694e001e4b91457674a490b723e17af2b7
+# isort
+3c499da3e44e4a8ae983407c0f07c71996d202d8
+06e54bd1eb663dd948973037f336d5190f4734ce


berquist:
@shivupa and @amandadumi , can one of you look at the `git blame` output of your favorite file after doing at least `git config blame.ignoreRevsFile .git-blame-ignore-revs` and see if it makes sense to you?

amandadumi:
I think this looks good and seems to have worked. In a quick/narrow check,  the nwchem parser went from seeing your name on every line of the git blame  to only on what looks like newly added lines, which seems correct to me.

shivupa (approved):

shivupa:
> A slightly more confusing case is adfparser.py, but the lines from that date are all whitespace and correspond to the CRLF -> LF commit, and those changes seem to constitute a complete change rather than reuse whatever empty line was there before. But, they're not code lines.

I think as long as the code lines are mostly git blame-able we are in business.

Issue #1266, opened 2023-08-28
Build, check, and push distributions in CI


shivupa:
- Create and check installation of distributions (source, wheel) on each commit
- Automatically upload distributions to PyPI when a tag is pushed

_Originally posted by @berquist in https://github.com/cclib/cclib/issues/1249#issuecomment-1672406702_

berquist:
See #1327

berquist:
Done in #1363, #1376, and #1382.

PR #1328, opened 2023-12-29
Fix package installation

### LINKED ISSUE #1327: Faulty install on `master`
@berquist: This is just a head's up in case you're not aware, but `pip` installing from `master` doesn't work right now.

```bash
pip install git+https://github.com/cclib/cclib.git
```

```python
import cclib
```

Traceback:
```python
ModuleNotFoundError: No module named 'cclib'
```

berquist:
Closes #1327
Caused by #1311 
First part of #1266

### DIFF
--- a/.github/workflows/cclib_pytest.yml
+++ b/.github/workflows/cclib_pytest.yml
@@ -55,3 +55,14 @@ jobs:
           fail_ci_if_error: true
           verbose: false
         if: matrix.container == 'shivupa/cclib-ci:py37'
+      - name: Install pypa/build
+        run: |
+          python -m pip install build
+      - name: Build a binary wheel and a source tarball
+        run: |
+          python -m build
+      - name: Store the distribution packages
+        uses: actions/upload-artifact@v3
+        with:
+          name: python-package-distributions
+          path: dist/
--- a/.github/workflows/cclib_pytest.yml
+++ b/.github/workflows/cclib_pytest.yml
@@ -66,3 +66,13 @@ jobs:
         with:
           name: python-package-distributions
           path: dist/
+  check-installation:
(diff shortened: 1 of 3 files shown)


berquist:
Need to see if I can get the test/build workflows to spawn only once pre-commit passes

berquist:
> Need to see if I can get the test/build workflows to spawn only once pre-commit passes

It doesn't look like this is possible with bots, so migrating to https://pre-commit.ci/lite.html would probably be required.  The bot isn't really faster, but we'd miss periodic automatic hook updates, so I won't bother for now.

.github/workflows/cclib_pytest.yml
```diff
@@ -55,3 +55,47 @@ jobs:
...
+      - name: Build distribution packages (binary wheel and source tarball)
+        run: |
+          python -m build
+      - name: Store the distribution packages
+        uses: actions/upload-artifact@v3
+        with:
+          name: python-package-distributions
+          path: dist/
+        # TODO not sure why this doesn't work
+        # if: matrix.container == 'shivupa/cclib-ci:py310'
```
shivupa, line 70: Shouldn't this become `'shivupa/cclib-ci:py310-edge'` and a bit above this `'shivupa/cclib-ci:py37-edge'`

shivupa (approved):
Awesome! Minor comment but I am also fine as is

.github/workflows/cclib_pytest.yml
berquist: No wonder codecov has been broken

shivupa: yeah it seems like we lost a bit of coverage since the list time codecov has been run but its good that it is fixed

PR #1329, opened 2023-12-30
Add Dependabot configuration

berquist:


berquist:
I checked the instructions in https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide and according to the repo settings, everything is already all set...just need to add the configuration file.

PR #1344, opened 2024-01-09
unpin pyscf and biopython optional dependencies

berquist:
- Pinning should happen in a requirements.txt, not pyproject.toml.
- PySCF changed their B3LYP recipe to Gaussian, and older versions are incompatible with SciPy 1.11 (https://github.com/pyscf/pyscf/issues/1783).


berquist:
This is a blocker for https://github.com/shivupa/cclib-ci/pull/24.

shivupa (approved):

shivupa:
Rest will fail because the current images are with an older pyscf right?

PR #1354, opened 2024-01-28
issue regarding the optdone boolean and list convergence resolved

Anjaneya02:
Single line changed about the optdone the convergence changed from boolean to list


berquist (approved):
Thanks!  Accepting this with the understanding that the behavior returning it as a list of integers, currently restricted to using `--future` or `future=True`, is becoming the default.

PR #1357, opened 2024-01-30
fix population method tests for switch from unittest to pytest

berquist:
Fixes for #1346


berquist:
Random failure, but I also need to add something to filter paths in CI so that modifying method tests themselves triggers method tests.

shivupa (approved):

PR #1363, opened 2024-02-07
Push to PyPI on tag

### LINKED ISSUE #1266: Build, check, and push distributions in CI

berquist:
Closes #1266

Maybe it should be `if: startsWith(github.ref, 'refs/tags/release-')`?

I have no idea how the TestPyPI pushing is going to work, but I ripped it from the official docs, so :crossed_fingers:

### DIFF
--- a/.github/workflows/cclib_pytest.yml
+++ b/.github/workflows/cclib_pytest.yml
@@ -66,6 +66,10 @@ jobs:
       - name: Build distribution packages (binary wheel and source tarball)
         run: |
           python -m build
+      - name: Check packages with twine
+        run: |
+          python -m pip install twine
+          python -m twine check dist/*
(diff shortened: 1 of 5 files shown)


berquist:
Screenshot from 2024-02-06 23-53-38

berquist:
The failure might be related to using `on: [push, pull_request]` instead of `on: push` as in the official docs. It does make more sense to only push to TestPyPI on successful merge to the default branch. I'll try breaking these jobs into a separate file that uses the different triggering rule.

Issue #1092, opened 2022-01-18
Error parsing DLPNO single-point calculation


jvalegre:
Dear cclib developers,

I get an error when parsing DLPNO single-point calculations in ORCA v4.0 and v4.2:

"
[ORCA a_DLPNO.out ERROR] Encountered error when parsing.
[ORCA a_DLPNO.out ERROR] Last line read:              **** THE GBW FILE WAS UPDATED (/PATH/FILENAME.gbw) ****
"

where PATH is just the folder used in the HPC server where I run the calculation. Have you encountered this error before?

Best,
Juan

shivupa:
Hi Juan,
Are you able to share the offending output file?

jvalegre:
Absolutely, thanks! This file is for ORCA 4.0 but the same happens in version 4.2
[a_DLPNO.txt](https://github.com/cclib/cclib/files/7909540/a_DLPNO.txt)

PR #1369, opened 2024-02-26
Changed SCF target parsing for ORCA to be more robust

oliver-s-lee:
This change was inspired by a log file that failed to parse, but the structure is private so I cannot share I'm afraid. The error is also difficult to reproduce, it probably requires a structure that has a very difficult to converge SCF.

I'll annotate the changes to explain them, hopefully the logic makes sense on its own without a regression test.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -2798,14 +2798,14 @@ Dispersion correction           -0.016199959
         # among other things.
         while not "Last Energy change" in line:
             line = next(inputfile)
-        
+
         deltaE_value = float(line.split()[4])
         deltaE_target = float(line.split()[7])
         maxDP_value = None
         rmsDP_value = None
         maxDP_target = None
         rmsDP_target = None
-        
+
         line = next(inputfile)
         if "Last MAX-Density change" in line:
             maxDP_value = float(line.split()[4])
@@ -2817,7 +2817,7 @@ Dispersion correction           -0.016199959
             else:
                 rmsDP_value = None
                 rmsDP_target = None
-                
+
         self.scfvalues[-1].append([deltaE_value, maxDP_value, rmsDP_value])
         self.scftargets.append([deltaE_target, maxDP_target, rmsDP_target])
 
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -2793,14 +2793,19 @@ Dispersion correction           -0.016199959
     # end of parse_scf_expanded_format
 
     def _append_scfvalues_scftargets(self, inputfile, line):
-        # The SCF convergence targets are always printed after this, but apparently
-        # not all of them always -- for example the RMS Density is missing for geometry
-        # optimization steps. So, assume the previous value is still valid if it is
-        # not found. For additional certainty, assert that the other targets are unchanged.
+        # The SCF convergence targets are always printed in this next section
+        # but which targets are available depends on the SCF method in use,
+        # among other things.
         while not "Last Energy change" in line:
             line = next(inputfile)
+        
         deltaE_value = float(line.split()[4])
         deltaE_target = float(line.split()[7])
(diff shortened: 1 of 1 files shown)


cclib/parser/orcaparser.py
```diff
@@ -2810,12 +2815,11 @@ def _append_scfvalues_scftargets(self, inputfile, line):
                 rmsDP_value = float(line.split()[4])
                 rmsDP_target = float(line.split()[7])
             else:
-                rmsDP_value = self.scfvalues[-1][-1][2]
-                rmsDP_target = self.scftargets[-1][2]
```
oliver-s-lee, line 2814: This line is the cause of the error. When the RMS density target is not available at the end of the current SCF cycle, it uses the value from the previous cycle. This is not desirable for at least two reasons:
 1) There is no guarantee the RMS density target was available in the last cycle (it could be missing in the very first)
 2) The logic here doesn't make sense to me. It's certainly rare for the SCF targets to change mid-job, but there's no reason this couldn't happen. Orca has a complicated SCF solver (I think actually multiple solvers?) that likely doesn't use all of the targets all of the time. If a target isn't reported, then it seems more likely to me that it wasn't used, rather than Orca forgot to mention it (but perhaps I'm being to generous to Orca...)

oliver-s-lee:
Happily, https://github.com/cclib/cclib/issues/1092 is a duplicate of this error and has a log file

cclib/parser/orcaparser.py
berquist: Yes, any program that can switch between solvers (Orca, Q-Chem, Psi4, even Gaussian) will have this problem.

berquist (approved):
I've confirmed that the regression in https://github.com/cclib/cclib-data/pull/186 no longer gives
```console
$ ccget --list ORCA/ORCA4.2/a_DLPNO.txt
Attempting to read ORCA/ORCA4.2/a_DLPNO.txt
[ORCA ORCA/ORCA4.2/a_DLPNO.txt ERROR] Encountered error when parsing.
[ORCA ORCA/ORCA4.2/a_DLPNO.txt ERROR] Last line read:              **** THE GBW FILE WAS UPDATED (/hpcinfo/jvalegre@colostate.edu/743250/BiPyrPh_SMD_DLPNO.gbw) ****

  File "/home/eric/.pyenv/versions/miniforge3-4.10.3-10/envs/cclib_311/bin/ccget", line 8, in <module>
    sys.exit(ccget())
             ^^^^^^^
  File "/home/eric/development/forks/python/cclib_pr_review/cclib/scripts/ccget.py", line 184, in ccget
    data = ccread(filename, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eric/development/forks/python/cclib_pr_review/cclib/io/ccio.py", line 187, in ccread
    return log.parse()
           ^^^^^^^^^^^
  File "/home/eric/development/forks/python/cclib_pr_review/cclib/parser/logfileparser.py", line 167, in parse
    self.extract(self.inputfile, line)
  File "/home/eric/development/forks/python/cclib_pr_review/cclib/parser/orcaparser.py", line 550, in extract
    self._append_scfvalues_scftargets(inputfile, line)
  File "/home/eric/development/forks/python/cclib_pr_review/cclib/parser/orcaparser.py", line 2814, in _append_scfvalues_scftargets
    rmsDP_target = self.scftargets[-1][2]
                   ~~~~~~~~~~~~~~~^^^^
IndexError: list index out of range
```
but
```console
$ ccget scftargets ORCA/ORCA4.2/a_DLPNO.txt
Attempting to read ORCA/ORCA4.2/a_DLPNO.txt
[ORCA ORCA/ORCA4.2/a_DLPNO.txt WARNING] Attribute nbasis changed value (213 -> 486)
scftargets
array([[1.e-08, 1.e-07,    nan],
       [1.e-08, 1.e-07,    nan]])
```

PR #1370, opened 2024-02-26
Tests for scripts

oliver-s-lee:
Expanded and fixed tests for ccget (primarily), ccwrite and ccframe.

Added test for URLs and all available log files with ccget.

### DIFF
--- a/test/bridge/testase.py
+++ b/test/bridge/testase.py
@@ -7,6 +7,7 @@
 
 """Tests for the cclib2ase bridge in cclib."""
 
+from pathlib import Path
 
 from cclib import ccopen
 from cclib.bridge import cclib2ase
@@ -125,12 +126,12 @@ class ASETest:
         assert np.isclose(ase_data.natom, len(data.atomnos))
         assert np.isclose(ase_data.temperature, 0)
 
-    def test_write_and_read_trivial_trajectories(self):
+    def test_write_and_read_trivial_trajectories(self, tmp_path):
         """Ensure write and read trajectory files with single structures."""
         # An open-shell single point calculation.
         data = ccopen("data/ORCA/basicORCA4.2/dvb_sp_un.out").parse()
-        cclib2ase.write_trajectory("dvb_sp_un.traj", data)
-        trajdata = cclib2ase.read_trajectory("dvb_sp_un.traj")
+        cclib2ase.write_trajectory(Path(tmp_path, "dvb_sp_un.traj"), data)
+        trajdata = cclib2ase.read_trajectory(Path(tmp_path, "dvb_sp_un.traj"))
 
         assert np.allclose(trajdata.atomcoords, data.atomcoords)
         assert np.allclose(trajdata.scfenergies, data.scfenergies)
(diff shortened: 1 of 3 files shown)


test/io/testscripts.py
```diff
@@ -61,8 +64,38 @@ def test_ccread_invocation_matching_args(self, mock_warn, mock_ccread):
...
+    def test_ccread_url(self, mock_ccread) -> None:
+        self.main()
+
+    @mock.patch("cclib.scripts.ccget.sys.argv", ["ccget", "metadata", "http://fo.bar"])
+    def test_ccread_bad_url(self, mock_ccread) -> None:
+        with pytest.raises(Exception):
+            self.main()
 
-@mock.patch("cclib.scripts.ccwrite.ccwrite")
+    @pytest.mark.parametrize("file_path", gettestdata())
```
berquist, line 76: I ran `python -m pytest -k 'ccgetTest' --co` and the parameterization names aren't informative:
```console
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.11.7, pytest-7.4.4, pluggy-1.3.0
rootdir: /home/eric/development/forks/python/cclib_pr_review
configfile: pytest.ini
plugins: xdist-3.5.0, cov-4.1.0, shutil-1.7.0
collected 5504 items / 5042 deselected / 462 selected

<Package io>
  <Module testscripts.py>
    <Class ccgetTest>
      <Function test_empty_argv>
      <Function test_ccread_invocation>
      <Function test_ccread_invocation_matching_args>
      <Function test_ccread_url>
      <Function test_ccread_bad_url>
      <Function test_all[file_path0]>
      <Function test_all[file_path1]>
      <Function test_all[file_path2]>
      <Function test_all[file_path3]>
      <Function test_all[file_path4]>
      <Function test_all[file_path5]>
      <Function test_all[file_path6]>
      <Function test_all[file_path7]>
...
```
The `ids` argument is how you change the name, one for each generated test case (https://docs.pytest.org/en/latest/example/parametrize.html#different-options-for-test-ids).  You can see an example at https://github.com/cclib/cclib/blob/6832e4656c3baca0f4804a183221f7b01957dddb/test/conftest.py#L392 (the arguments to the function/method are the same as to the decorator).

My other concern is that this is re-parsing 457 files, which seems like overkill, and many of them are spewing warnings that are not useful in this scenario.  Can we achieve the same effect with a subset of the files?

berquist (changes requested):
Looks good but some questions about `test_all`.

test/io/testscripts.py
oliver-s-lee: Cool, updated the IDs and reduced the tests to just one per parser version (~50 in total).

The logging is still an issue tho. Due to the way we construct a different logger for each log file it's very difficult to silence them properly, which probably suggests we're setting up the logger wrong. For this PR I would suggest leaving as is and then I'll create a separate issue for the logging setup?

berquist: Yes, that's fine.

berquist (changes requested):
The parameterization ids look good now, just a minor nit remains

test/bridge/testase.py
```diff
@@ -125,12 +126,12 @@ def test_makease_works_with_closedshells(self):
...
 
-    def test_write_and_read_trivial_trajectories(self):
+    def test_write_and_read_trivial_trajectories(self, tmp_path):
         """Ensure write and read trajectory files with single structures."""
         # An open-shell single point calculation.
         data = ccopen("data/ORCA/basicORCA4.2/dvb_sp_un.out").parse()
-        cclib2ase.write_trajectory("dvb_sp_un.traj", data)
-        trajdata = cclib2ase.read_trajectory("dvb_sp_un.traj")
+        cclib2ase.write_trajectory(Path(tmp_path, "dvb_sp_un.traj"), data)
+        trajdata = cclib2ase.read_trajectory(Path(tmp_path, "dvb_sp_un.traj"))
```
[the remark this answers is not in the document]
oliver-s-lee: Yeah NP, I just tracked it down after having those weird files appear out of nowhere

berquist (approved):

PR #1371, opened 2024-02-26
Added functionality to read zipped cubes

oliver-s-lee:


cclib/parser/logfilewrapper.py
```diff
@@ -250,6 +262,12 @@ def readline(self) -> str:
         """
         return next(self)
 
+    def readlines(self) -> str:
```
berquist, line 265:
```suggestion
    def readlines(self) -> List[str]:
```

test/method/testvolume.py
```diff
@@ -104,3 +104,8 @@ def test_roundtrip_cube(self):
         density_recovered = volume.read_from_cube("coarsewater.cube")
 
         assert_allclose(density.data, density_recovered.data, rtol=0.05)
+
+    def test_zip_cube(self):
+        """Check we can read from a zipped file."""
+        data = volume.read_from_cube(f"{os.path.dirname(os.path.realpath(__file__))}/co.cube.zip")
```
berquist, line 110: We should use `pathlib.Path` in as many places as possible:
```suggestion
        data = volume.read_from_cube(Path(__file__).resolve().parent / "co.cube.zip")
```

berquist (changes requested):
Minor nits

test/method/testvolume.py
oliver-s-lee: I'm a pathlib fanboy so no complaints from me

berquist (approved):
It's awesome that it was basically as simple as reusing `FileWrapper`.

PR #1373, opened 2024-03-01
Calculate version from tags and `versioningit`

berquist:


berquist:
I'm going to disable the regression tests that pass with `master` but fail on the release branch.

berquist:
Actually, I'm not sure how that will work trivially.  We probably need to follow the same release process for cclib-data as we do here...

PR #1383, opened 2024-03-12
Tree for parsing

shivupa:
This is the tree implementation for cclib2 parsing


cclib/notes.md
amandadumi, line 1: is this an intentional add?

shivupa: oops

cclib/tree/tree.py
amandadumi, line 1: would it be safe to remove commented code or is that there for reference as we develop?

amandadumi (approved):
perfect, thanks!

PR #1389, opened 2024-03-18
Merge release 1.8.1 branch back into default branch

berquist:


berquist:
I don't think this has happened before, need to figure out the best way to do it.  The simplest would be to merge `master` into the release branch, but that isn't an option because the release branch should remain pure even after the tag so that backported fixes could be allowed.

berquist:
Closing in favor of https://github.com/cclib/cclib/pull/1390. I think the solution is to create a new merge branch that contains the contents of both the release and default branches.

PR #1407, opened 2024-04-09
Porting test for version 2

amandadumi:
This will begin porting tests. So far I am just focusing on changing how we access/pass the new data objects into tests.  This isn't ready yet, but figured it would be nice to have a draft as reference.

- [x] change skip feature to check for parsed program in driver
- [ ] change each test to access the object within the driver (in progress)
- [x] add missing parsers needed for test_SP (in progress, but likely should happen in a separate PR.)
- [ ] reroute so only a ccdata object is passed instead of the driver: I don't love the idea of having to pass the driver instead of just a ccdata object. I'll  clean that up, but this was just quick to get up and running to start fielding some issues.


amandadumi:
This fails in the ` -> dict | None` type hinting as using the [union operator](https://docs.python.org/3/library/stdtypes.html#types-union) was only introduced in 3.10.  I could replace these all with `or` and it will work in both new and old versions. Is this an acceptable fix?

berquist:
My personal preference is for `Optional[Dict]`

amandadumi:
Okay great, I forgot this existed and I like it better too. Thanks

amandadumi:
Just updating here, I  decided to go about turning off regression tests  the bash script, then adding just the files for which I've ported the test to the list in pytest.ini. 

I rebased to include the QChem parser. I included those test files, but haven't actually gone through to see why the tests are failing.

amandadumi:
currently enabling only SP, SPun, and Basis test files, which cover most of the attributes that have been ported. I've turned off the methods testing and regression testing through commenting the run lines out int the scripts. I could remove it from the .github/workflows/test_and_package.yml  if preferred.

PR #1412, opened 2024-04-28
Always compute freeenergy if not present and required attributes are available

berquist:
First part of https://github.com/cclib/cclib/issues/1288.

PR #1416, opened 2024-04-28
Jaguar: implement entropy, enthalpy, and freeenergy

berquist:


berquist:
This will probably fail coverage because of https://github.com/cclib/cclib/pull/1412#issuecomment-2081321656.

PR #1438, opened 2024-06-15
Merge master into main

berquist:


berquist:
Things I changed:
- Atomic units in new parsing code for SCF and MO energies
- Put all the tests back in `testdata` but commented them out
- Disabled some ~new metadata tests
- Temporarily disabled the docs build (see #1439)

PR #1441, opened 2024-06-17
Add tests for tree, Psi4 BSSE example, and code-in-code example

berquist:
There are a bunch of things I don't understand about the behavior of the implementation right now, mostly marked as TODO in the tests, particularly around `get_next_index`.


amandadumi (approved):
This all looks good to me. Thanks for removing the unintentional prints that were left in. The changes in the tree implementation look reasonable to me. 

If you want another pair of eyes on get_next_index, let me know as I also am not all that familiar with the tree functions yet.

berquist:
If `get_next_index` really is supposed to be a DFS, then the results aren't right for the more complex trees.  I didn't want to address that in this PR but in a follow-up, maybe combined with the automatic tree generation.

shivupa (approved):

PR #1443, opened 2024-06-20
Apply pyupgrade and fix ruff lints

### LINKED ISSUE #1313: Enable ruff lints and fix them
A follow up and more systematic approach to #1303.

This is, at least initially, a catch-all for cleaning up the codebase.

Deleting this line https://github.com/cclib/cclib/blob/2193e9b3db0aa9a9b391cf4ea69f007565f4651a/pyproject.toml#L18 reverts to the default set of linting rules, which is (https://docs.astral.sh/ruff/configuration/)
```toml
select = ["E4", "E7", "E9", "F"]
```
which enables all rules beginning with the given prefix.  Above, this is all rules originating pyflakes and some originating from pycodestyle.

All available rules are at https://docs.astral.sh/ruff/rules/.  Beginning with the default rules seems like an ok starting place.

berquist:
Closes #1313

This now runs clean, and enforces a lot of default flake8 and pylint checks using ruff.
```diff
diff --git pyproject.toml pyproject.toml
index 837c3b4c..afa3293b 100644
--- pyproject.toml
+++ pyproject.toml
@@ -86,13 +86,11 @@ sections = ["FUTURE", "STDLIB", "FIRSTPARTY", "THIRDPARTY", "LOCALFOLDER"]
 [tool.ruff]
 line-length = 100
 src = ["cclib"]
+target-version = "py37"

 [tool.ruff.format]
 skip-magic-trailing-comma = true

-[tool.ruff.lint]
-select = []
-
 [tool.ruff.lint.isort]
 lines-after-imports = 2
 section-order = ["future", "standard-library", "first-party", "third-party", "local-folder"]
```

Most of the `noqa` ignores are for unused variables that contain data we may want to parse in the future.  Some of the unused variables I was able to either delete or, if they were the result of a `next(inputfile)`, convert that into a `skip_line` or `skip_lines` call.

Despite the name of the branch, I decided to hold off on combining more type annotations into this PR.

I'll add commit hashes to `.git-blame-ignore-revs` in a followup PR for both the lint fixes and the type annotations.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -125,14 +125,17 @@ repos:
       - id: check-yaml
       - id: check-toml
       - id: check-added-large-files
-      - id: fix-encoding-pragma
-        args: [--remove]
       - id: no-commit-to-branch
         args: [--branch, main, --branch, master]
   - repo: https://github.com/Mateusz-Grzelinski/actionlint-py
     rev: "v1.7.1.15"
     hooks:
       - id: actionlint
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.16.0
+    hooks:
+      - id: pyupgrade
+        args: [--py37-plus]
   - repo: https://github.com/PyCQA/isort
     rev: "5.13.2"
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -143,9 +143,9 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.6.1
     hooks:
+      - id: ruff-format
       - id: ruff
         args: [--fix]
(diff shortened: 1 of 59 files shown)


shivupa (commented):
This replaces a ton of fstrings. Do we want that? I am ok with that if so.

berquist:
There is a reason for converting these back from f-strings, specifically for logging calls.  If your logging level is below what would be sent to the log, the arguments passed to the call are not evaluated.  However, if you pass an f-string, the stringification will happen no matter what.  This technically isn't a problem if the arguments you pass have cheap `__str__` calls, but it's good practice still.

berquist:
I will tag 2.0a2 on `main` once this is merged to `master` and `master` is merged to `main`.

PR #1450, opened 2024-06-29
Capitalize first letter of all attribute class names

berquist:
Exactly what it says on the tin.  I think we should do this for safety reasons, so there's no confusion about having both a variable and a class name called `scfenergies`, for example.

We can also use this opportunity to change more than just the first letter for some of the more awkward names, but I don't know what makes sense, so I'd like some input.

### DIFF
--- a/cclib/attributes/attribute.py
+++ b/cclib/attributes/attribute.py
@@ -36,10 +36,10 @@ class _Attribute(ABC):
             else:
                 return self._impl
 
-    def __str__(self):
+    def __str__(self) -> str:
         return str(self.__class__.__name__)
 
-    def typecheck(self):
+    def typecheck(self) -> None:
         """Check the types of all attributes.
 
         If an attribute does not match the expected type, then attempt to
@@ -61,7 +61,7 @@ class _Attribute(ABC):
             )
 
 
-class aonames(_Attribute):
+class Aonames(_Attribute):
     """atomic orbital names (list of strings)"""
 
     def __init__(self, *args, **kwargs):
@@ -70,7 +70,7 @@ class aonames(_Attribute):
         )
 
 
-class aooverlaps(_Attribute):
+class Aooverlaps(_Attribute):
     """atomic orbital overlap matrix (array[2])"""
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


cclib/attributes/attribute.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, *args, **kwargs):
         )
 
 
-class atombasis(_Attribute):
+class Atombasis(_Attribute):
```
berquist, line 82: `AtomBasis`?

cclib/attributes/attribute.py
```diff
@@ -61,7 +61,7 @@ def typecheck(self):
             )
 
 
-class aonames(_Attribute):
+class Aonames(_Attribute):
```
berquist, line 64: `AoNames`? `AONames`?

cclib/attributes/attribute.py
```diff
@@ -448,7 +448,7 @@ def __init__(self, *args, **kwargs):
         )
 
 
-class nbasis(_Attribute):
+class Nbasis(_Attribute):
```
berquist, line 451: `NBasis`?

berquist (commented):
I'm personally ok with leaving things as just the first letter capitalized, but so we can see what a few of them might look like...

cclib/attributes/attribute.py
amandadumi: I like that first letter leaves no real question and can see this being easy to remember as developing. Obviously AONames makes the most  sense in terms of being clear, so open to that too.

amandadumi (commented):
Looks good! I agree with the change right now being first letter only. My heart is split between wanting consistency between classes and the more sensible capitalization options you presented. 

I was wondering if the class capitalization is needed to access the parsed attribute from the user perspective?

amandadumi (approved):

berquist:
I do also prefer just the first letter.  Anything else will be difficult to make consistent and have everyone be happy.

> I was wondering if the class capitalization is needed to access the parsed attribute from the user perspective?

I don't know what this looks like yet, but we can dynamically change this if need be by introspecting the class name.

berquist:
Waiting to hear from @shivupa and @langner since this is a major change.

I do expect that users will be exposed to this in some way.

shivupa:
> Waiting to hear from @shivupa and @langner since this is a major change.
> 
> I do expect that users will be exposed to this in some way.

Sorry for the delay. I was out of the office for a bit and swimming under work for the other bit... 
I also really like just adopting the capital first letter as a rule even though some cases like `AONames` would warrant other capitalization schemes. Capital first letter lets us just not think about it.

PR #1452, opened 2024-07-04
Gaussian: fix parsing of platform line

### LINKED ISSUE #1420: Error Getting platform_full_version for Apple M1 Computer
Trying to parse Gaussian log files generated on an Apple M1 Computer generates an error. 

```
_, _, platform_full_version, compile_date = next(inputfile).split()
ValueError: too many values to unpack (expected 4)
```

This is because the computer name is two words in the Gaussian log file. On the line after the citation, shown below, splitting the line results in 5 words instead of the expected 4
```
Gaussian 16:  Apple M1-G16RevC.02  7-Dec-2021
```

Error is not observed if the line is manually corrected to: (note removal of space between Apple and M1)

```
Gaussian 16:  AppleM1-G16RevC.02  7-Dec-2021
```

Is it possible to get a fix for this, rather than having to adjust all files manually?

berquist:
Closes #1420

https://github.com/cclib/cclib-data/pull/190 should be merged after this PR.

### DIFF
--- a/cclib/parser/gaussianparser.py
+++ b/cclib/parser/gaussianparser.py
@@ -67,8 +67,18 @@ class Gaussian(logfileparser.Logfile):
     }
 
     def before_parsing(self):
+        # Examples:
+        #  Gaussian 16:  Apple M1-G16RevC.02  7-Dec-2021
+        #  Gaussian 16:  ES64L-G16RevC.01  3-Jul-2019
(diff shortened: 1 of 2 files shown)


berquist:
This is failing because the regression testing code no longer allows defining a test that's missing a file.  Since this is a common scenario with PRs made to both repos simultaneously, I could change it back to the previous behavior...or we merge cclib-data PRs with the understanding that the next cclib PR has to be the regression fix.

Issue #1448, opened 2024-06-26
[XTB] Fails to parse double letter elements coordinates


alkorolyov:
Hi,

I have found that when running geometry optimization of molecules with double letter elements `Cl`, `Br`, `Si` etc, the XTB parser skipps them when parsing atomcoords.

Step to reproduce

coord file:
```
5

C     -0.183852   -0.000158   -0.003151
Br     1.750781    0.001502    0.030011
H     -0.507476   -0.616006   -0.843507
H     -0.521209    1.030189   -0.124698
H     -0.538243   -0.415528    0.941345
```

run:
```
xtb coord --opt > output
```

Part of the output corresponding to final structure:
```
================
 final structure:
================
5
 xtb: 6.6.1 (8d0f1dd)
C           -0.18542799765754       -0.00015906836608       -0.00317867081557
Br           1.75174521992695        0.00150217653496        0.03002782385518
H           -0.50735322186249       -0.61262621542538       -0.83892126122008
H           -0.52101203571700        1.02453063910753       -0.12405865892429
H           -0.53795096468993       -0.41324853185103        0.93613076710476
```

when parsing output by cclib:
```
cclib.io.ccread('output').atomcoords

array([[[-1.85427998e-01, -1.59068268e-04, -3.17867056e-03],
        [-5.07353222e-01, -6.12626216e-01, -8.38921262e-01],
        [-5.21012036e-01,  1.02453064e+00, -1.24058659e-01],
        [-5.37950964e-01, -4.13248532e-01,  9.36130767e-01]]])

cclib.io.ccread('output').atomnos

array([ 6, 35,  1,  1,  1], dtype=int32)
```

Versions of the packages:
```
xtb 6.6.1
ccllib 1.8.1
```

shivupa:
Hi thanks for reporting this! Could you upload the output file so we can replicate this + fix this?

alkorolyov:
Sure, please find attached the corresponding [output.txt](https://github.com/user-attachments/files/15991156/output.txt)

Andrew-S-Rosen:
The bug is in the following line:
https://github.com/cclib/cclib/blob/ee91131e9c278081d1ff88fa763d3c2b5e501da6/cclib/parser/xtbparser.py#L437

```python
a="Br     1.750781    0.001502    0.030011"
a.split()[0].isupper()
>>> False
```

berquist:
@shivupa Please make the PR with the output to cclib-data.

PR #1463, opened 2024-07-25
Fix parsing of elements in xtb parser

Andrew-S-Rosen:
Closes https://github.com/cclib/cclib/issues/1448 by fixing the parsing of double-letter element names.

Naturally, a test should be added, but I don't have the capacity to do so at this time. Any external help would be greatly appreciated!


shivupa (approved):

shivupa:
We can add the file from #1448 as a regression test and I am ok with this

berquist (approved):

Issue #1460, opened 2024-07-25
versioningit pain


berquist:
versioningit has been a source of pain since I introduced it (https://github.com/cclib/cclib/pull/1373).

What the package does, as part of https://github.com/cclib/cclib/blob/ee91131e9c278081d1ff88fa763d3c2b5e501da6/pyproject.toml#L68, is to determine the version of the working copy of the source code from the local Git history and make it so `cclib.__version__` works.

This operation does not work as currently implemented in cclib when one has a copy of the source code that is missing a full `.git` directory as discovered in #1444.  `cclib/_version.py` is written when doing a source or wheel build in a Git checkout, but the GitHub "Source code" entries do not have it, and the source/wheel build would also fail when `.git` is missing.

The need to generate this file also created cryptic errors when devs where using an existing venv with an editable install.  In fact, in order to always have the correct version available, one would always have to be reinstalling after every commit.  (Though, the need for the version like this is tiny.)

`nix build` currently produces, once I add `versioningit` and `git`,
```
python3.11-cclib> Sourcing python-remove-tests-dir-hook
python3.11-cclib> Sourcing python-catch-conflicts-hook.sh
python3.11-cclib> Sourcing python-remove-bin-bytecode-hook.sh
python3.11-cclib> Sourcing pypa-build-hook
python3.11-cclib> Using pypaBuildPhase
python3.11-cclib> Sourcing python-runtime-deps-check-hook
python3.11-cclib> Using pythonRuntimeDepsCheckHook
python3.11-cclib> Sourcing pypa-install-hook
python3.11-cclib> Using pypaInstallPhase
python3.11-cclib> Sourcing python-imports-check-hook.sh
python3.11-cclib> Using pythonImportsCheckPhase
python3.11-cclib> Sourcing python-namespaces-hook
python3.11-cclib> Sourcing python-catch-conflicts-hook.sh
python3.11-cclib> Sourcing pytest-check-hook
python3.11-cclib> Using pytestCheckPhase
python3.11-cclib> Running phase: unpackPhase
python3.11-cclib> unpacking source archive /nix/store/y6d5r9yc43yx3crc80crqc9p9bg1g71b-source
python3.11-cclib> source root is source
python3.11-cclib> setting SOURCE_DATE_EPOCH to timestamp 315619200 of file source/tools/versioningit_override.py
python3.11-cclib> Running phase: patchPhase
python3.11-cclib> Running phase: updateAutotoolsGnuConfigScriptsPhase
[...]

PR #1467, opened 2024-08-11
Fix Nix build and add to CI

berquist:
Part of https://github.com/cclib/cclib/issues/1460

There is not a universal way of dealing with `versioningit` in Nixpkgs, but it seems like a common solution is to bypass it altogether, so I combined https://github.com/NixOS/nixpkgs/blob/0d8f17feeaf68d76bc7ffe69be358e9f01b81c93/pkgs/development/python-modules/scramp/default.nix#L39 with the idea from https://github.com/NixOS/nixpkgs/blob/0d8f17feeaf68d76bc7ffe69be358e9f01b81c93/pkgs/development/python-modules/limits/default.nix#L53.

Waiting to update the flake lockfile until the `gau2grid` part of https://github.com/Nix-QChem/NixOS-QChem/issues/513 is resolved, which it sounds like they've already started on.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -24,6 +24,7 @@ jobs:
           name: nix-qchem
       # Checkout of the current head in the working dir
       - uses: actions/checkout@v4
-      # Build the nix package
-      - name: Build nix package
+      - name: Check nix flake
         run: nix flake check -L
+      - name: Build nix package
+        run: nix build -L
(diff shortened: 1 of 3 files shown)


berquist:
I need to change the version in the Nix package (and likely `pyproject.toml` too) to something that's between 1.8.1 and 2.0 to deal with the breaking changes...

PR #1481, opened 2024-09-17
PySCF to cclib bridge support

oliver-s-lee:
Ok, this is now working to the point where it's ready for at least an initial review.

The main workhorse is `cclib.bridge.cclib2pyscf.cclibfrommethods()` which accepts a number of PySCF method objects (SCF, MP, excited states etc.) and returns a ccData object.

`cclib.bridge.cclib2pyscf.makecclib()` is a slightly nicer wrapper on top that does a good job of guessing what type of PySCF method you pass to it.

Supported so far are:
 - aonames
 - aooverlaps
 - atombasis
 - atomcharges (mulliken and lowdin)
 - atomcoords
 - atommasses
 - atomnos
 - ccenergies (including CCSD(T) although this is slightly fudged due to the way it's handled in PySCF)
 - charge
 - coreelectrons
 - etenergies
 - etoscs
 - etdips
 - etsecs
 - etsyms (but only multiplicities for now)
 - gbasis
 - geovalues (but only energies, no targets yet)
 - homos
 - metadata (most of important things are here, no performance or timing data yet tho)
 - mocoeffs
 - moenergies
 - moments (both dipole and quadrupole, although quad are only available in development versions of PySCF)
 - mosyms
 - mpenergies (MP2)
 - mult
 - natom
 - nbasis
 - nmo
 - optdone (although slightly hacked due to the way PySCF handles optimisations)
 - optstatus (see above)
 - rotconsts
 - scfenergies (HF and DFT)
 - scftargets
 - scfvalues
 - vibfreqs
 - vibfconsts
 - vibirs
 - vibrmasses

Not currently supported but should be available with only a bit more work
- etdips and friends (just forgot about them)
- grads (not looked for yet)
- hessian (this is readily available, but I'm not knowledgeable enough to understand either the PySCF or cclib formats)
- vibdisps (the IR part of PySCF is not so well documented and I'm not sure where this is hidden)

There are also tests for all of the supported properties (unless I've forgotten of course). The tests use the normal parser tests rather than the ones in `bridge`. To get this to work, I've slightly modified `conftest.py` so `testdata` can specify python files which get imported, run, and the results cached as necessary. Most of the PySCF calcs are extremely quick, but the frequencies on DVB can take a few minutes. Not sure if we want to look at saving the raw PySCF data rather than regenerating each time...


berquist:
You can get rid of the F821 warnings by prefixing the variable with `_`.

oliver-s-lee:
At the moment this PR parses energies into hartree, but that change is (or at least should be) in a single commit so easy to reverse if need be.

berquist:
> At the moment this PR parses energies into hartree, but that change is (or at least should be) in a single commit so easy to reverse if need be.

Ok; sorry to leave you hanging for a review on this.

cclib/bridge/cclib2pyscf.py
```diff
@@ -112,4 +129,491 @@ def makepyscf_mos(ccdata, mol):
...
+        # PySCF tracks convergence for each state which is great.
+        attributes["metadata"]["success"] = all(
+            itertools.chain(*(etmethod.converged for etmethod in et))
+        )
+
+        # In cclib 1.x, 'energies' are actually expected to be cm-1. In 2.x, this will change to Hartree.
+        attributes["etenergies"] = list(itertools.chain(*(etmethod.e for etmethod in et)))
+        attributes["etoscs"] = list(
+            itertools.chain(*(etmethod.oscillator_strength(gauge="length") for etmethod in et))
+        )  # or do we want velocity?
```
berquist, line 504: Yes, length gauge, except for `transprop` where we can store both.

cclib/bridge/cclib2pyscf.py
```diff
@@ -112,4 +129,491 @@ def makepyscf_mos(ccdata, mol):
...
+        # et.analyse() prints real symmetries, so they must be available somewhere...
+
+        attributes["etsyms"] = []
+        for etmethod in et:
+            if isinstance(etmethod, pyscf.tdscf.rks.TDA):
```
berquist, line 509: I'd replace all this with a dict that maps `pyscf` class to the string, and on the dict use `.get(etmethod, type(etmethod).__name__)` to simultaneously handle the default case.

cclib/bridge/cclib2pyscf.py
```diff
@@ -112,4 +129,491 @@ def makepyscf_mos(ccdata, mol):
...
+            if etmethod.singlet and mol.multiplicity != 2:
+                et_mult = "Singlet"
+
+            elif etmethod.singlet:
+                et_mult = "Doublet"
+
+            else:
+                et_mult = "Triplet"
+
+            attributes["etsyms"].extend([et_mult for i in range(len(etmethod.e))])
```
berquist, line 539:
```suggestion
            attributes["etsyms"].extend([et_mult for _ in range(len(etmethod.e))])
```

cclib/bridge/cclib2pyscf.py
```diff
@@ -112,4 +129,491 @@ def makepyscf_mos(ccdata, mol):
...
+            x, y = list(itertools.chain(*(etmethod.xy for etmethod in et)))[index]
+
+            if not scf.istype("UHF"):
+                # The coefficients of x (and presumably also y) are normalised to 0.5,
+                # but we expect 1.0
+                # Renormalize.
+                # Taken from pyscf.tdscf.rhf.get_nto()
+                #
+                # Would appreciate someone checking this makes sense?
+                x *= 1.0 / np.linalg.norm(x)
```
berquist, line 559: IIRC I ripped this from pyscf many years ago, but it indicates that for CIS/TDA you're correct, but if there are deexcitation vectors you need to normalize them together:
```python
    def norm_xy(z: np.ndarray, nocc: int, nvirt: int) -> Tuple[np.ndarray, np.ndarray]:
        x, y = z.reshape(2, nvirt, nocc)
        norm = 2 * (np.linalg.norm(x) ** 2 - np.linalg.norm(y) ** 2)
        norm = 1 / np.sqrt(norm)
        return (x * norm).flatten(), (y * norm).flatten()
```

cclib/bridge/cclib2pyscf.py
```diff
@@ -112,4 +129,491 @@ def makepyscf_mos(ccdata, mol):
...
+                    attributes["etsecs"][index].append(
+                        [(occupied, 1), (nocc[1] + virtual, 1), x[1][occupied, virtual]]
+                    )
+
+    # Hessian/frequencies
+    if hess:
+        pass
+        # TODO: Don't know enough to work out what this is
+        # cclib wants a rank 2 array, pyscf gives us a rank 4 array?
+        # Dimensions appear to be natoms x natoms x 3 x 3
```
berquist, line 595: If true it would then require a reorder to `(3 x natoms, 3 x natoms)`.

cclib/bridge/cclib2pyscf.py
```diff
@@ -112,4 +129,491 @@ def makepyscf_mos(ccdata, mol):
...
+        # attributes["hessian"] = hess.de
+
+    if freq:
+        # Freq data, a dict with 'freq_error', 'freq_au', 'freq_wavenumber', 'norm_mode', 'reduced_mass',
+        # 'vib_temperature', 'force_const_au', 'force_const_dyne'
+        # TODO: This can include imaginary numbers, convert to 'negative' frequencies?
```
berquist, line 601: Yes

berquist (changes requested):
Here is a partial review.

So far I agree with all of your assumptions that I didn't comment on.

cclib/bridge/cclib2pyscf.py
oliver-s-lee: Sadly we can't do that because then we wouldn't be checking class inheritance (I think the UHF methods base class of these or something). We could iterate over each dictionary key and check but IMO that's no more elegant than the current implementation?

cclib/bridge/cclib2pyscf.py
oliver-s-lee: Ah thanks, TD was originally missed because the DVB calculation wouldn't converge for whatever reason, I've now added a new test for TD with water to test the normalisation with both x and y. The implementation is now:

```python
def norm_xy(x, y):
    norm_factor = 1.0 / np.sqrt(
        np.linalg.norm(x) ** 2 - np.linalg.norm(y) ** 2
    )
    return (x * norm_factor, y * norm_factor)
```

This logic is beyond me really, but the `2 *` multiplication factor in your function was giving the wrong results?

cclib/bridge/cclib2pyscf.py
oliver-s-lee: Might have to leave this for now, the reshape is easy enough but we don't currently have a test for hessian values so I can't verify if they're correct. There are probably also some unit conversion issues (I'm unsure what the units for cclib or PySCF are...)

cclib/bridge/cclib2pyscf.py
berquist: Ah, I missed that.  This is the way to do it then.

cclib/bridge/cclib2pyscf.py
berquist: Yes, I'm perfectly fine with leaving TODOs.

berquist (changes requested):

cclib/bridge/cclib2pyscf.py
berquist: This is fine for now but I'm going to try and understand this.

test/testdata
```diff
@@ -45,6 +45,7 @@ Basis     Psi4        GenericBasisTest      basicPsi4-1.7         dvb_sp_rhf.out
 Basis     Psi4        GenericBasisTest      basicPsi4-1.7         dvb_sp_rks.out
 Basis     QChem       GenericBasisTest      basicQChem5.1         dvb_sp.out
 Basis     QChem       GenericBasisTest      basicQChem5.4         dvb_sp.out
+Basis     PySCF       GenericBasisTest      basicPySCF2.6         dvb_sp.py
```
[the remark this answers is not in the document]
oliver-s-lee: Yep no problem.

Personally I'm a YAML fanboy so my default position is yes,  but in this case the table format is actually pretty useful and it would be a shame to lose that. Maybe an argument for the condensed list form?
Eg
```
- [Basis, Psi4,  GenericBasisTest, basicPsi4-1.7, dvb_sp_rks.out]
- [Basis, QChem, GenericBasisTest, basicQChem5.1, dvb_sp.out]
...
```

Or maybe could nest? Something like:
```
Basis:
  Psi4:
    - [GenericBasisTest, basicPsi4-1.7, dvb_sp_rks.out]
  QChem:
    - [GenericBasisTest, basicQChem5.1, dvb_sp.out]
```

test/data/testTDun.py
```diff
@@ -8,6 +8,8 @@
 from skip import skipForLogfile, skipForParser
 
 
+# The Gaussian log files for this test are a normal restricted calculation,
+# is this class misnamed?
```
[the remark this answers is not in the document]
oliver-s-lee: Yeah I'll be honest this did confuse me for a sec, funny how things can slip through. I wonder if the intention was originally to do a delta-SCF type calculation (the original file name in G03 was apparently deltasym.log) and maybe this was the ground state? Just speculation.

That Turbomole one was probably me I'd guess so woops there too. Will address in another PR.

test/conftest.py
```diff
@@ -294,13 +295,38 @@ def data(request) -> ccData:
     files = request.param
     first = files[0]
     if first not in _CACHE:
-        logfile = ccopen([str(f) for f in files], future=True)
-        data = logfile.parse()
-        filenames = logfile.filename
+        # For 'normal' log files we use ccopen to parse.
+        # For pseudo parsers (like PySCF) we use a different mechanism.
+        if "PySCF" in str(first):
```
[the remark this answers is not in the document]
oliver-s-lee: Yeah a bit of a mess right now. This 'bridge' does feel quite a lot like a parser, but then again we might want an actual PySCF parser (from the PySCF log files) in the future in which case this guy would need another name or something

berquist:
This is good for now but am currently rebasing to purge all the pre-commit autofix commits.

pyproject.toml
```diff
@@ -60,6 +63,10 @@ test = [
     "coverage",
     "pytest-cov",
     "pyyaml",
+    "pyscf",
+    "pyscf-properties @ git+https://github.com/pyscf/properties",
+    "pyberny",
+    "geometric"
```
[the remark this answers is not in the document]
berquist: This is driving me crazy. I would like to push https://github.com/njzjz/openbabel-wheel/issues/6 and https://github.com/openbabel/openbabel/issues/2408 as a solution but not sure I have the energy.

berquist (approved):
This looks great.  Thanks for all your hard work on it.

PR #1495, opened 2024-10-20
Nix: update flake.lock and reenable build in CI

berquist:


berquist:
Struggling to use the cache locally (https://github.com/Nix-QChem/NixOS-QChem/issues/568) and it looks like our CI setup isn't using it properly either.  We should be able to build the package without building libtensor...

PR #1517, opened 2025-01-08
Initial support for Orca 6

oliver-s-lee:
All major features and tests are present, the only major outstanding issue I'm aware of is possibly missing metadata parsing for solvents. I believe @FNTwin was working on this along with some other bits and pieces, I would suggest we merge this in later once they're ready.


oliver-s-lee:
The two test failures are from other parsers, so are hopefully unrelated to this PR.

FNTwin:
@oliver-s-lee Yes sorry for delaying you guys, I got a bit sidetracked with life things. I'll try to wrap up my end this end of the week

berquist:
I'll hold off on the review then, but want to ask if anyone did https://github.com/cclib/cclib/issues/555#issuecomment-410431978 /  https://cclib.github.io/development.html#adding-a-new-program-version (I'll fix the list formatting there...)

oliver-s-lee:
Ah no not yet, I'll do it now

oliver-s-lee:
Is done, updated some bridge and io test that were using ORCA 4.x files too. For some reason I had to slightly reduce the precision on one of the ASE tests (it was still very close)

berquist:
> Ah no not yet, I'll do it now

I was also referring to https://cclib.github.io/development.html#adding-a-new-program-version.  (This wasn't clear enough before, hence https://github.com/cclib/cclib/pull/1532.)  IIRC v6 is different enough that this is important.  Since adding these files is tightly tied into the branch history, I can try doing this once the PR is "done".

test/data/testTD.py
```diff
@@ -296,6 +299,13 @@ def testsyms(self, data) -> None:
         pass
 
 
+class Orca6ROCISTest(OrcaROCISTest):
+    # The energies are different in 5 Vs 6 for some reason I can't fathom,
```
berquist, line 303:
```suggestion
    # TODO The energies are different in 5 Vs 6 for some reason I can't fathom,
```

berquist (changes requested):
Looking good so far, especially in the code.  The one things that needs to change is that all the files marked as v6.1 are actually v6.0 (6.0.1).

berquist (approved):

Issue #1518, opened 2025-01-10
NWChem 7.2.3 Mulliken Charges Not Parsed by cclib


lsvvt:
**Description**  
While using NWChem version 7.2.3 to compute the Mulliken charges on a simple water molecule, cclib (v1.8.1) fails to parse the `atomcharges`. The output file contains the relevant data (including `mulliken` in the `dft` section), yet cclib reports that no charges can be extracted.

---

**Steps to Reproduce**  
1. Run the following NWChem input (simplified water molecule):
   ```
   start water_calc

   geometry units angstrom nocenter noautosym noautoz
     O  0.0000000000000000e+00  0.0000000000000000e+00  1.1926200000000001e-01
     H  0.0000000000000000e+00  7.6323900000000000e-01 -4.7704700000000000e-01
     H  0.0000000000000000e+00 -7.6323900000000000e-01 -4.7704700000000000e-01
   end

   charge 0

   basis noprint
     * library def2-TZVP
   end

   dft
     xc PBE0
     mult 1
     mulliken
   end

   task dft energy
   ```
2. Obtain the NWChem output file (example attached as [test.log](https://github.com/user-attachments/files/18379889/test.log)).
3. Attempt to parse using cclib:
   ```bash
   $ ccget atomcharges test.log
   Attempting to read test.log
   Could not parse atomcharges from this file.
   ```

---

**Expected Behavior**  
- cclib should successfully read and parse the Mulliken charges (under `atomcharges`) from NWChem 7.2.3 output.
   ```bash
   $ ccget atomcharges test.log
   Attempting to read test.log
   atomcharges
   {'mulliken': array([-0.63,  0.32,  0.32])}
   ```

**Actual Behavior**  
- cclib fails to detect the Mulliken charges, displaying:
  ```bash
  $ ccget atomcharges test.log
  Attempting to read test.log
  Could not parse atomcharges from this file.
  ```

---

**Software Versions**  
- **NWChem:** 7.2.3  
- **cclib:** 1.8.1  
- **Python:** 3.13.0
- **Operating System:** CentOS Stream 8

berquist:
Done in https://github.com/cclib/cclib/pull/1519

PR #1519, opened 2025-01-10
add NWChem 7.2.3 Mulliken charges

lsvvt:
Fix for https://github.com/cclib/cclib/issues/1518

### DIFF
--- a/cclib/parser/nwchemparser.py
+++ b/cclib/parser/nwchemparser.py
@@ -801,19 +801,51 @@ class NWChem(logfileparser.Logfile):
         #    1 C    6     6.00   1.99  1.14  2.87
         #    2 C    6     6.00   1.99  1.14  2.87
         # ...
+        #
+        # If `print "mulliken ao"` is requested, after the section header the
+        # breakdown is different:
+        #
+        # Bfn.    Population  Atom+Function
+        # ---------------------------------
+        #   1      1.997265     1 C  s
+        #   2      0.743630     1 C  s
+        #   3      0.691045     1 C  px
+        #   4      0.687280     1 C  py
+        #
+        # followed by the "Shell Charges" section, which means the
+        # Atom+Function block can be skipped.  TODO Use this section when
+        # available, since it has higher precision.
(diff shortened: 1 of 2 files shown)


cclib/parser/nwchemparser.py
```diff
@@ -791,6 +791,30 @@ def extract(self, inputfile, line):
...
+        #
+        #        Total Density - Mulliken Population Analysis
+        #     ---------------------------------------------
+        #
+        #    Atom       Charge   Shell Charges
+        # -----------   ------   -------------------------------------------------------
+        #    1 O    8     8.89   0.79  1.19  0.57  0.98  0.28  2.01  2.08  0.96  0.02 -0.03  0.04
+        #    2 H    1     0.55   0.20  0.26  0.05  0.04
+        #    3 H    1     0.55   0.20  0.26  0.05  0.04
+        if line.strip() == "Total Density - Mulliken Population Analysis":
```
berquist, line 804: This line will be modified in the block below to something like
```python
if line.strip() in (
    "Mulliken analysis of the total density", "Total Density - Mulliken Population Analysis"
):
```

berquist (changes requested):
Thanks for the fix!  Most of this code is duplicated with the section below it with the comment starting "For Hartree-Fock, the atomic Mulliken ..." and they can be combined if the conditional is modified to accept both trigger strings.

berquist:
Something happened with CI in the initial PR commit and I've found locally that this breaks a regression test.  Normally the contents of the Mulliken section may contain
```
Atom       Charge   Shell Charges
 -----------   ------   -------------------------------------------------------
    1 O    8     8.89   0.79  1.19  0.57  0.98  0.28  2.01  2.08  0.96  0.02 -0.03  0.04
    2 H    1     0.55   0.20  0.26  0.05  0.04
```
but in this old regression, which adding this new header now triggers, contains
```
Bfn.    Population  Atom+Function
   ---------------------------------
     1      1.997265     1 C  s
     2      0.743630     1 C  s
```

berquist:
Thanks for getting this started!

Issue #1571, opened 2025-04-13
Single `pip install` command including URL-based requirements


berquist:
The proper fix to https://github.com/cclib/cclib/pull/1382 and https://github.com/cclib/cclib/issues/1570.

berquist:
It looks like it isn't possible to do this, since each entry under `tool.setuptools.dynamic.optional-dependencies` is expected to be an object with a `file` key.  At best that means having a separate `requirements*.txt` for each group, maybe restricted to `dev` and `bridges`.  No idea if this would pass submitting to PyPI though...will try and do a revert if it fails.

Here is a starter:
```diff
diff --git pyproject.toml pyproject.toml
index 4128d260..d99d8678 100644
--- pyproject.toml
+++ pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "cclib"
-dynamic = ["version"]
+dynamic = ["version", "optional-dependencies"]
 description = "parsers and algorithms for computational chemistry"
 readme = "README.md"
 authors = [
@@ -42,31 +42,6 @@ ccget = "cclib.scripts.ccget:ccget"
 ccwrite = "cclib.scripts.ccwrite:main"
 cda = "cclib.scripts.cda:main"
 
-[project.optional-dependencies]
-bridges = [
-    "ase>=3.21",
-    "biopython",
-    "pandas",
-    "pyscf",
-    "pyscf-properties @ git+https://github.com/pyscf/properties",
-    "pyberny",
-    "geometric",
-    "qc-iodata>=1.0.0a2",
-]
-docs = [
-    "sphinx",
-    "sphinx_rtd_theme",
-]
-test-infrastructure = [
-    "pytest",
-    "coverage",
-    "pytest-cov",
-    "pyyaml",
-]
-test = ["cclib[bridges,docs,test-infrastructure]"]
-dev = ["cclib[bridges,docs,test-infrastructure]"]
-all = ["cclib[bridges]"]
-
 [build-system]
 requires = [
     "setuptools>=61.0",
@@ -77,6 +52,22 @@ build-backend = "setuptools.build_meta"
 [tool.setuptools]
 zip-safe = false
 
+[tool.setuptools.dynamic.optional-dependencies]
+bridges = { file = ["requirements-dev.txt"] }
+# docs = [
+#     "sphinx",
+#     "sphinx_rtd_theme",
+# ]
+# test-infrastructure = [
+#     "pytest",
+#     "coverage",
+#     "pytest-cov",
[...]

Issue #1570, opened 2025-04-13
Remove URL-based requirement for PySCF properties repo


berquist:
Package uploading to (Test)PyPI is broken because https://github.com/cclib/cclib/pull/1517 put a direct reference in the pyproject.toml optional dependencies.  This is a repeat of https://github.com/cclib/cclib/pull/1382.

A check for this is still not implemented upstream (https://github.com/pypa/twine/issues/430).

A new complication over #1382 is that CI is doing `pip install .[test]` and we are probably further encouraging the use of third-party dependencies via bridges.  Using `requirements-dev.txt` means `pyquante2` and `pyscf-properties` are not available via a single `pip install` command.  https://stackoverflow.com/a/73600610/ might be a solution to this.

PR #1585, opened 2025-04-25
Switch optional dependencies to requirements.txt style

berquist:
Possible fix for #1570 and #1571 

I won't know until merging and attempting the TestPyPI upload.


berquist:
`pip install '.[foo]'` doesn't error when `foo` doesn't exist :roll_eyes:

PR #1586, opened 2025-04-25
Revert "Switch optional dependencies to requirements.txt style"

berquist:
Reverts cclib/cclib#1585


berquist:
I'm going to make a separate PR that does https://github.com/cclib/cclib/issues/1570 only to make the merge commit less confusing.

PR #1610, opened 2025-06-01
Nuclear: fix off-diagonal element sign in moment of inertia tensor

berquist:
I was wondering why most principal moments of inertia and rotational constants were testing ok, but some were not, and the full MOI tensor wasn't correct.  Turns out the code I adapted from doing the Crawford exercises had a bug, fixed in https://github.com/CrawfordGroup/ProgrammingProjects/pull/26.  When you only look at the eigenvalues the problem is mostly hidden.  Discovered this when reading https://gitlab.com/dalton/dalton/-/blob/8e2942165f899909c4afddec30eb31a87ce268e0/DALTON/abacus/abavib.F#L715 to confirm units.

Just the nuclear methods part of #985, without the QCElemental updates.

### DIFF
--- a/cclib/method/nuclear.py
+++ b/cclib/method/nuclear.py
@@ -13,7 +13,7 @@ from cclib.parser.utils import PeriodicTable, convertor
 
 import numpy as np
 import periodictable as pt
-import scipy.constants
+import scipy.constants as spc
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


amandadumi (approved):
This looks good to me!

shivupa (approved):

PR #1614, opened 2025-06-07
PySCF: don't run unit tests unless they change

### LINKED ISSUE #1602: Testing is slow
Particularly after PySCF was added to `testdata`; these could be added under a filter similar to methods, or made nightly.

berquist:
...but always run basic bridge tests

Closes #1602


berquist:
Looks like regular unit/regression tests are down to 2 minutes without coverage and 4 minutes with.  Docs build is slow...

PR #1626, opened 2025-06-28
FChk: parse multipole moments saved by Q-Chem

berquist:
These sections are specific to Q-Chem.


berquist:
Found this in my local branches and fixed it up.  The CI failure led me to discover that the geometry optimization tests aren't run for formatted checkpoint files, so there will be a follow-up that implements some of that.

PR #1640, opened 2025-07-24
Avoid InvalidVersion when parsing orca with `-f.N` version suffix

ktns:
Hi. 

I tried to parse logs from orca 6.0.1-f.3 with cclib@master, and got `packaging.version.InvalidVersion` error.

I'm not sure this is right way to go, but replacing `-f.N` with `.postN` just worked.

### DIFF
--- a/cclib/parser/orcaparser.py
+++ b/cclib/parser/orcaparser.py
@@ -150,8 +150,10 @@ class ORCA(logfileparser.Logfile):
         if "Program Version" == line.strip()[:15]:
             # Handle development versions.
             self.metadata["legacy_package_version"] = line.split()[2]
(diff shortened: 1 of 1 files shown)


berquist (approved):
Yes, I think that's a fine mapping onto the PEP440 version.

Could you post a sample output file that exhibits the problem as a PR to https://github.com/cclib/cclib-data, so that we can run this in tests?  My copy of 6.0.1 doesn't do this.

ktns:
I'm using a commercial version of ORCA and it seems to have `-f.N` after a release version. Please take a look at cclib/cclib-data#233.

berquist (approved):
Thanks for doing this!

PR #1647, opened 2025-08-01
Add more type annotations

berquist:
A follow-up to https://github.com/cclib/cclib/pull/1179.


berquist:
This still doesn't enforce mypy/ty checks, which won't happen until 2.0 or after is released.

PR #1653, opened 2025-08-14
Fix handling of geometry indices for FIleWriter

### LINKED ISSUE #1566: OpenMolcas .xyz file generation from .log file not working
The command ccwrite xyz x.log gives the coordinates of the initial input instead of the optimised coordinates.

berquist:
Closes #1566


berquist:
These tests so far don't expose the bug, which resides somewhere in FileWriter or ccwrite, but are to be sure I don't break anything. Coverage here wasn't great.

berquist:
The offending code is https://github.com/cclib/cclib/blob/c91d478116e9d533c444ac7af87fa0e1ae657f66/cclib/scripts/ccwrite.py#L117-L120 which causes `index == 0` to fail.  It should just be removed.  I still need to figure out how to write a test for this.

PR #1670, opened 2025-10-02
Total Energy

lukas-paetow:


cclib/parser/serenityparser.py
```diff
@@ -45,3 +45,7 @@ def extract(self, inputfile, line):
 
         if line[5:21] == "Basis Functions:":
             self.set_attribute("nbasis", int(line.split()[2]))
+
+        # Extract total energy
```
berquist, line 49: You can get rid of this comment, since it is clear what this section does from the next line.

berquist (changes requested):

PR #1672, opened 2025-10-02
dispersionenergies

lukas-paetow:


cclib/parser/serenityparser.py
```diff
@@ -45,3 +45,7 @@ def extract(self, inputfile, line):
 
         if line[5:21] == "Basis Functions:":
             self.set_attribute("nbasis", int(line.split()[2]))
+
+        # Extract dispersion correction
```
berquist, line 49: This comment doesn't add anything that isn't already clear from the next line.

cclib/parser/serenityparser.py
```diff
@@ -45,3 +45,7 @@ def extract(self, inputfile, line):
...
+        if "Dispersion Correction (" in line:
+            self.set_attribute("dispersionenergies", [float(line.split()[3])])
```
berquist, line 51:
```suggestion
            self.append_attribute("dispersionenergies", float(line.split()[3]))
```
because there will be one for each step of a geometry optimization, PES scan, etc.

berquist (changes requested):

berquist (approved):

PR #1675, opened 2025-10-08
Serenity: Molecular Orbital Information

lukas-paetow:
Serenity keeps MO information in binary .h5 files that can be found in a directory with the given molecular system name. My implementation assumes that the output file is still where it was left by Serenity, so that the molecule directory can be found relative to that path.


cclib/parser/serenityparser.py
```diff
@@ -32,11 +36,34 @@ def before_parsing(self):
         self.unrestricted = False
 
     def after_parsing(self):
+        # Get molecular orbital information
+        orbpath = self.path.parent / self.systemname / f"{self.systemname}.orbs.res.h5"
+        if os.path.isfile(orbpath):
```
berquist:
```suggestion
        if orbpath.is_file():
```

cclib/parser/serenityparser.py
```diff
@@ -32,11 +36,34 @@ def before_parsing(self):
...
+                self.set_attribute("moenergies", eigenvalues)
+                self.set_attribute("mocoeffs", coeffs)
+                self.set_attribute("nmo", len(eigenvalues[0]))
+
         super().after_parsing()
 
     def extract(self, inputfile, line):
         """Extract information from the file object inputfile."""
 
+        self.path = Path(inputfile.filenames[0]).resolve()
```
berquist: Set this in `before_parsing`, otherwise it will be done on every call to `extract`.

berquist (changes requested):

berquist (approved):

PR #1684, opened 2025-10-15
V2 pint unit management

amandadumi:
This pull request integrates pint in for managing units. Specifically this pull request
- starts a single unit registry for the code base
- attaches units to the scfenergies attributes
- changes the test which use scfenergies, but don't have units implemented themselves to reference just the magnitude of the unit-labeled number (testMP and testCC)


cclib/attribute_parsers/ccenergies.py
```diff
@@ -6,7 +6,7 @@
 
 from cclib.attribute_parsers import utils
 from cclib.attribute_parsers.base_parser import base_parser
-
+from cclib import ureg
```
shivupa: can we call this something different? Like unit_types or something like that

shivupa (approved):

cclib/__init__.py
```diff
@@ -19,6 +19,9 @@
 
 # ruff: noqa: F401
 from cclib._version import __version__
+from pint import UnitRegistry
+ureg = UnitRegistry()
+Q_ = ureg.Quantity
```
berquist: Do we need Q?  I forget what it's for.

cclib/__init__.py
```diff
@@ -19,6 +19,9 @@
 
 # ruff: noqa: F401
 from cclib._version import __version__
+from pint import UnitRegistry
+ureg = UnitRegistry()
```
berquist: I agree with Shiv about the name but think it should be an alias, since everyone else is using `ureg`.

cclib/attribute_parsers/scfenergies.py
```diff
@@ -21,15 +23,16 @@ def gaussian(file_handler, ccdata) -> Optional[dict]:
         line = file_handler.last_line
         if line[1:9] == "SCF Done":
             constructed_data = utils.float(line.split()[4])
-            return {scfenergies.__name__: [constructed_data]}
+            constructed_data = [constructed_data] * ureg.hartree
```
berquist: TIL
```
In [1]: import pint

In [2]: x = 4.5

In [3]: xl = [x]

In [4]: x
Out[4]: 4.5

In [5]: xl
Out[5]: [4.5]

In [6]: ureg = pint.UnitRegistry()

In [7]: ureg.hartree
Out[7]: <Unit('hartree')>

In [8]: xl * ureg.hartree
Out[8]: <Quantity([4.5], 'hartree')>

In [9]: (xl * ureg.hartree).magnitude
Out[9]: array([4.5])
```

test/data/testCC.py
```diff
@@ -17,14 +17,14 @@ class GenericCCTest:
     def testsizeandshape(self, data) -> None:
         """Are the dimensions of ccenergies correct?"""
         assert data._ccCollection._parsed_data[0].ccenergies.shape == (
-            len(data._ccCollection._parsed_data[0].scfenergies),
+            len(data._ccCollection._parsed_data[0].scfenergies.magnitude),
```
berquist: Do we need to?  This worked for me:
```
In [11]: xl = [x, x, x]

In [12]: len(xl * ureg.hartree)
Out[12]: 3
```

berquist (changes requested):
This is great.  The one thing IMO we need to decide on is where exactly to attach the units in the parsers.  My initial thought was right at the parse location (`utils.float(line.split()[0]) * ureg....` etc.) but there is something to be said for in the return value like `return {scfenergies.__name__: [constructed_data] * ureg.hartree}`.  That way is cleaner but further from where the data was parsed, esp. if it's in non-standard units.

test/data/testCC.py
berquist: (The other stuff comparing the actual values definitely doesn't work without `magnitude`, that's all fine)

amandadumi:
> This is great. The one thing IMO we need to decide on is where exactly to attach the units in the parsers. My initial thought was right at the parse location (`utils.float(line.split()[0]) * ureg....` etc.) but there is something to be said for in the return value like `return {scfenergies.__name__: [constructed_data] * ureg.hartree}`. That way is cleaner but further from where the data was parsed, esp. if it's in non-standard units.

Yes this is true. I have introduced one change, but i think there may be another. 
- I made one change which is instantiating the list with units as the array is created.
- What might be missing is attaching units to the specific parsed value that results from this entry into the specific parser. 

```
# initialize empty array with units attached
>>> a = np.array([])*ureg.hartree
>>> a
<Quantity([], 'hartree')>

# adding unit tagged option of same unit is a clean process.
>>> b = 4.5*ureg.hartree
>>> b
<Quantity(4.5, 'hartree')>
>>> np.append(a,b)
<Quantity([4.5], 'hartree')>

# if for any reason different units come up during parsing, this can still be cleanly kept. (can't think of a use case, but just what's possible)
>>> c = 3.5*ureg.eV
>>> c
<Quantity(3.5, 'electron_volt')>
>>> np.append(a,c)
<Quantity([0.12862263], 'hartree')>
```

From in person conversations, I am just attaching native units from the given program and we will determine when to convert to cclib units at a different point, but it's not clear where that will happen yet.

cclib/attribute_parsers/ccenergies.py
amandadumi: i renamed this, but didn't quite alias with ureg so i can change this some other way if needed

cclib/__init__.py
amandadumi: removed

cclib/__init__.py
amandadumi: I did make it more clear, but didn't do an alias as i wasn't sure where that should happen

test/data/testCC.py
amandadumi: yeah you are right, this is fixed now.

PR #1688, opened 2025-10-21
Fix: updated the parser to handle ORCA cube files.

hanaol:
I have an ORCA-generated cube file that follows the format described in this [link](https://h5cube-spec.readthedocs.io/en/latest/cubeformat.html). The original script processes the {DSET_IDS (#x int)} line together with the voxel data, which I have now updated. The additional code block I introduced is based on the implementation from [ASE](https://gitlab.com/ase/ase/-/blob/master/ase/io/cube.py?ref_type=heads#L78).

Example snippet:

```
Cube data generated by ORCA            # comment
Molecular orbital 0 of operator 0      # comment
  -89  -19.960787  -21.151690  -14.790883  # 89: number of atoms, (-) → {DSET_IDS (#x int)}
   40    0.951233    0.000000    0.000000  # grid info
   40    0.000000    1.049523    0.000000
   40    0.000000    0.000000    0.768611
   45   17.000000    0.329116    2.304392   -1.051723  # atomic info
    6    6.000000    7.086399   -3.364878    1.916365
    ...
    1    1.000000   -7.524159   -5.810954    6.626053
    1    0  # {DSET_IDS (#x int)}
   0.00000E+00   0.00000E+00   0.00000E+00   0.00000E+00   0.00000E+00   0.00000E+00
   0.00000E+00   0.00000E+00   1.97079E-14   2.50774E-14   4.81111E-14  -7.67704E-14
  -1.70954E-13  -3.56271E-13  -6.83331E-13  -1.21962E-12  -1.98377E-12  -3.07282E-12
  -4.45770E-12  -6.04516E-12  -7.83636E-12  -9.70775E-12  -1.15760E-11  -1.32736E-11
```

### DIFF
--- a/.github/scripts/run_pytest_methods.bash
+++ b/.github/scripts/run_pytest_methods.bash
@@ -22,4 +22,4 @@ else
 fi
 
 PYTEST_ADDOPTS="-v -s ${PYTEST_PARALLELISM} ${PYTEST_COVERAGE} --cov-report=xml:coverage-method.xml -m 'is_method'" python -m pytest
-PYTEST_ADDOPTS="-v -s ${PYTEST_PARALLELISM} ${PYTEST_COVERAGE} --cov-report=xml:coverage-method.xml --cov-append -m 'is_method'" python -m pytest test/regression_method.py
+PYTEST_ADDOPTS="-v -s ${PYTEST_PARALLELISM} ${PYTEST_COVERAGE} --cov-report=xml:coverage-method.xml --cov-append" python -m pytest test/regression_method.py


cclib/method/volume.py
```diff
@@ -431,10 +440,50 @@ def read_from_cube(filepath):
...
+        datas = numpy.array(datas)
+        tmp = datas[0]
+        tmp = tmp.flatten()
+        
+        '''
         tmp = []
         for line in lines[skiplines:]:
             tmp.extend(line.split())
         tmp = numpy.asanyarray(tmp, dtype=float)
+        '''
```
berquist: If this code is no longer used, you can delete it instead of commenting it out, since it will still be a part of version control history.

cclib/method/volume.py
```diff
@@ -460,5 +509,4 @@ def read_from_cube(filepath):
 
     # Assign grid data
     vol.data = numpy.resize(tmp, vol.data.shape)
-
```
berquist: Erroneous change?

berquist (changes requested):
Thank you for the fix.  Do you have an example cube file that's small and triggers the problem?  I can write a test for your fixes but need the file for reference.  Attaching it in a comment is fine.

cclib/method/volume.py
hanaol: I’ve committed the change.

berquist:
This will fail because I haven't added the two cube files to the cclib-data repo.  I'll rerun tests once that's done.

berquist (approved):
Thanks for the fixes and your patience!

PR #1694, opened 2025-10-25
Serenity: Metadata

lukas-paetow:
Metadata: Methods will be done at a later point when the parser can differentiate between "system" and "task" sections in Serenity output files. Warnings printed may not be captured completely for now, which is easiert to fix later when thoroughly testing at the end.


test/data/testSP.py
```diff
@@ -612,7 +608,7 @@ def testmetadata_symmetry_detected(self, data) -> None:
     @skipForParser("NBO", "attribute not implemented in this version")
     @skipForParser("Turbomole", "reading point group symmetry and name is not implemented")
     @skipForParser("xTB", "not implemented yet")
-    @skipForParser("Serenity", "not implemented yet")
+    @skipForParser("Serenity", "not applicable")
```
berquist: If true, say instead something like
```suggestion
    @skipForParser("Serenity", "Serenity doesn't use symmetry")
```
because while this message isn't strictly necessary, it's a good habit to get into because sometimes skip reasons are more complicated.

test/data/testSP.py
```diff
@@ -600,7 +596,7 @@ def testmetadata_success(self, data) -> None:
...
```
berquist: Same

PR #1695, opened 2025-10-27
Serenity: TDDFT

lukas-paetow:


cclib/parser/serenityparser.py
```diff
@@ -194,3 +206,82 @@ def extract(self, inputfile, line):
...
+
+            exc_iterator = 1
+            transition_data = []
+            while not line.strip().startswith("--"):
+                if line.strip().startswith(str(exc_iterator)):  # new exc
+                    if exc_iterator > 1:
+                        self.append_attribute("etsecs", transition_data)
+                        transition_data = []
+                    self.append_attribute("etenergies", line.split()[1])
+                    i = int(line.split()[6])
```
berquist: Store the results from `line.split()` in a variable that you reuse (this happens in a few places in the PR).

cclib/parser/serenityparser.py
```diff
@@ -52,6 +53,17 @@ def after_parsing(self):
 
         super().after_parsing()
 
+    def convert_to_spin(self, symbol):
```
berquist: If you don't need the flexibility of a function, another way of doing it is https://github.com/cclib/cclib/blob/ddcf48ea79019427e93103cdcb39930ddbede66d/cclib/parser/orcaparser.py#L1306-L1331.  But I don't think it would be possible to as cleanly handle the error case, so the function is nicer in that regard.

cclib/parser/serenityparser.py
```diff
@@ -52,6 +53,17 @@ def after_parsing(self):
...
+        if symbol == "a":
+            return 0
+        elif symbol == "b":
+            return 1
+        else:
+            self.logger.warning(
```
berquist: Since you're making this unrecoverable by throwing `StopParsing`, this could probably be an error instead of a warning.

berquist (changes requested):

PR #1698, opened 2025-11-08
Serenity: Polarizability

lukas-paetow:


cclib/parser/serenityparser.py
```diff
@@ -402,6 +414,7 @@ def extract(self, inputfile, line):
             self.metadata["package_version"] = line.split()[2]
 
         ### EXCITED STATE
+        # TODO add CC2 etc
```
berquist: This comment is stale.

berquist (changes requested):
Can you use the same system and settings as https://github.com/cclib/cclib/blob/f8a17414debec02a7e823275fbb7c22a044d6db2/data/QChem/basicQChem5.4/Trp_polar.in ?

berquist:
Sorry for taking so long to get back to this.  Now that it's the summer I have time.

I have finally figured out how to control everything.  The numerical input is

```
+system
 name trp_polar
 geometry trp_polar.xyz
 method hf
 +basis
  label sto-3g
 -basis
-system

+task scf
 act trp_polar
-task

+task ff
 act trp_polar
 frequency 0.0
-task
```

and the analytical input is

```
+system
 name trp_polar
 geometry trp_polar.xyz
 method hf
 +basis
  label sto-3g
  # Otherwise RI will be attempted in the linear response part
  densFitJ NONE
 -basis
-system

+task scf
 act trp_polar
-task

+task ff
 act trp_polar
 frequency 1.0e-9
-task
```

because we generate tests without any form of RI unless absolutely necessary.  The analytical result for the isotropic polarizability matches Q-Chem exactly.

For the branch history, I am going to try to cherry-pick everything into a new branch to get rid of the merges.

berquist:
On second thought, it will be easier for all of us if I just squash this.

I will do some test data cleanup in a follow up.

berquist (approved):

PR #1775, opened 2026-05-02
pyproject.toml: Make licensing conformant to PEP 639

berquist:
See https://peps.python.org/pep-0639/ for more information.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,34 +1,34 @@
 [project]
 name = "cclib"
-dynamic = ["version"]
 description = "parsers and algorithms for computational chemistry"
 readme = "README.md"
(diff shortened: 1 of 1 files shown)


berquist:
It looks like the last version of setuptools that will install on Python 3.8 is 75.3.4.

Related: https://endoflife.date/python

I know that RHEL 8 Python default is 3.6 but it's easy to install newer versions.

Current QCElemental is requiring 3.7 but will switch to 3.9 once Pydantic v2 support is finished.

I think bumping `master` to 3.9 for 1.9 is ok, and 2.0 on `main` will probably require 3.9 or 3.10.  The latest pint version requires 3.11.

PR #1789, opened 2026-05-21
End of spin down orbitals section properly detected

ecnicolas:
In Orca 6.1.1, the final line ends with 25*'-', it was not detected with the previous definition. It should not affect previous versions (the previous lines are > 25 chars long anyway)


berquist (approved):
Thanks!

PR #1802, opened 2026-06-11
Fix NWChem parsing when atom names are not capitalized

### LINKED ISSUE #1801: NWChem parsing error: KeyError: 'N'
cclib version: 1.8.1
NWChem version: 7.3.0
python version: 3.11.13

This is an example from the NWChem tutorial, with the input

```
 title "Nitrogen cc-pvdz SCF geometry optimization"  
 geometry    
   n 0 0 0  
   n 0 0 1.08  
 end  
 basis  
   n library cc-pvdz  
 end  
 task scf optimize
```

output attached.  When I import this I get

```
>>> import cclib
>>> file = '/Users/jasonb/testdir123/n2_opt.out'
>>> data = cclib.io.ccread(file)
[NWChem /Users/jasonb/testdir123/n2_opt.out ERROR] Encountered error when parsing.
[NWChem /Users/jasonb/testdir123/n2_opt.out ERROR] Last line read:       n                 14.003070

  File "/Users/jasonb/miniconda3/envs/my-cclib-env/lib/python3.11/site-packages/cclib/io/ccio.py", line 185, in ccread
    return log.parse()
           ^^^^^^^^^^^
  File "/Users/jasonb/miniconda3/envs/my-cclib-env/lib/python3.11/site-packages/cclib/parser/logfileparser.py", line 165, in parse
    self.extract(self.inputfile, line)
  File "/Users/jasonb/miniconda3/envs/my-cclib-env/lib/python3.11/site-packages/cclib/parser/nwchemparser.py", line 103, in extract
    masses = [name2mass[self.table.element[number]] for number in self.atomnos]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jasonb/miniconda3/envs/my-cclib-env/lib/python3.11/site-packages/cclib/parser/nwchemparser.py", line 103, in <listcomp>
    masses = [name2mass[self.table.element[number]] for number in self.atomnos]
              ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'N'
```

I don't know how the n gets turned into an N.

[n2_opt.txt](https://github.com/user-attachments/files/28815578/n2_opt.txt)

gaoflow:
Fixes #1801.

NWChem echoes atom names with the capitalization the user typed in the input geometry — the attached `n2_opt.out` from the issue has `n 14.003070` in its Atomic Mass block because the input said `n 0 0 0`. The parser, however, resolves atoms through standard `PeriodicTable` symbols (`'N'`), so `name2mass['N']` raised the reported `KeyError: 'N'`. The same as-typed-name vs. canonical-symbol mismatch exists in the two basis summary sections (`gbasis_dict`/`atombasis_dict` via `name2element`) and in `self.shells`, whose `after_parsing` consumer looks up by element and silently dropped `aonames` with a warning when the keys didn't match.

Changes:
- `name2element` now capitalizes the extracted symbol, fixing the `gbasis`/`atombasis` lookups; the Atomic Mass block gets the same normalization, and `self.shells` is keyed by element so `aonames` are built.
- With the KeyError gone, the issue's file then crashed in `Symmetry analysis of basis`: that block has one line per irrep, and the parser read `pg_order` lines, which only matches for abelian groups. The file's D4h has order 16 but 10 irreps, so the parser ran past the block into a blank line (`IndexError`). It now reads until the blank line that ends the block.

With both fixes the file from the issue parses end to end: `atommasses` is `[14.00307, 14.00307]`, `atombasis`/`gbasis` resolve, and `aonames` (`N1_1S`…) are produced instead of being dropped.

Tests: two unit tests in `test/parser/testspecificparsers.py` drive the parser over a minimal fragment modeled on the issue's file — one with lowercase atom names, one with capitalized names isolating the non-abelian symlabels case. Both fail on master and pass with the fix; the existing NWChem data tests (486) pass unchanged.

### DIFF
--- a/cclib/parser/nwchemparser.py
+++ b/cclib/parser/nwchemparser.py
@@ -38,7 +38,10 @@ class NWChem(logfileparser.Logfile):
 
     @staticmethod
     def name2element(lbl: str) -> str:
-        return "".join(itertools.takewhile(str.isalpha, str(lbl)))
+        # NWChem echoes atom names with the capitalization used in the input
+        # (e.g. "n" or "FE"), while these are looked up against the standard
+        # symbols in PeriodicTable, so normalize the case.
+        return "".join(itertools.takewhile(str.isalpha, str(lbl))).capitalize()
 
     def extract(self, inputfile: "FileWrapper", line: str) -> None:
         """Extract information from the file object inputfile."""
@@ -106,7 +109,9 @@ class NWChem(logfileparser.Logfile):
                 tokens = next(inputfile).split()
                 mtch = re.search(r"[a-zA-Z]+", tokens[0])
                 assert mtch is not None
-                name2mass[mtch.group()] = float(tokens[1])
+                # The atom name is echoed as typed in the input (e.g. "n"),
+                # but is looked up below via the standard PeriodicTable symbol.
+                name2mass[mtch.group().capitalize()] = float(tokens[1])
             masses = [name2mass[self.table.element[number]] for number in self.atomnos]
             self.set_attribute("atommasses", masses)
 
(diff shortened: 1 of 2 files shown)


berquist:
This is failing because of changes I made to the underlying CI images after the PR was made but before I approved CI to run.

However, while we develop a policy on AI-based contributions (https://github.com/cclib/cclib/issues/1749), particularly those from agents, we won't be accepting any AI-generated code.

PR #1814, opened 2026-06-14
Optionally depend on `openbabel`, not separate deprecated wheel

### LINKED ISSUE #1813: `openbabel-wheel` is deprecated, switch back to `openbabel`
https://github.com/cclib/cclib/pull/1729 sat for so long that OB 3.2.0 now packages wheels for Py 3.10+.

berquist:
Closes #1813

Open Babel 3.2.0+ now provides pre-built wheels on PyPI.


berquist:
I could turn off `VERSIONINGIT_FOR_PACKAGE_INDEX` for < 3.10 but I think that will create multiple build artifacts, and I'm not sure the publishing workflow will handle that properly.

berquist:
> I could turn off `VERSIONINGIT_FOR_PACKAGE_INDEX` for < 3.10 but I think that will create multiple build artifacts, and I'm not sure the publishing workflow will handle that properly.

https://github.com/cclib/cclib/blob/15f1d7702e826b4c25e00a967f8c3759a3f7fb50/.github/workflows/test_and_package.yml#L102-L107 shows that we only save a single build artifact, and it's for a newer version of Python that wouldn't have this problem, so the versioningit workaround is worth trying.

PR #1847, opened 2026-06-30
Fixes and additions for xTB

oliver-s-lee:
Just some minor fixes and additions for xTB.

Fixed moenergies (wrong units) and homos (which was not implemented correctly, seemed to be tracking occupied orbital indices), and added moments and a few other small things.

The quadrupole moment test is currently failing, I'm not sure if this is as unit conversion problem or if the xTB method just results in a drastically different quadrupole to DFT


cclib/parser/xtbparser.py
```diff
@@ -42,7 +42,9 @@ def before_parsing(self) -> None:
 
     def after_parsing(self) -> None:
         """Actions after parsing"""
-        pass
+        # We can work out our multiplicity based on the number of unpaired electrons we have.
```
berquist: This can probably go in the parent class `after_parsing` since it's general and not specific to xTB.

cclib/parser/xtbparser.py
```diff
@@ -72,18 +74,40 @@ def extract(self, inputfile: "FileWrapper", line: str) -> None:
...
             self.set_attribute("mult", mult)
+        
+        if _is_scc_line(line):
+            # Parse SCF (SCC here) convergence.
+            # We have 5 fields we can parse (E, dE, RMSdq, gap, omega)
+            # For now, we'll just take E.
+            # Note that the SCC energy != final energy, there seems to be some additional minor contribution(s)...
+            scc_values = []
+            line = next(inputfile)
+            while line.strip() != "":
```
berquist:
```suggestion
            while line.strip():
```
(I think)

cclib/parser/xtbparser.py
```diff
@@ -288,6 +328,41 @@ def extract(self, inputfile: "FileWrapper", line: str) -> None:
                     grads[-1].append([float(v) for v in line.split()])
                 line = next(inputfile)
             self.set_attribute("grads", np.array(grads))
+        
+        if _is_moment_line(line):
+            # We don't seem to be given the origin unfortunately.
```
berquist, line 331: I'm looking at the xtb code trying to figure this out.  In the meantime, aren't there some parsers that set this to `[None, None, None]` when they can't figure it out?  I don't remember.

berquist (commented):
Partial review, lmk if you want me to look at the quadrupole stuff.

cclib/parser/xtbparser.py
oliver-s-lee: Just had a check through and it seems most just assume `[0.0, 0.0, 0.0]` if it's not available

cclib/parser/xtbparser.py
oliver-s-lee: Ah yes good point

PR #1883, opened 2026-07-29
Jaguar 13.3 test data

shivupa:
From Ty Balduf (@TyBalduf):
```
The format isn't too different from the old 8.3 examples you already have. If it's helpful, I included all the relevant output files from which data could be parsed. For example, besides the .out, Jaguar produces:
* a .[01.in](http://01.in/) file, which functions as a restart file to launch a new calculation with the final geometry and SCF/Hessian guess from a prior calculation
* a .01.mae file, which is a structured format containing the main scalar structure/atomic properties. The format has a publicly available parser: https://github.com/schrodinger/maeparser 
* TDDFT calculations produce a _uvv.spm file, used for constructing a basic stick spectrum plot
* Similarly, frequency calculations produce an .spm file for the IR (and Raman if computed), as well as .vib file which includes the normal mode coordinates.
* Scan calculations produce a Macromodel (Schrodinger's conformer search program) style .grd file, which contains the relative energy in kcal/mol of all the computed steps.
```

So we could actually parse even more than we are right now. But this should parse + test this data


shivupa:
GTG besides codecov which is probably fine

berquist (approved):

